PR #6, opened 2018-02-08
Various fixes

gyoshev:
Moved from https://github.com/telerik/kendo-theme-default/pull/866


gyoshev:
Seems like this was already merged during the transition of the themes.

PR #8, opened 2018-02-08
chore: change upload dropzone styles

gyoshev:
see telerik/kendo-angular-upload#115

Moved from https://github.com/telerik/kendo-theme-default/pull/867

### DIFF
--- a/packages/default/scss/upload/_layout.scss
+++ b/packages/default/scss/upload/_layout.scss
@@ -27,7 +27,7 @@
             border-width: 0;
             background-color: transparent;
 
-            em {
+            .k-dropzone-hint {
                 margin-left: $spacer-x;
                 flex: 1;
                 overflow: hidden;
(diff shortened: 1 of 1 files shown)


packages/default/scss/upload/_layout.scss
```diff
@@ -27,7 +27,7 @@
             border-width: 0;
             background-color: transparent;
 
-            em {
+            .k-dropzone-hint {
```
gyoshev: I think this change is for the better, however it will require a change in the rendering of the Kendo UI for jQuery upload, too. This in turn will require the change to all LESS themes -- @joneff what do you think?

gyoshev: By the way, a way to keep the theme backwards-compatible is to keep both selectors for a time, and open an issue in KUI jQuery to change the rendering. Keeping both selectors increases the style size, but prevents the theme from breaking until the necessary changes are implemented across suites. Look for `BACKCOMPAT` in the themes' source for instances of this method.

PR #10, opened 2018-02-09
Material pager

gyoshev:
Moved from https://github.com/telerik/kendo-theme-material/pull/25


packages/material/scss/pager/_layout.scss
```diff
@@ -0,0 +1,47 @@
...
+    $pager-elements-margin-lg: 32px !default;
+    $pager-page-link-size: 36px !default;
+
+    .k-pager-wrap {
+        padding: $pager-padding-y $pager-padding-x;
+
+        .k-icon {
+            width: $icon-size;
+            height: $icon-size;
+            line-height: $line-height-lg;
```
gyoshev: What is the reasoning behind this rule? With it, the icons are misaligned vertically:

image

Without this rule, this issue is resolved.

packages/material/scss/pager/_layout.scss
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        .k-pager-nav {
+            padding: $pager-nav-padding;
+            width: auto;
+            height: auto;
```
gyoshev: Perhaps using $pager-buttons-size from theme-default should remove the need for this rule, as well as the extra variable?

This will also remove the need for the last rule in this file.

packages/material/scss/pager/_theme.scss
```diff
@@ -0,0 +1,57 @@
+    @include exports("pager/theme") {
+    $pager-text: $header-text !default;
+    $pager-bg: $header-bg !default;
+    $pager-items-hover-bg: $list-item-hovered-bg !default;
+    $pager-numbers-text: $accent !default;
+    $pager-labels-text: rgba(0, 0, 0, .54) !default;
```
gyoshev: Most of the code in this file can be omitted if you include the `_theme` file from theme-default. Most likely only the border-radius is specific to the material theme.

packages/material/scss/pager/_layout.scss
```diff
@@ -0,0 +1,29 @@
...
+@import "~@progress/kendo-theme-default/scss/pager/layout";
+
+@include exports ('pager/layout/material') {
+    $pager-elements-margin: 8px !default;
+    $pager-elements-margin-lg: 32px !default;
+
+    .k-pager-wrap {
+        .k-link,
+        .k-state-selected {
+            padding: 0;
```
gyoshev: You can omit the rule altogether if the padding is exposed in theme-default. Since we are now using the monorepo, you can add a `$pager-buttons-padding` variable to theme-default (with a default value of the current padding), and then use it here, just like `$pager-buttons-size`.

packages/material/scss/pager/_theme.scss
```diff
@@ -0,0 +1,25 @@
+@include exports("pager/theme") {
+    $pager-text: rgba(0, 0, 0, .54) !default;
+
+    .k-pager-wrap {
+        color: $pager-text;
```
gyoshev: You can drop this declaration and the `.k-pager-numbers` rule if you include the `_theme` file from theme-default. Bonus: this will also provide the missing hover and focus states.

PR #11, opened 2018-02-09
Material upload

gyoshev:
Moved from https://github.com/telerik/kendo-theme-material/pull/26


packages/material/scss/upload/_theme.scss
```diff
@@ -0,0 +1,163 @@
+@include exports('upload/theme') {
+    $upload-text: $panel-text !default;
+    $upload-padding: 14px !default;
+
+    $upload-padding-sm: $popup-padding-y !default;
+
+    $upload-button-bg: rgba(0, 0, 0, 0.12) !default;
+    $upload-button-active-bg: rgba(0, 0, 0, 0.28) !default;
+
+    $upload-border-color: rgba(0, 0, 0, .08) !default;
```
gyoshev: @elena-gancheva after the ScrollView branch gets merged, apply the [suggestions for splitting the layout / theme files](https://github.com/telerik/kendo-themes/pull/12/files/8ede27abb3af79c2d9c8138df2ab5ea052553bed#diff-fb4ad1037708b4666473baedd4ef12ff) to this file.

packages/material/scss/upload/_theme.scss
```diff
@@ -38,18 +37,14 @@ $upload-success-text: $text-color !default;
             color: $upload-disabled-text;
 
             &:hover {
-                background: white;
+                background-color: transparent;
```
elena-gancheva: Don't we have convetsion to use only `background` property rule for color? :)

gyoshev: Not necessarily. Setting the `background` property sets _a lot_ of other properties, and we try to use it sparingly. It seems that in this case `background-color` suffices, thus the change :)

PR #12, opened 2018-02-09
feat: add styling for scrollview component

gyoshev:
Moved from https://github.com/telerik/kendo-theme-material/pull/27

### DIFF
--- a/packages/default/scss/scrollview/_layout.scss
+++ b/packages/default/scss/scrollview/_layout.scss
@@ -4,6 +4,7 @@
 @include exports('scrollview/layout') {
     $scrollview-page-button-size: 8px;
     $scrollview-arrow-icon-size: 4.5em;
+    $scrollview-spacing: 10px;
 
     .k-scrollview-wrap {
         position: relative;
@@ -45,12 +46,12 @@
         list-style: none;
         position: absolute;
         left: 0;
-        bottom: 10px;
+        bottom: $scrollview-spacing;
         pointer-events: none;
     }
     .k-pages > li,
     .k-scrollview-pageable > li.k-button {
-        margin: 10px;
+        margin: $scrollview-spacing;
         padding: 0;
         width: $scrollview-page-button-size;
         height: $scrollview-page-button-size;
--- a/packages/material/scss/scrollview/_theme.scss
+++ b/packages/material/scss/scrollview/_theme.scss
@@ -1,10 +1,10 @@
 $scrollview-navigation-color: $base-text;
+$scrollview-spacing: $scrollview-pager-spacing-x * 2;
 
 @import "~@progress/kendo-theme-default/scss/scrollview/theme";
 
 @include exports("scrollview/theme/material") {
     $scrollview-text-opacity: .4 !default;
-    $scrollview-bottom: $scrollview-pager-spacing-x * 2;
 
     .k-scrollview-next:hover,
     .k-scrollview-prev:hover {
@@ -14,7 +14,7 @@ $scrollview-navigation-color: $base-text;
     }
 
     .k-scrollview-pageable {
-        bottom: $scrollview-bottom;
+        bottom: $scrollview-spacing;
 
         li.k-button {
             background-color: rgba(black, $scrollview-text-opacity);
--- a/packages/material/scss/scrollview/_theme.scss
+++ b/packages/material/scss/scrollview/_theme.scss
@@ -1,33 +1,27 @@
+$scrollview-navigation-color: $base-text;
+
 @import "~@progress/kendo-theme-default/scss/scrollview/theme";
 
 @include exports("scrollview/theme/material") {
     $scrollview-text-opacity: .4 !default;
-    $scrollview-padding-y: 10px !default;
-    $scrollview-padding-x: $padding-y !default;
+    $scrollview-bottom: $scrollview-pager-spacing-x * 2;
 
-    .k-scrollview-next,
-    .k-scrollview-prev {
-        span {
-            color: $base-text;
-
-            &:hover {
-
-                &::before {
-                    background: rgba($base-bg, $scrollview-text-opacity);
-                }
-            }
+    .k-scrollview-next:hover,
+    .k-scrollview-prev:hover {
+        &::before {
+            background-color: rgba($base-bg, $scrollview-text-opacity);
         }
     }
 
     .k-scrollview-pageable {
-        bottom: $scrollview-padding-y * 2;
+        bottom: $scrollview-bottom;
 
         li.k-button {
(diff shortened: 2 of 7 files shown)


packages/material/scss/scrollview/_theme.scss
```diff
@@ -0,0 +1,34 @@
+@import "~@progress/kendo-theme-default/scss/scrollview/theme";
+
+@include exports("scrollview/theme/material") {
+    $scrollview-text-opacity: .4 !default;
+    $scrollview-padding-y: 10px !default;
+    $scrollview-padding-x: $padding-y !default;
+
+    .k-scrollview-next,
+    .k-scrollview-prev {
+        span {
```
gyoshev: Use `.k-icon` instead of a tag selector.

packages/material/scss/scrollview/_theme.scss
```diff
@@ -0,0 +1,34 @@
...
+            color: $base-text;
+
+            &:hover {
+
+                &::before {
+                    background: rgba($base-bg, $scrollview-text-opacity);
```
gyoshev: When setting the bg color, use the `background-color` property. It may be longer to write, but the `background` shorthand sets a _lot_ of other properties that you probably don't need. This applies to all `background` declarations in the file.

packages/material/scss/scrollview/_theme.scss
```diff
@@ -0,0 +1,34 @@
...
+
+@include exports("scrollview/theme/material") {
+    $scrollview-text-opacity: .4 !default;
+    $scrollview-padding-y: 10px !default;
+    $scrollview-padding-x: $padding-y !default;
+
+    .k-scrollview-next,
+    .k-scrollview-prev {
+        span {
+            color: $base-text;
```
gyoshev: The same effect can be achieved without additional styles, by setting `$scrollview-navigation-color: $base-text` before importing the default theme.

packages/material/scss/scrollview/_theme.scss
```diff
@@ -0,0 +1,34 @@
...
+            }
+        }
+    }
+
+    .k-scrollview-pageable {
+        bottom: $scrollview-padding-y * 2;
+
+        li.k-button {
+            background: rgba(black, $scrollview-text-opacity);
+            margin: $scrollview-padding-y $scrollview-padding-x;
```
gyoshev: Instead of setting the margins here, introduce these variables in the `theme-default` layout file, and override them before importing it in `_layout`. Renaming them to `$scrollview-pager-spacing-x` / `-y` seems more descriptive. This may also take care of the `bottom` property a few lines up, if it is calculated in theme-default.

packages/default/scss/scrollview/_layout.scss
```diff
@@ -4,6 +4,8 @@
 @include exports('scrollview/layout') {
     $scrollview-page-button-size: 8px;
     $scrollview-arrow-icon-size: 4.5em;
+    $scrollview-pager-spacing-x: $padding-y;
+    $scrollview-pager-spacing-y: 10px;
```
gyoshev: Don't you find it odd to have unused variables in the file? :smile: The variables are the public API of the theme. Changing them should allow users of the theme to change its properties. Having a variable that is unused is as frustrating as a component property that does nothing.

With my previous comment, I meant that introducing such variables in theme-default will allow you to remove the layout styles from theme-material.

To illustrate, instead of this...

```scss
// default/component/_layout.scss
.foo { padding: 10px; }

// material/component/_layout.scss
@import "default/component/_layout.scss";
.foo { padding: 16px; }
```
... you can use variables:

```scss
// default/component/_layout.scss
$foo-padding: 10px !default;
.foo { padding: $foo-padding; }

// material/component/_layout.scss
$foo-padding: 16px !default;
@import "default/component/_layout.scss";
```

The latter is preferred, because it simply changes the default value for theme-material, without the need to write extra CSS. This allows us to have less code to maintain, and also reduces the CSS output for the derived theme.

packages/material/scss/scrollview/_theme.scss
```diff
@@ -0,0 +1,34 @@
...
+    .k-scrollview-prev {
+        .k-icon {
+
+            &:hover {
+
+                &::before {
+                    background-color: rgba($base-bg, $scrollview-text-opacity);
+                }
+            }
+        }
```
joneff: The above could be `.k-icon:hover { ...  }` or even better

```scss
.k-scrollview-next:hover,
.k-scrollview-prev:hover {
    .k-icon { ... }
}

File: packages/material/scss/scrollview/_theme.scss [hunk 7]
```diff
@@ -0,0 +1,34 @@
...
+        li.k-button {
+            background: rgba(black, $scrollview-text-opacity);
+            margin: $scrollview-pager-spacing-y $scrollview-pager-spacing-x;
+
+            &.k-primary {
+                background: $primary;
+            }
+        }
+    }
+}
```
joneff: In the default theme, there are styles about this, albeit using different selectors. I guess we could unify the selectors and not add extra styles.

elena-gancheva: @joneff we have different rendering of elements' classes and IMO we should consider make it the same everywhere. @gyoshev what do you think?
These selectors are for overriding the buttons's styles from `theme-material` - `packages\material\scss\buttons\_theme.sccs`, not those from `theme-default`.

packages/material/scss/scrollview/_theme.scss
```diff
@@ -0,0 +1,28 @@
...
+
+    .k-scrollview-next:hover,
+    .k-scrollview-prev:hover {
+        &::before {
+            background-color: rgba($base-bg, $scrollview-text-opacity);
+        }
+    }
+
+    .k-scrollview-pageable {
+        bottom: $scrollview-bottom;
```
gyoshev: This rule overrides [this hard-coded line](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/scrollview/_layout.scss#L48) from theme-default. It would be great if we didn't have two rules (one from the default theme and one from the material theme) in the output CSS. To achieve this, you can introduce the `$scrollview-padding-*` variables in theme-default and take them into account. Then, override them before importing the `_layout` file.

packages/material/scss/scrollview/_theme.scss
```diff
@@ -0,0 +1,28 @@
...
+
+        li.k-button {
+            background-color: rgba(black, $scrollview-text-opacity);
+            margin: $scrollview-pager-spacing-y $scrollview-pager-spacing-x;
```
gyoshev: The same goes for this `margin`. There is a margin declaration in theme-default, which can use these variables.

PR #14, opened 2018-02-12
fix: bootstrap 4 input groups styling

gyoshev:
see https://github.com/telerik/kendo-angular/issues/1305

### DIFF
--- a/packages/default/scss/common/_forms.scss
+++ b/packages/default/scss/common/_forms.scss
@@ -335,6 +335,12 @@
             }
         }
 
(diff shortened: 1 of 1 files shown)


gyoshev:
Merging this since the bootstrap integration is documented only in Angular, and is currently broken on the live site.
Consider updating the [bootstrap integration help topic for jQuery](https://docs.telerik.com/kendo-ui/third-party/using-kendo-with-twitter-bootstrap) to mention that the SASS themes don't need extra work to work with Bootstrap 4.

PR #16, opened 2018-02-13
fix(dialog): improve close button alignment

mvgmvg:
related https://github.com/telerik/kendo/issues/8156

### DIFF
--- a/packages/default/scss/window/_layout.scss
+++ b/packages/default/scss/window/_layout.scss
@@ -47,8 +47,7 @@
 
     // Actions
     .k-window-actions {
-        margin-top: add-two(-$button-padding-y, -$button-border-width);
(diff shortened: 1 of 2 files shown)


packages/default/scss/window/_layout.scss
```diff
@@ -47,8 +47,7 @@
 
     // Actions
     .k-window-actions {
-        margin-top: add-two(-$button-padding-y, -$button-border-width);
-        margin-bottom: add-two(-$button-padding-y, -$button-border-width);
+        margin: add-two(-$button-padding-y, -$button-border-width);
```
gyoshev: Shouldn't horizontal margins be connected to `$button-padding-x` (instead of `-y`), or am I missing something?

mvgmvg: Button icon is square and uses $button-padding-y for the both directions.

PR #28, opened 2018-02-21
Split common

joneff:
Split common into several files.


packages/default/scss/common/_typography.scss
```diff
@@ -0,0 +1 @@
+@include exports( "common/typography" ) {}
```
gyoshev: I'm still puzzled by placeholders. Why not add the file when it is needed?

packages/default/scss/common/utils/_flex.scss
```diff
@@ -0,0 +1,46 @@
+@include exports( "common/utils/flex" ) {
```
gyoshev: Which component uses these utility classes?

joneff: For both questions the answer is the upcoming chat and card components

packages/default/scss/common/_typography.scss
joneff: I will need it for the card / chat components.

packages/default/scss/common/utils/_flex.scss
gyoshev: Sounds good. In that case I suggest adding these classes along with the PR for the chat and card components, along with the example usage, and using a `feat:` commit to notify customers of the new theme features.

joneff: Done

Issue #26, opened 2018-02-21
Implement Material for Button - Kendo UI


mvgmvg:
related https://github.com/telerik/kendo/issues/8250

Dimitar-Goshev:
- [ ] Displaced icons of "k-button k-button-icon k-outline" and "k-button k-button-icon k-outline k-primary" in IE:
![image](https://user-images.githubusercontent.com/6614183/38369078-914cd9e0-38ef-11e8-9600-ce4c5a621a99.png)

Dimitar-Goshev:
- [ ] Displaced button icon, if RTL is activated with dir="rtl", instead of class="k-rtl":
![image](https://user-images.githubusercontent.com/6614183/38621044-b08d6e60-3da8-11e8-93d5-9311078f39b3.png)

PR #30, opened 2018-02-22
feat(button): add styles for button in Material

mvgmvg:
related https://github.com/telerik/kendo-themes/issues/26


packages/material/scss/_variables.scss
```diff
@@ -242,11 +242,11 @@ $button-shadow: 0 3px 1px -2px rgba(0, 0, 0, .2), 0 2px 2px 0 rgba(0, 0, 0, .14)
 
 /// The text color of hovered buttons.
 /// @group buttons
-$button-hovered-text: $hovered-text !default;
+$button-hovered-text: $button-text !default;
 
 /// The background of hovered buttons.
 /// @group buttons
-$button-hovered-bg: rgba(0, 0, 0, .12) !default;
+$button-hovered-bg: mix( #000, $button-bg, 12%) !default;
```
gyoshev: :+1: on binding this to `$button-bg`

packages/material/scss/button/_theme.scss
```diff
@@ -23,13 +25,23 @@
...
-        &:not(.k-no-focus).k-state-focused {
-            background-color: $button-focused-bg;
+        &:focus,
+        &.k-state-focused {
             @include appearance( focused-button );
+            background-color: $button-focused-bg;
+        }
+        &.k-no-focus {
+            @include appearance( button );
+            box-shadow: $button-shadow;
```
gyoshev: I think that the ripple effect should hide the focused background, too (which comes from the `:focus` class on line 40). This is how Angular Material behaves, but that leaves the button without a focused state when the ripple effect finishes. Thoughts?

packages/material/scss/button/_theme.scss
```diff
@@ -23,13 +25,23 @@
...
+
+        // Hovered state
+        &:hover,
+        &.k-state-hover {
+            @include appearance( hovered-button );
+        }
 
         // Focused state
-        &:not(.k-no-focus):focus,
-        &:not(.k-no-focus).k-state-focused {
```
gyoshev: Yay!

packages/material/scss/button/_theme.scss
```diff
@@ -148,18 +118,20 @@
...
-            color: $accent;
-            background-color: $flat-primary-active-bg;
+        &:focus,
+        &.k-state-focused {
+            background-color: $flat-primary-focused-bg;
+        }
+        &.k-no-focus {
+            background: transparent;
```
gyoshev: Perhaps `background-color` will suffice here.

packages/material/scss/button/_theme.scss
```diff
@@ -25,18 +27,47 @@
         font-weight: 500;
         transition: box-shadow 280ms cubic-bezier(.4, 0, .2, 1);
 
+        &::before {
```
gyoshev: I'd suggest moving the structural styles to the `_layout` file, if appropriate.

packages/default/scss/button/_layout.scss
```diff
@@ -354,3 +276,69 @@
...
+
+        // Focused state
+        &:focus,
+        &.k-state-focused {
+            &::before {
+                opacity: $flat-button-focused-opacity;
+            }
+        }
+        &.k-no-focus:not(:hover),
+        &.k-no-focus:not(.k-state-hover) {
```
joneff: This will not work as expected, because `.k-state-hover` will be positive for the first selector and `:hover` for the second... In other words, it will catch all selectors...

packages/default/scss/button/_layout.scss
```diff
@@ -354,3 +276,69 @@
...
 
 }
+
+
+
+@include exports( "button/layout/thin-background" ) {
+    $flat-button-hover-opacity: .08 !default;
+    $flat-button-focused-opacity: .12 !default;
+    $flat-button-active-opacity: .16 !default;
+    $flat-button-selected-opacity: .20 !default;
```
joneff: Moved the entire "flat" thing in a separate "module" with its own variables

packages/default/scss/button/_layout.scss
```diff
@@ -25,11 +24,15 @@
...
-        &.k-state-disabled:focus,
         [disabled] & {
             @extend %disabled;
         }
+        .k-ie9 &.k-state-disabled:hover,
+        .k-ie9 &.k-state-disabled:focus,
+        .k-ie10 &.k-state-disabled:hover,
+        .k-ie10 &.k-state-disabled:focus {
+            @extend %disabled;
+        }
```
joneff: I split this, because IE 9 and 10 do not support `pointer-events`. I guess we could rewrite it with `.k-no-pointer-events &.kstate-disabled...` at some point.

packages/default/scss/button/_layout.scss
```diff
@@ -211,13 +208,18 @@
...
-        background-color: transparent;
-        background-image: none;
-        box-shadow: none;
+        background: none !important; // sass-lint:disable-line no-important
+        box-shadow: none !important; // sass-lint:disable-line no-important
+        transition: color .2s ease-in-out;
+
+        &::before {
+            display: block;
+        }
```
joneff: Assuming we always want flat / bare buttons to behave like such, I think it's safe to `!important` rules.

packages/default/scss/button/_layout.scss
```diff
@@ -354,3 +276,69 @@
...
+@include exports( "button/layout/thin-background" ) {
+    $flat-button-hover-opacity: .08 !default;
+    $flat-button-focused-opacity: .12 !default;
+    $flat-button-active-opacity: .16 !default;
+    $flat-button-selected-opacity: .20 !default;
+
+    // Thin background
+    .k-button {
+
+        &::before {
```
joneff: Always "create" the flat / bare background, but make it available when needed.

packages/material/scss/button/_layout.scss
```diff
@@ -1,36 +1,41 @@
...
-            .k-button ~ .k-button {
-                margin-left: -$outline-button-border-width;
-            }
+    // Outline button
+    .k-button.k-outline {
+        padding-top: calc( #{$button-padding-y} - #{$outline-button-border-width} );
+        padding-bottom: calc( #{$button-padding-y} - #{$outline-button-border-width} );
+        padding-left: calc( #{$button-padding-x} - #{$outline-button-border-width} );
+        padding-right: calc( #{$button-padding-x} - #{$outline-button-border-width} );
+        border: $outline-button-border-width solid currentColor;
```
joneff: * Assuming we always want the outline border to behave as such.
* use calculated padding values

gyoshev: That's great, and I believe it fixes a a bug in the calculation (`$button-padding-y / 2` was previously expected to equal `$outline-button-border-width`, if I understand correctly).

I suggest using `add-two()`, as it's cleaner and may remove the `calc()` expression.

packages/material/scss/button/_layout.scss
```diff
@@ -1,36 +1,41 @@
...
+    }
+
+
+    // Outline button group
+    .k-button-group-outline {
+        .k-button ~ .k-button {
+            border-left-width: 0;
```
gyoshev: I think the approach that used `margin-left` / `-right` has a benefit that we can raise the `z-index` of the hovered / active button. This way changing the `currentColor` of these states will show the full border of the button. Does this make sense?

packages/default/scss/button/_layout.scss
gyoshev: Are these required? While this is okay for most scenarios, it will be very hard to override, if we or our customers ever need to.

packages/material/scss/button/_layout.scss
joneff: It was like that.

However, in disabled groups, Material guidelines require translucent color. Having two opacity stacked on top each other gives a nasty visual artifacts.

I am OK with reverting the behaviour. It yields cleaner code.

packages/material/scss/button/_layout.scss
joneff: Prior adding postcss-calc as dependency, I would use add-two.

Now that we have it, we can safely rely, as it will optimize the expression if the values are using the same units.

packages/default/scss/button/_layout.scss
joneff: IMO, yes. If we agree that a flat button is one that behaves a certain way, then that's the way to go.

Alternatively, we could move the styles in `button/theme` for each theme. It will not change the output, per say, but will allow for a developer, be it us or a customer, to tweak each theme individually, should there is a need.

packages/material/scss/button/_layout.scss
gyoshev: I suggest using `add-two` because of the less symbols involved. Plus, the themes may be compiled without postcss-calc in customer applications (angular-cli, react-cli)

packages/default/scss/button/_layout.scss
gyoshev: I still feel a bit uneasy for using `!important` on buttons, but let's see how far it goes. If this obstructs functionality, we can split this to the theme files.

joneff:
@gyoshev I've fixed the issues.

PR #34, opened 2018-02-23
docs: add section to toc, add minor updates

dmihaylo:


README.md
```diff
@@ -40,44 +41,31 @@ During development, the SCSS files are linted on every `commit` and built on eve
...
-# installs theme dependencies and cross-links theme packages
-npx lerna bootstrap
-```
+  > The `npx` command runs packages from `./node_modules/.bin` without requiring a global install.
 
-> `npx` runs packages from `./node_modules/.bin`, without requiring a global install
+1. After running these commands and in the root of the repository:
+  1. Run `npx lerna run lint` to run the linting over all the themes.
+  1. Run `npx lerna run build` to build all the themes.
+  1. Run `npm test` to run the tests (lint, JS tests, and build).
```
gyoshev: These commands are not part of the setup, but a list of helpful commands that can be run in the root of the repo. Perhaps this should be outside the setup instructions.

Issue #47, opened 2018-02-28
Implement Material for Splitter - Kendo UI


mvgmvg:
related telerik/kendo#8250

Dimitar-Goshev:
@mvgmvg Please review the comments on the [Tooltip pull request](https://github.com/telerik/kendo-themes/pull/50) and resolve any issues there. Note that the Splitter styles went there unexpectedly.

PR #48, opened 2018-02-28
feat(splitter): add Material skin for Splitter

mvgmvg:
related https://github.com/telerik/kendo-themes/issues/47

### DIFF
--- a/packages/material/scss/splitter/_theme.scss
+++ b/packages/material/scss/splitter/_theme.scss
@@ -1,0 +1,38 @@
+$splitter-bg: rgba($base-text, .12);
+$splitter-hovered-bg: rgba($base-text, .22);
+
+@include exports("splitter/theme") {
+    .k-splitbar {
+        color: $base-text;
+        background: $splitter-bg;
+
+        & > .k-icon {
+            opacity: .54;
+        }
+    }
+
+    // Ghost
+    .k-ghost-splitbar {
+        background-color: $splitter-bg;
+    }
+
+    // Hover state
+    .k-splitbar-horizontal-hover,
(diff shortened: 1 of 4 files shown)


mvgmvg:
As all Material icons uses opacity, I thing that this approach is more consistent with other scenarios with icons. It cannot be mixed, as its parent background (splitbar) has opacity (rgba) as well. Colors used in the mix() function can be only rgb. In this case we are using only 2 splitter variables for Material theme. If we inherit Default theme, we will have four redundant variables.

gyoshev:
Is there a problem with using `rgba()` for the icon colors in order to preserve the opacity?

As for the amount of variables -- they come for free, and allow developers to change the styling of specific interactions with the splitter. You can link the variables that are not part of the default styling.

Reusing existing styles allows us to make changes to one file to fix problems, rather than going through all themes to fix them.

mvgmvg:
You are right, using rgba instead opacity can resolve the issues.

packages/material/scss/splitter/_theme.scss
```diff
@@ -0,0 +1,10 @@
+$splitter-text: rgba($base-text, .54) !default;
+$splitter-bg: rgba($base-text, .12) !default;
+$splitter-hovered-text: $base-text !default;
+$splitter-hovered-bg: rgba($base-text, .22);
+$splitter-selected-text: $selected-text !default;
+$splitter-selected-bg: $selected-bg !default;
+
+@import "~@progress/kendo-theme-default/scss/splitter/theme";
+
+@include exports("splitter/theme") {}
```
gyoshev: This line is not needed, as the module is inherited from the default theme.

Issue #49, opened 2018-03-01
Implement Material for Tooltip - Kendo UI


mvgmvg:
related telerik/kendo#8250

Guidelines:
https://material.io/guidelines/components/tooltips.html#tooltips-interaction

Dimitar-Goshev:
@mvgmvg Please review the comments on the pull request and resolve any issues there. Also note that the Splitter styles went there unexpectedly.

Dimitar-Goshev:
- [ ] Closable tooltip is higher than a non-closable. Also its text is bigger and the close icon is displaced:
![image](https://user-images.githubusercontent.com/6614183/37467318-8f74b51c-2868-11e8-9e7a-e9ea4a160d99.png)

- [ ] Ajax and Content template demos - tooltips flicker, as their content goes out of the tooltip and may be hovered. The issue is reproducible with Bootstrap v4 as well - https://demos.telerik.com/kendo-ui/tooltip/ajax and https://demos.telerik.com/kendo-ui/tooltip/template - [video](https://www.screencast.com/t/1Fm1rpxtGT)

Dimitar-Goshev:
- [ ] Tooltip callout is not centered in all themes - it will become a regression, if not fixed for R2:
![image](https://user-images.githubusercontent.com/6614183/39231043-6d4e49fc-4871-11e8-8781-a3daf43e752d.png)

PR #50, opened 2018-03-01
Material/tooltip

mvgmvg:
related https://github.com/telerik/kendo-themes/issues/49


gyoshev:
It seems that this branch contains commits for the splitter component. You can filter them out during rebase to keep the history clean. Alternatively, you can [squash all commits](https://makandracards.com/makandra/527-squash-several-git-commits-into-a-single-commit) into one. Either way, having this many feat() commits will add each of them to the changelog, which is not great.

packages/material/scss/tooltip.scss
```diff
@@ -0,0 +1,13 @@
+@import "variables";
+@import "mixins";
+
+
+// Dependencies
+@import "common/all";
+@import "common/forms";
```
joneff: Do we really need `common/forms`?

packages/material/scss/tooltip/_theme.scss
```diff
@@ -0,0 +1,11 @@
+$tooltip-color: $base-bg !default;
+$tooltip-bg: rgba(lighten($base-text, 38), .9) !default;
+
+@import "~@progress/kendo-theme-default/scss/tooltip/theme";
+
+@include exports("tooltip/theme/material") {
+
+    .k-callout {
+        border-color: transparent;
```
joneff: If the idea is to have no callout, why not hide it e.g.:

```scss
// in _layout.scss, right after the import

@include exports("tooltip/layout/material") {
    .k-callout { display: none; }
}
```

Another way, would be to have a flag for that. @gyoshev what's your thought on the flag?

packages/material/scss/tooltip/_theme.scss
```diff
@@ -0,0 +1,11 @@
+$tooltip-color: $base-bg !default;
```
joneff: According to https://material.angular.io/components/tooltip/examples and https://material.io/guidelines/components/tooltips.html, tooltips are always white text on dark background (gray-700).

packages/material/scss/tooltip/_theme.scss
gyoshev: I don't think it is wise to always hide the callout. Both the jQuery and Angular tooltips have a `callout` configuration option, and adding either rules will disable the effect that the property has, even if customers use `callout: true`, which is frustrating.

mvgmvg:
The commented issues are fixed.

PR #54, opened 2018-03-02
WIP: Visual tests

gyoshev:
This PR adds visual tests for the components as a final build step, as discussed in telerik/kendo-theme-default#720.

Should we use a single set of fixtures for both the karma and visual tests?

WIP: I would like to <strike>reduce the size of the screenshots and</strike> make the build predictable across platforms before merging this.


joneff:
@gyoshev I've added more tests for buttons, as well as tests for dropdowns, inputs (sans slider and switch) and dateinputs (sans calendar)

gyoshev:
Turns out that having a predictable screenshot across different platforms is a lost cause.

I'll be updating the build so that Travis pushes back the generated images. This way a single platform will be making the screenshots.

tsvetomir:
image

PR #56, opened 2018-03-02
fix(grid): fix filter menu double border

rkonstantinov:
This PR fixes the double border of the filter menu in `default` theme:
image

as well as in `bootstrap` theme:
image

The `bootstrap` requires a bit more overrides, which seems not used in the `default` theme.


rkonstantinov:
I'm closing this PR as we will address it in the grid component itself.

PR #59, opened 2018-03-06
WIP: fix(tooltip): add styles for Angular

elena-gancheva:
Tooltip styles in progress. Needs to be linked with https://github.com/telerik/kendo-angular-tooltip/pull/1 for properly styling


gyoshev:
Some implementation notes:

- We need `k-popup-transparent` in order to hide the popup container styling, as we cannot have the same rendering in Angular (as we do in jQuery). The cause of this is that in the jQuery widgets, the popup functionality is added to an existing element (`div.k-tooltip.k-popup`) and in Angular, it wraps a template (`div.k-popup > div.k-tooltip`). The `-transparent` class reverts the styling for drop-down popups.
- `k-animation-container` gets `overflow: visible` through JS in the jQuery widgets. The `k-animation-container-shown` rule removes this requirement, allowing both the tooltip and the menu to function properly.
- The callout wasn't centered properly, thus we needed the `margin-` rules. This is not visible in the jQuery widgets, as it is positioned with JS (I believe this has to do with the screen boundary detection).

gyoshev:
@joneff I second that! This is being improved upon in https://github.com/telerik/kendo-themes/pull/74.

PR #64, opened 2018-03-13
fix(button): update spacing of button with icon

gyoshev:
This seems a bit controversial, due to the math. I'm open for ideas.

See https://github.com/telerik/kendo-angular-buttons/issues/199

### DIFF
--- a/packages/default/scss/button/_layout.scss
+++ b/packages/default/scss/button/_layout.scss
@@ -58,8 +58,7 @@
             .k-icon,
             .k-image,
             .k-sprite {
-                margin: 0 ($button-padding-x / 2) 0 (-$button-padding-x / 2);
-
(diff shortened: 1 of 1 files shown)


packages/default/scss/button/_layout.scss
```diff
@@ -55,22 +55,25 @@
         &-icontext {
             overflow: visible; // IE9
 
+            $icon-spacing-before: -$button-padding-x / 4;
+            $icon-spacing-after: $button-padding-x - $icon-spacing-before;
```
joneff: There is a theme variable called `$icon-spacing`, we could use that one or extend it. Also, according to design, $icon-spacing-after should be 4px in default theme.

We could perhaps switch to direction independent names such as `$icon-spacing-start` / `$icon-spacing-end`.

gyoshev:
Decided to drop this change and close the issue. The improvement is marginal, and it will change the layout without a delivering much value.

Issue #45, opened 2018-02-28
Implement Material for Multiselect - Kendo UI


mvgmvg:
related telerik/kendo#8250

Dimitar-Goshev:
- [ ] Hovered token does not change its background
- [ ] Margins around token Close button are incorrect. By design they are left 6px and right 4px, but it is implemented as left 7px and right 6px. Thus there is too much space to the right between the button and the right side of the token, compared to above and below the button:
![image](https://user-images.githubusercontent.com/6614183/37157625-01411374-22f2-11e8-852a-754db1a1d226.png)
- [ ] Bottom border width should become 2px on hover
- [ ] Bottom border of disabled ComboBox is not dotted
- [ ] k-state-invalid does not change the border color
- [ ] No focus indicator for focused dropdown item

mvgmvg:
Waiting for design for grouping implementation to finish the issue.

Dimitar-Goshev:
- [ ] Grouping header is not all caps and it should be higher than other items - design:
![image](https://user-images.githubusercontent.com/6614183/38561289-55e040dc-3ce0-11e8-9055-2e7c3655e19a.png)
- [ ] LTR - Loading icon does not cover the clear icon completely. It is possible to click the clear icon. (Use Server filtering demo -> stop internet in Chrome -> type a in input:
![image](https://user-images.githubusercontent.com/6614183/38563591-e842a078-3ce5-11e8-9ef2-c8fc76e0b8ed.png)

- [ ] RTL - Loading icon does not cover the clear icon (x):
![image](https://user-images.githubusercontent.com/6614183/38563168-e451f9ec-3ce4-11e8-9dda-a48bf73545ca.png)

- [ ] RTL - Wrong horizontal margins between tokens, wrong vertical margins (RTL and LTR), displaced delete circle:
![image](https://user-images.githubusercontent.com/6614183/38563304-2bb7d248-3ce5-11e8-98e4-bec5015d5731.png)

Dimitar-Goshev:
- [ ] Wrong padding of Group header in RTL:
![image](https://user-images.githubusercontent.com/6614183/39121496-7e4fccda-46fa-11e8-90e1-6e877d2543a3.png)

NOTE: Not reproducible in all cases, e.g. on the demo it is reproducible, if you type a, but not reproducible, if you type z.

mvgmvg:
[...]

PR #73, opened 2018-03-15
Material/dropdowns

mvgmvg:
related https://github.com/telerik/kendo-themes/issues/45


packages/default/scss/multiselect/_layout.scss
```diff
@@ -30,7 +30,8 @@
 
                 .k-button {
                     width: auto;
-                    margin: $padding-y-sm $padding-y-sm 0 0;
+                    margin-right: $padding-y-sm;
+                    margin-left: 0;
```
gyoshev: How does this look when the selected items (.k-button) wrap on a second line? Add a visual test for that.

packages/material/scss/autocomplete/_layout.scss
```diff
@@ -1 +1,7 @@
 @import "~@progress/kendo-theme-default/scss/autocomplete/layout";
+
+@include exports('autocomplete/layout/material') {
+    .k-autocomplete {
+        border-radius: 0;
```
gyoshev: Introduce a `$autocomplete-border-radius` in theme-default and use it to reset it to zero here.

packages/material/scss/common/_forms.scss
```diff
@@ -56,6 +57,11 @@
             .k-select {
                 color: $icon-hovered-color;
             }
+
+            &::after {
+                transform: scale(1);
+                opacity: 1;
```
gyoshev: What are these styles used for? It seems these are for the input ripple effects, which are achieved through the textbox container, and have no place in the theme.

packages/default/scss/multiselect/_layout.scss
mvgmvg: Those changes are applied only for RTL mode. The idea is to update only horizontal margins for RTL mode. The vertical should be the same.

packages/material/scss/autocomplete/_layout.scss
mvgmvg: the variable is added

packages/material/scss/common/_forms.scss
mvgmvg: Those styles add ripple effect for all dropdowns and picker without a necessity of using javascript.

packages/material/scss/multiselect/_theme.scss
```diff
@@ -52,4 +51,13 @@
         }
     }
 
+    // Fix vertical centering icon
+    .k-multiselect .k-button .k-icon.k-i-arrow-60-down {
+        line-height: 1.2;
```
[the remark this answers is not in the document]
mvgmvg: Even though the arrow is vertically aligned, it does not look such in Chrome and Firefox. The token icon is 12 pixels height and the triangle is 5 pixels. The remaining pixels are divided into two parts - 3 and 4 pixels. Both browsers place 3 pixels above the triangle and 4 below it. For that reason, the icon looks not aligned. The line-height 1.2 is not calculated, using this value the icon looks well.

packages/material/scss/input/_theme.scss
```diff
@@ -28,10 +33,13 @@
...
-    // Textbox
-    .k-textbox,
-    .k-input.k-textbox {
-        border-color: $input-border;
+    .k-textbox {
+
+        &:hover,
+        &:focus {
+            padding-bottom: add-two($input-padding-y, -$input-border-width);
+            border-bottom-width: $input-border-width * 2;
```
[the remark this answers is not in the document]
mvgmvg: Styles for ripple effect were removed.

packages/material/scss/multiselect/_theme.scss
```diff
@@ -21,6 +21,7 @@
             min-height: $form-line-height-em;
             margin: $input-padding-y;
             margin-left: 0;
+            // padding: 0 ($button-padding-x / 3);
```
gyoshev: No need to commit commented out code.

packages/material/scss/common/_forms.scss
```diff
@@ -94,4 +86,34 @@
...
+    .k-state-invalid,
+    .ng-invalid.ng-touched,
+    .ng-invalid.ng-dirty {
+
+        &.k-autocomplete,
+        > .k-dateinput-wrap,
+        > .k-dropdown-wrap,
+        > .k-picker-wrap,
+        > .k-multiselect-wrap,
+        > .k-numeric-wrap {
```
gyoshev: This rule will create a *lot* of selectors (90, to be precise, because of the nested classes, 3 * 6 * 5).
Looking at the screenshots, they show some improvement with the MultiSelect, but the amount of styles required for that is too much, and the styles have no effect on most components. Perhaps there is some detail that is missing from the screenshots?

packages/material/scss/multiselect/_theme.scss
mvgmvg: the redundant line was removed

packages/material/scss/common/_forms.scss
mvgmvg: As .k-state-invalid should be applied to .k-widget, the selectors should be correct. I have fixed the tested page moving the class name to the correct element.

packages/default/scss/common/_forms.scss
```diff
@@ -263,15 +263,8 @@ $input-border-radius: $border-radius !default;
...
         &.k-maskedtextbox > .k-textbox {
-
-            &,
-            &.k-state-hover,
-            &:hover,
-            &.k-state-focused,
-            &:focus {
-                color: $error;
-                border-color: $invalid-border;
-            }
```
mvgmvg: After those changes, the red invalid border is changed when a widget is hovered/selected. It should be always red.

packages/default/scss/common/_forms.scss
```diff
@@ -280,18 +273,14 @@ $input-border-radius: $border-radius !default;
         > .k-numeric-wrap,
         > .k-picker-wrap,
         > .k-multiselect-wrap {
```
mvgmvg: After those changes, the red invalid border is changed when a widget is hovered/selected. It should be always red.

joneff: I tweaked the selector slightly, so it doesn't happen

PR #74, opened 2018-03-15
feat(tooltip): improve tooltip close button position

mvgmvg:
related https://github.com/telerik/kendo-themes/issues/49


packages/material/scss/tooltip/_layout.scss
```diff
@@ -1 +1,14 @@
 @import "~@progress/kendo-theme-default/scss/tooltip/layout";
+
+@include exports("tooltip/layout/material") {
+    .k-tooltip-button .k-icon {
+        margin-right: -$padding-x;
```
gyoshev: These styles try to undo the [large padding on the closable tooltip](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/tooltip/_layout.scss#L18) in the default theme. Looking at the visual tests, this padding seems inadequate for the Material theme. It would be better to add variables for this padding and override them before importing the default theme, and remove these rules.

mvgmvg: In Material guidelines is missing scenario for closable tooltip. We added bigger padding in Default theme for such case. As for this moment we do not have design for the scenario, I have logged it as a UX issue [here](https://github.com/telerik/kendo/issues/7946)

gyoshev: The current values look right for the default and bootstrap themes. These values can be the default for the variable:

```scss
// variables in default theme
$tooltip-padding-x: $padding-x * 2 !default;
$tooltip-padding-y: $padding-y * 3 !default;
```

These can be overridden in the material theme:

```scss
$tooltip-padding-x: $padding-x !default;

@import "~@progress/kendo-theme-default/scss/tooltip/layout"; 
```

This way the additional `-$padding-x` margin won't be needed, as the padding will be initially smaller. It is likely that this will yield a pretty decent design, which is a good starting point.

mvgmvg: $tooltip-padding-x variables are used for no closable tooltips. Closable tooltips should have larger padding. In the Default theme, such paddings are two and three times larger. If paddings in Material closable tooltip are $padding-x, then such tooltip will have the same padding as no closable tooltip, what is not expected. To specify padding for the closable tooltip, we need a design.

gyoshev: My bad, the variables should have been `$tooltip-closable-padding-x / y`. My argument still stands.

mvgmvg: we came to the decision with @gyoshev to wait for design for the closable tooltip. After that, we will add a variable for its padding.

PR #76, opened 2018-03-20
feat: add visual tests for scrollview

elena-gancheva:


packages/material/scss/scrollview/_layout.scss
```diff
@@ -1 +1,11 @@
 @import "~@progress/kendo-theme-default/scss/scrollview/layout";
+
+@include exports('scrollview/layout') {
```
gyoshev: The same module is defined in `~@progress/kendo-theme-default/scss/scrollview/layout`. This means the code in this block is never included, because the names are the same.

packages/default/scss/scrollview/_layout.scss
```diff
@@ -46,12 +48,12 @@
...
         position: absolute;
         left: 0;
-        bottom: $scrollview-spacing;
+        bottom: $scrollview-pager-spacing-y;
         pointer-events: none;
     }
     .k-pages > li,
     .k-scrollview-pageable > li.k-button {
-        margin: $scrollview-spacing;
+        margin: $scrollview-pager-spacing-y;
```
gyoshev: If this is changed to `margin: $scrollview-pager-spacing-y $scrollview-pager-spacing-x`, you can delete the line in the material layout file. Bonus: this will start using the `-x` variable 😉

Issue #71, opened 2018-03-14
Implement Material for Calendar


theOrlin:

theOrlin:
Waiting for design for prev/next months in month view, hover states to complete.

PR #80, opened 2018-03-21
Material calendar styles

theOrlin:
Dealing with https://github.com/telerik/kendo-themes/issues/71


gyoshev:
image

The difference in calendar height between views will produce an unpleasant effect when navigating between the month and year views in the jQuery calendar. The current styles match the design, but they both should be changed to prevent this, and this should make a great integrity test.

PR #81, opened 2018-03-21
fix(datetimepicker): remove button borders

theOrlin:


gyoshev:
Great! Looks much better. We should strive to implement the above approach to reduce the amount of overrides in the themes.

Issue #78, opened 2018-03-20
Implement Material for Menu- Kendo UI


mvgmvg:
related telerik/kendo#8250

Dimitar-Goshev:
- [ ] No padding around separators - <li class="k-separator> (image below)
- [ ] No padding before first and after last item of Sub Sub items
Used Keyboard Navigation demo for easier setup - [snippet](http://dojo.telerik.com/OxIcazuG).
![image](https://user-images.githubusercontent.com/6614183/38422775-9f9056f6-39b4-11e8-9904-45e3d8dd886e.png)
- [ ] Wrong height of scrollable menu submenu items - also reproducible for Default and Bootstrap:
![image](https://user-images.githubusercontent.com/6614183/38423007-542bf78c-39b5-11e8-8037-52d708d31f81.png)

gyoshev:
The last two items should be resolved with an added rendering change from https://github.com/telerik/kendo-themes/commit/833debeb592a761e47a3e15a17d45ccd803ebde4. I've [added a PR](https://github.com/telerik/kendo/pull/8380) that should resolve this in KUI for jQuery.

PR #82, opened 2018-03-22
Material/menu

mvgmvg:
related https://github.com/telerik/kendo-themes/issues/78
related https://github.com/telerik/kendo-themes/issues/83


packages/material/scss/menu/_theme.scss
```diff
@@ -28,8 +31,6 @@ $nav-selected-text: $base-text !default;
     .k-menu-scroll-wrapper {
 
         .k-menu-scroll-button {
-            color: inherit;
```
mvgmvg: This rule is necessary to prevent changing color by hover in Material

packages/default/scss/menu/_theme.scss
```diff
@@ -1,13 +1,20 @@
...
+$menu-selected-text: $list-item-selected-text !default;
+$menu-focused-shadow: $list-item-focused-shadow !default;
+
+$menu-scrolling-bg: $bg-color !default;
+$menu-scrolling-border: $widget-border !default;
+$menu-scrolling-icon: $link-text !default;
+$menu-scrolling-icon-hover: $link-hover-text !default;
+
 @include exports("menu/theme") {
-    $nav-text: $link-text !default;
```
mvgmvg: The same variables can be re-used for tabstrip, so I think that the $nav- variable is more appropriate.

gyoshev: If so, they are an abstraction that should be declared in the `_variables` file, in order to be shared between components. After that, they should be propagated to the specific `$menu-*` variables.

That said, I'm not sure if such an abstraction is necessary at that point. It may be, if the variable values are the same across themes.

packages/material/scss/menu/_theme.scss
gyoshev: This exact property is now generated by theme-default, thanks to `$menu-scrolling-icon: inherit`. We can set `$menu-scrolling-icon-hover: inherit` to remove the hover effect, without the need for overrides.

Issue #85, opened 2018-03-26
 Implement Material for Window - Kendo UI


mvgmvg:
related telerik/kendo#8250

Issue #31, opened 2018-02-22
Implement Material for Dialog - Kendo UI


mvgmvg:
related telerik/kendo#8250

Dimitar-Goshev:
- [ ] Scrollable content has no top and bottom border:
![image](https://user-images.githubusercontent.com/6614183/38371157-6a3dd250-38f4-11e8-81c4-145d10a1679b.png)

- [ ] Long button text overflows dialog. Should buttons fit on only one row and their excess text be with ... ?
![image](https://user-images.githubusercontent.com/6614183/38371261-a8d3563e-38f4-11e8-829b-50b1180efaac.png)

PR #86, opened 2018-03-26
Material/dialog

mvgmvg:
related https://github.com/telerik/kendo-themes/issues/31
https://github.com/telerik/kendo-themes/issues/85


packages/material/scss/dialog/_layout.scss
```diff
@@ -1,27 +1,50 @@
...
-            border-radius: $border-radius;
-            flex: 0 1 auto;
-            align-self: center;
-            padding: 0 $padding-x;
-            border-width: 0;
-            transition: none;
-
-            &:first-child {
-                margin-right: $window-content-padding;
+            @extend .k-button.k-flat;
```
joneff: After the material/button branch, we can now `@extend` the selectors to have flat buttons

packages/material/scss/window/_layout.scss
```diff
@@ -1 +1,34 @@
...
+@include exports("window/layout/material") {
+
+    $window-title-font-size: 20px !default;
+    $window-title-line-height: 1.4 !default;
+
+    .k-window-titlebar {
+        border-width: 0;
+    }
+    .k-window-title {
+        margin: -.2em 0;
```
joneff: In the material guidelines, padding refers to the space between the edge of the rendered and surrounding containers. In essence, it's like having line-height 1, but with the minor difference that when title is multi-line, the line height is 28px.

Thus, I've resorted to some trickery to almost get the appearance right.

packages/bootstrap/scss/dialog/_layout.scss
```diff
@@ -1 +1,35 @@
-@import "~@progress/kendo-theme-default/scss/dialog/layout";
\ No newline at end of file
+@import "~@progress/kendo-theme-default/scss/dialog/layout";
+
+@include exports("dialog/layout/bootstrap") {
+
+    .k-dialog-buttongroup {
+        padding: $window-inner-padding-y $window-inner-padding-x;
```
joneff: As per bootstrap modal

packages/default/scss/dialog/_layout.scss
```diff
@@ -51,54 +51,21 @@
 
     // Buttons
     .k-dialog-buttongroup {
-        border-width: 1px 0 0;
+        border-width: 0;
```
joneff: Not sure if we must have top border 1px by default and remove it where we don't need it, or have 0 and add it, where we need it.

@mvgmvg @gyoshev any thoughts on this?

gyoshev: IMO we should have a variable with a default value that matches the Default theme, and override it before including the layout file in derived themes, and not include border sizing in the `_theme` file.

joneff: I will amend the commit

packages/default/scss/dialog/_layout.scss
```diff
@@ -54,51 +54,18 @@
...
-        background-color: transparent;
         display: flex;
+        flex-direction: row;
+        flex-wrap: wrap;
+        align-items: center;
         flex: 0 0 auto;
 
         .k-button {
-            // sass-lint:disable no-important
+            width: auto !important; // sass-lint:disable-line no-important
```
gyoshev: Please document uses of `!important` with an inline comment. There should be a reason why are we prohibiting customers from setting the button width. While this rule was not added in this PR, it is a good time to explain why we need this. I personally don't recall, and future maintainers won't know, too.

packages/default/scss/dialog/_theme.scss
```diff
@@ -3,6 +3,7 @@
     $dialog-titlebar-border-color: inherit !default;
     $dialog-titlebar-bg: $accent !default;
     $dialog-titlebar-text: $accent-contrast !default;
+    $hover-active-selector: ":hover, .k-state-hover, :active, .k-state-active";
```
gyoshev: I don't think that's a pattern that we should use. It hides away the selector complexity.

Also, [according to MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/:not#Browser_compatibility), the `:not(.foo, .bar)` is supported only in Safari 9+ at the moment, so probably we don't need these rules?

packages/material/scss/window/_theme.scss
```diff
@@ -1,32 +1,24 @@
 $window-content-padding: $padding-y * 2 !default;
 
 @include exports("window/theme") {
-    $window-button-width: $window-content-padding * 3 !default;
```
gyoshev: :tada:

packages/default/scss/dialog/_layout.scss
mvgmvg: I have added an explaining comment. The reason is to prevent buttons to stretch when "stretched button layout" is set as Material guidelines do not support such appearance.

Issue #93, opened 2018-03-28
Implement Material for ListBox


theOrlin:

matanasov:
Some issues found:
1. The items paddings (left and right) should be 24px, but currently are 16.
2. When an item is dragged-dropped, it doesn't have correct background color set (see dragged item in the design)

theOrlin:
Design has been updated, and the dragged item now is a selected item with opacity, with no background.

theOrlin:
Regarding **3**:

The colors are inherited from the Popup styling. After consulting with @mvgmvg, I was informed that the hovered item text color fix is currently in a separate branch, to be merged into develop at a future time.

joneff:
@matanasov I am closing this item, as we have already released ListBox as part of material theme.

Please double check if there are any outstanding issues, including the ones you've already mentioned and if  needed, open a new issue.

PR #94, opened 2018-03-28
Material listbox

theOrlin:
Dealing with https://github.com/telerik/kendo-themes/issues/93


packages/material/scss/listbox/_theme.scss
```diff
@@ -0,0 +1,26 @@
...
+@include exports( 'listbox/theme/material') {
+    $listbox-item-focused-bg: rgba(0, 0, 0, .07) !default;
+    $listbox-drag-clue-bg: $hovered-bg !default;
+
+    .k-listbox {
+        .k-list .k-item {
+            color: $base-text;
+
+            &.k-state-focused {
+                background-color: $listbox-item-focused-bg;
```
gyoshev: In the Default theme, `.k-list .k-item` gets its styles from the `popup/_theme.scss` file, which is inherited by the Material theme. Do we need the additional styles?

theOrlin:
The drag clue's `!important` width is needed because the jQuery widget sets the drag clue width inline, and the Material theme's drag clue does not have a preset width. The rest is a mistake caused by a copy-paste job from the Grid's drag clue override, where all the above properties are set inline.

packages/material/scss/popup/_theme.scss
```diff
@@ -1 +1,17 @@
 @import "~@progress/kendo-theme-default/scss/popup/theme";
+
+@include exports( "list/theme/material") {
```
gyoshev: These styles are not needed. The focused state is added in the material/dropdowns branch, and everything else is already handled by the popup theme.

packages/default/scss/listbox/_layout.scss
```diff
@@ -1,7 +1,12 @@
 @include exports( 'listbox/layout' ) {
     $listbox-margin: $padding-x !default;
+    $listbox-button-margin: $padding-x !default;
```
joneff: By default, `$listbox-button-margin` should be equal to `$listbox-margin`

packages/default/scss/listbox/_theme.scss
```diff
@@ -1,4 +1,7 @@
 @include exports('listbox/theme') {
+    $listbox-drag-clue-bg: $list-item-selected-bg !default;
+    $listbox-text: $list-item-selected-text !default;
```
joneff: The variable name is misleading. I would expect to change the text color of the listbox widget, not the dragclue.

tests/visual/listbox.html
```diff
@@ -0,0 +1,75 @@
...
+                            <span class="k-icon k-i-x"></span>
+                        </a>
+                    </li>
+                </ul>
+            </div>
+            <div class="k-list-scroller k-selectable" data-role="selectable">
+                <ul class="k-reset k-list" role="listbox" tabindex="0">
+                    <li class="k-item" role="option" aria-selected="false" data-uid="6bd99714-38a2-4053-80c4-48494c8f6afb">Item</li>
+                    <li class="k-item k-state-selected" role="option" aria-selected="false" data-uid="19d76c92-35e9-4e0a-9135-f03d49dfe159">Selected</li>
+                    <li class="k-item k-state-focused" role="option" aria-selected="false" data-uid="a5a173ed-8c8e-488c-b297-faa2f68518f0">Focused</li>
```
joneff: The test does not show hovered, hover-selected and disabled item.

Additionally, there is not drag clue.

tests/visual/listbox.html
```diff
@@ -0,0 +1,75 @@
...
+                        <a href="#" class="k-button k-button-icon" data-command="transferAllFrom" title="Transfer All From" aria-label="Transfer All From"
+                            role="button">
+                            <span class="k-icon k-i-arrow-double-60-left"></span>
+                        </a>
+                    </li>
+                    <li>
+                        <a href="#" class="k-button k-button-icon" data-command="remove" title="Delete" aria-label="Delete" role="button">
+                            <span class="k-icon k-i-x"></span>
+                        </a>
+                    </li>
```
joneff: While the buttons do inherit styles, showing hovered, focused, pressed and disabled state will be a nice addition.

PR #96, opened 2018-03-28
feat: add ListView styles

theOrlin:
Dealing with https://github.com/telerik/kendo-themes/issues/95


gyoshev:
It would be great if you add also a test case along with the PR. Also, please mention that this adds the listview for the Material theme in the commit message.

joneff:
@gyoshev, fwiw listview has only 1 relevant style and that's focus. It's a template control and I am not sure we can add tests for it.

Issue #87, opened 2018-03-26
Implement Material for Toolbar - Kendo UI


mvgmvg:
related telerik/kendo#8250

Dimitar-Goshev:
- [ ] Overflow button overlaps last item - reproducible on Resizing demo:
![image](https://user-images.githubusercontent.com/6614183/39181031-0e9dfe94-47c1-11e8-818a-4c44917b21d7.png)
- [ ] No padding between separator and label in template - reproducible on Resizing demo:
![image](https://user-images.githubusercontent.com/6614183/39181106-486c614c-47c1-11e8-829e-2791a5457efa.png)
- [ ] No visible difference between appearance of selected and selected+focused button
- [ ] Separator is as high as a normal item in overflow:
![image](https://user-images.githubusercontent.com/6614183/39183319-45c60248-47c8-11e8-8e30-6d8a7b98f6fc.png)

PR #112, opened 2018-04-05
Material/toolbar

mvgmvg:
related https://github.com/telerik/kendo-themes/issues/87


packages/default/scss/toolbar/_theme.scss
```diff
@@ -1,3 +1,14 @@
+$toolbar-separator-border: inherit !default;
+
+$toolbar-button-expanded-bg: null !default;
+$toolbar-button-expanded-shadow: none !default;
+$toolbar-button-hovered-opacity: .05 !default;
+$toolbar-button-active-opacity: .22 !default;
+$toolbar-button-focused-opacity: 0 !default;
+$toolbar-button-focused-outline-opacity: .08 !default;
+
+$toolbar-button-mode: flat !default;
```
gyoshev: This is probably a good idea, but it would be great if the values for this variable are documented. Looking at the code below these are `flat` and `thin`, and it seems that `thin` means that the default state is invisible, but it would be great if a comment explained this.

packages/material/scss/multiselect/_layout.scss
```diff
@@ -1,2 +1,9 @@
 @import "~@progress/kendo-theme-default/scss/multiselect/layout";
 
+@include exports("multiselect/layout/material") {
+
+    .k-button {
```
gyoshev: This style is not scoped and reverts the `text-transform` from `button/_layout.scss`. It should either be scoped to the multiselect, or both rules should be deleted.

packages/default/scss/toolbar/_theme.scss
```diff
@@ -1,3 +1,19 @@
...
+$toolbar-button-hovered-opacity: .05 !default;
+$toolbar-button-active-opacity: .22 !default;
+$toolbar-button-focused-opacity: 0 !default;
+$toolbar-button-focused-outline-opacity: .08 !default;
+
+/// Toolbar button types
+/// null - default buttons type
+/// thin - flat buttons appearance in the normal state and normal buttons appearance in other states
+/// flat - flat buttons type
+/// outline - outline buttons type
```
gyoshev: I suggest leaving this variable internally documented (`//` instead of `///`) for now.

packages/material/scss/_variables.scss
```diff
@@ -84,13 +84,16 @@ $textarea-floating-label-padding: 25px !default;
...
 /// The vertical padding of the container.
 /// @group toolbar
-$toolbar-padding-y: $padding-x !default;
-$toolbar-calc-size: #{$button-calc-size} + #{$toolbar-padding-y * 2} + 2px;
-$toolbar-inner-calc-size: #{$button-calc-size} + #{$toolbar-padding-y * 2};
+$toolbar-padding-y: 0 !default;
+$toolbar-button-padding-x: $button-padding-x !default;
+$toolbar-button-padding-y: $padding-y*2.5 !default;
+$toolbar-button-calc-size: #{$form-line-height-em} + #{$toolbar-button-padding-y * 2} + #{$button-border-width * 2};
+$toolbar-calc-size: #{$toolbar-button-calc-size} + #{$toolbar-padding-y * 2} !default;
```
gyoshev: This variable is not used anywhere.

mvgmvg: this variable comes from the _variables in the Default theme

mvgmvg:
issues are fixed

mvgmvg:
Please, add comment after this branch is merged in this issue: https://github.com/telerik/kendo-themes/issues/93#issuecomment-378241906

packages/default/scss/_variables.scss
```diff
@@ -104,8 +104,10 @@ $toolbar-padding-x: $padding-x !default;
 /// The vertical padding of the container.
 /// @group toolbar
 $toolbar-padding-y: $padding-x !default;
-$toolbar-calc-size: #{$button-calc-size} + #{$toolbar-padding-y * 2} + 2px;
-$toolbar-inner-calc-size: #{$button-calc-size} + #{$toolbar-padding-y * 2};
+$toolbar-button-padding-x: $button-padding-x !default;
+$toolbar-button-padding-y: $button-padding-y !default;
+$toolbar-button-calc-size: #{$form-line-height-em} + #{$toolbar-button-padding-y * 2} + 2px;
```
gyoshev: The `2px` at the end of the line appear to be the button border width. If so, this means that they can be substituted by the formula `#{$button-border-width * 2}`. This, in turn, makes the math in the Default theme compatible with the Material theme, meaning that there is no need to override the `$toolbar-button-calc-size` and the `$toolbar-inner-calc-size` variables in Material.

packages/bootstrap/scss/_variables.scss
```diff
@@ -138,8 +138,9 @@ $toolbar-padding-x: $table-cell-padding / 2 !default;
 /// @group toolbar
 $toolbar-padding-y: $table-cell-padding / 2 !default;
 
-$toolbar-calc-size: #{$button-calc-size} + #{$toolbar-padding-y * 2} + 2px;
-$toolbar-inner-calc-size: #{$button-calc-size} + #{$toolbar-padding-y * 2};
+$toolbar-button-padding-x: $button-padding-x !default;
+$toolbar-button-padding-y: $button-padding-y !default;
```
gyoshev: These are the default values in the Default theme. No need to re-state them here.

PR #114, opened 2018-04-05
fix(listview): Improve focused and selected state appearance of listview in accordance with design

joneff:
Fixes https://github.com/telerik/kendo-theme-default/issues/434


joneff:
Please close related item as well

packages/bootstrap/scss/listview/_theme.scss
```diff
@@ -1,9 +1,16 @@
-@include exports('listview/theme') {
-    .k-listview > .k-state-focused {
-        box-shadow: inset 0 0 0 3px try-darken( $node-bg, 7 );
+@include exports( "listview/theme" ) {
```
gyoshev: This file is identical to the default theme file, so I'd suggest importing the default theme, in order to reduce code duplication.

packages/material/scss/listview/_theme.scss
```diff
@@ -1 +1,19 @@
...
+
+        > .k-state-focused {
+            background-color: rgba(0, 0, 0, .07);
+        }
+
+        > .k-state-selected {
+            color: inherit;
+            background-color: rgba(0, 0, 0, .04);
+        }
+        > .k-state-selected.k-state-focused {
```
gyoshev: It seems that if the states are declared like

```scss
.k-state-selected {}
.k-state-focused {}
```

There won't be the need for a combined selector.

joneff: My bad. They were supposed to be different.

packages/bootstrap/scss/listview/_theme.scss
joneff: Done

PR #116, opened 2018-04-09
refactor(button): Improve styles for button group and add tests

joneff:
Notable changes

* Use better default contrast colors for Bootstrap theme
* Remove redundant selectors for Bootstrap button theme (inherited from button styles)
* Extract styles for legacy ie (those should be no-pointer-events styles)
* Split button styles from a single exports mixin to multiple
* Add tests for button group

As a side note, while grid layout for tests gives a really good overview, few things are not as perfect, like the button-group actual size...


packages/default/scss/button/_theme.scss
```diff
@@ -170,44 +164,98 @@
...
         // Focused state
         &:focus,
         &.k-state-focused {
             box-shadow: $button-focused-shadow;
         }
+
+        // Active state
+        &:active,
+        &.k-state-active {
+            @include fill( contrast-wcag( $button-text ), $button-text, $button-text, none );
```
gyoshev: This previously used the `$button-pressed-*` variables. Is the change intentional?

joneff: The outline button depends solely on text color and nothing more, hence the change.

joneff:
@gyoshev I've updated the preview page and addressed your comment.

PR #125, opened 2018-04-11
Typography, Cards and flexbox layout utils

joneff:
Part of https://github.com/telerik/kendo/issues/8225


packages/material/scss/_variables.scss
```diff
@@ -424,8 +452,84 @@ $icon-size: 24px;
...
+    display-4: font-map($display4-font-size, 40px, $font-weight-normal, $font-family),
+    h1: font-map($h1-font-size, 32px, $font-weight-normal, $font-family),
+    h2: font-map($h2-font-size, 32px, $font-weight-bold, $font-family),
+    h3: font-map($h3-font-size, 28px, $font-weight-normal, $font-family),
+    h4: font-map($h4-font-size, 24px, $font-weight-normal, $font-family),
+    h5: font-map($h5-font-size, $line-height, $font-weight-normal, $font-family),
+    h6: font-map($h6-font-size, $line-height, $font-weight-normal, $font-family),
+    body: font-map($font-size, $line-height, $font-weight-normal, $font-family)
+), $typography-config);
+$typography-config: map-merge((
```
gyoshev: As for the typography configuration, would it work if the styles are applied to the selectors directly in a `_theme.scss` file and using `@extends` to inherit the typography styles in the card? For example:

```scss
// instead of the custom DSL...
$typography-config: map-merge((
    // ...
    h1: font-map($h1-font-size, 32px, $font-weight-normal, $font-family),
    // ...
    body: font-map($font-size, $line-height, $font-weight-normal, $font-family)
), $typography-config);
$typography-config: map-merge((
    card-title: map-get($typography-config, h1),
    card-subtitle: map-get($typography-config, body)
), $typography-config);

// ... we can apply styles directly to the selectors
// _theme.scss
.k-h1 {
    font-size: $h1-font-size;
    line-height: 32px;
    font-weight: $font-weight-normal;
    font-family: $font-family;
}

.k-card-title {
    @extend .k-h1;
}
```

packages/default/scss/card/_layout.scss
```diff
@@ -0,0 +1,240 @@
...
+}
+
+
+
+@include exports("card/layout/ie11") {
+
+    .k-ie11 {
+
+        .k-card-body {
+            flex: 1 1 auto;
```
gyoshev: Perhaps a comment about why this is necessary would be of use here. Perhaps it is a flex bug and a URL can help understanding this, too?

packages/material/scss/_variables.scss
```diff
@@ -424,8 +452,84 @@ $icon-size: 24px;
...
 $pager-padding-x: 8px !default;
 $pager-padding-y: 10px !default;
 
+
+// Card
+$card-padding-x: 16px !default;
+$card-padding-y: 16px !default;
+$card-border-width: 0 !default;
+$card-border-radius: 2px !default;
+$card-box-shadow: mat-elevation-value(2) !default;
```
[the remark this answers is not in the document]
joneff: The idea is to use `mat-elevation-value` in all scenarios where we need to. Unfortunately, the shadow is a bit more complicated than just a single shadow -- it's actually 3 of those. I was wondering if we could could have, say 5 hard-coded levels of shadows, like we do in Asp.Net Ajax and use those.

Would that make sense?

packages/default/scss/common/utils/_text.scss
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    // Named font-weight
+    .k-font-weight-light            { font-weight: $font-weight-light !important; } // sass-lint:disable-line no-important
+    .k-font-weight-normal           { font-weight: $font-weight-normal !important; } // sass-lint:disable-line no-important
+    .k-font-weight-bold             { font-weight: $font-weight-bold !important; } // sass-lint:disable-line no-important
+
+    // Numbered font-weight
+    @each $weight in $font-weight {
+        .k-font-weight-#{$weight}   { font-weight: $weight !important; } // sass-lint:disable-line no-important
```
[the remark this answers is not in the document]
joneff: Agreed. I will remove them

packages/default/scss/card/_layout.scss
joneff: I will add a comment. It's an issue with IE 11 and shorthand flex props. We've had it numerous times already...

Issue #126, opened 2018-04-12
Use bootstrap variables for font sizes


svetq:
Add the following lines to the default bootstrap theme:

$twbs-font-size-base: $font-size-base;
$twbs-input-padding-x: $input-btn-padding-x;
$twbs-input-padding-y: $input-btn-padding-y;
$twbs-input-line-height: $line-height-base;

PR #128, opened 2018-04-13
fix(forms): inherit bootstrap vars for forms integration

gyoshev:
Currently the Bootstrap theme uses the hard-coded values from the Default theme. This change gets the values from the bootstrap framework variables.

closes https://github.com/telerik/kendo-themes/issues/126


gyoshev:
The integration code requires these values to exactly match the ones in Bootstrap, no matter what the values of the theme variables. This allows the customers to change the values of, say, `$font-size-lg`, but use the Bootstrap integration without it breaking.

PR #135, opened 2018-04-19
feat(calendar): add styles for range selection

ggkrustev:
Calendar range selection styling following the [default theme design](https://github.com/telerik/kendo-ux/blob/master/themes/default/calendar.psd).

Apart from adding the range selection styles, these commits also introduce variables that allow the Bootstrap and Material themes to inherit the Default theme.


ggkrustev:
- Some of the styles should be refactored.
- No styling for Bootstrap or Material

PR #137, opened 2018-04-23
chore(button): imrpove icon position for rtl mode

mvgmvg:
related #26


packages/default/scss/button/_layout.scss
```diff
@@ -61,7 +61,7 @@
             }
 
             .k-rtl &,
-            &[dir='rtl'] {
+            [dir='rtl'] & {
```
gyoshev: In the Angular button, the `dir` attribute is set on the `<button>` element ([example](https://www.telerik.com/kendo-angular-ui/components/buttons/rtl/)). Unless this targets a different scenario in Kendo UI for jQuery, I believe the selector should target `.k-button[dir=rtl]` specifically.

PR #145, opened 2018-04-25
Implement Kendo SASS themes for the DropDownTree widget

mvgmvg:
Implement Default, Bootstrap, and Material to DropDownTree widget

related https://github.com/telerik/kendo/issues/8252

DropDownTree widget uses k-multiselect-wrap class name in its rendering for the multiple selection mode.
To inherit styles from the Multiselect widget, I have updated its selectors as well.


packages/bootstrap/scss/multiselect/_theme.scss
```diff
@@ -1,28 +1,25 @@
...
         transition: $transition;
 
+        &:hover,
         &.k-state-hover {
             @include appearance( hovered-input );
         }
 
+        &:focus,
         &.k-state-focused {
             .k-multiselect-wrap {
```
gyoshev: This rule searches for an element that is `.k-multiselect-wrap.k-state-focused > .k-multiselect-wrap`, which doesn't exist, and breaks the focused state:

image

packages/material/scss/popup/_layout.scss
```diff
@@ -3,6 +3,8 @@ $popup-header-padding: $padding-y-lg $padding-x-lg !default;
 $popup-header-font-size: $font-size-sm !default;
 $popup-header-text: rgba($base-text, .54) !default;
 $popup-filter-padding-x: $padding-x;
+$dropdowntree-popup-spacer-y: 0 !default;
+$dropdowntree-popup-spacer-x: $popup-padding-y !default;
```
gyoshev: These variables are for the dropdowntree, and they should not be in the popup files.

packages/bootstrap/scss/popup/_layout.scss
```diff
@@ -1 +1,3 @@
-@import "~@progress/kendo-theme-default/scss/popup/layout";
\ No newline at end of file
+$dropdowntree-popup-spacer-y: 0 !default;
```
gyoshev: This variable is for the dropdowntree, and it should not be in the popup file.

packages/default/scss/dropdowntree.scss
```diff
@@ -0,0 +1,12 @@
+@import "variables";
+@import "mixins";
+
+
+// Dependencies
+@import "common/all";
+@import 'common/forms';
+
```
gyoshev: It seems that the dropdowntree depends on the popup, so this should be listed as a dependency.

packages/default/scss/popup/_layout.scss
```diff
@@ -3,6 +3,8 @@ $popup-header-padding: $list-item-padding-y $list-item-padding-x !default;
 $popup-header-font-size: null !default;
 $popup-header-text: null !default;
 $popup-filter-padding-x: false !default;
+$dropdowntree-popup-spacer-x: $padding-y-lg !default;
+$dropdowntree-popup-spacer-y: $padding-y-lg !default;
```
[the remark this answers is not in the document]
mvgmvg: I have moved it to the correct file. Actually, regarding the design dropdowntree popup spacer should use $padding-y-lg in both directions.

tests/visual/dropdowns.html
```diff
@@ -2,17 +2,19 @@
 <html lang="en">
 <head>
     <title>Dropdowns</title>
-    <link rel="stylesheet" href="../../packages/default/dist/all.css" />
+    <link rel="stylesheet" href="../../packages/bootstrap/dist/all.css" />
```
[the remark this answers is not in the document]
mvgmvg: Cannot reproduce, probably it had been fixed. The provided solution is appropriate maybe for one-time load testing.

PR #146, opened 2018-04-26
feat(chat): add styles for chat widget

joneff:


packages/material/scss/chat/_theme.scss
```diff
@@ -0,0 +1,22 @@
+@import "~@progress/kendo-theme-default/scss/chat/theme";
+
+@include exports("chat/theme/theme") {
```
gyoshev: `chat/theme/material`, perhaps

packages/material/scss/chat/_theme.scss
```diff
@@ -0,0 +1,22 @@
...
+
+    // Chat
+    .k-chat {
+
+
+        // Bubbles
+        .k-bubble {
+            line-height: 1.25;
```
gyoshev: Seems like the `line-height: 1.25` is found both here and in bootstrap. Introducing a variable for these in the default theme can shave off these rules.

packages/material/scss/_variables.scss
```diff
@@ -521,6 +521,49 @@ $card-actions-padding-y: ($card-padding-y - $button-padding-y) !default;
 $card-actions-border-width: 0 !default;
 
 
+// Chat
+$chat-padding-x: 16px !default;
+$chat-padding-y: 16px !default;
+$chat-spacing-x: 8px !default;
+$chat-spacing-y: 16px !default;
```
gyoshev: Not sure what the difference between the `-padding` and `-spacing` is. From the code, it appears that the `-spacing` is used between the various chat elements? If so, using the name `-item-spacing` might be a better fit.

packages/default/scss/chat/_layout.scss
```diff
@@ -0,0 +1,383 @@
...
+                &:nth-of-type(#{$i}) {
+                    animation: 1s blink infinite ($i * .3333s);
+                }
+            }
+        }
+        span + span {
+            margin-left: 5px;
+        }
+
+        @keyframes blink {
```
gyoshev: I believe we should prefix animation identifiers in order to prevent clashes with animations in customer applications.

packages/default/scss/chat/_layout.scss
```diff
@@ -0,0 +1,383 @@
...
+            background-color: currentColor;
+            opacity: 0.4;
+            @for $i from 1 through 3 {
+                &:nth-of-type(#{$i}) {
+                    animation: 1s blink infinite ($i * .3333s);
+                }
+            }
+        }
+        span + span {
+            margin-left: 5px;
```
gyoshev: `$chat-typing-indicator-dot-spacing`, maybe?

joneff:
@gyoshev I've addressed the comments and added tests. Note the test suite is incomplete, but I will add more tests at a later point

PR #149, opened 2018-04-30
feat: add slider styles for material theme

elena-gancheva:


elena-gancheva:
Great 👍 We could log an issue?

PR #151, opened 2018-04-30
feat: add treeview for material theme

elena-gancheva:


tests/visual/treeview.html
```diff
@@ -0,0 +1,173 @@
...
+<head>
+    <title>Treeview</title>
+
+    <link rel="stylesheet" href="../../packages/default/dist/all.css" />
+    <link rel="stylesheet" href="assets/styles.css" />
+    <script src="assets/scripts.js"></script>
+
+    <style>
+        #test-area {
+            max-width: 200px;
```
gyoshev: This crops the screenshot area, so it should probably be deleted.

image

tests/visual/treeview.html
```diff
@@ -0,0 +1,173 @@
...
+    </style>
+</head>
+<body>
+    <div id="test-area">
+        <span></span>
+        <span>Example</span>
+
+        <span>Normal</span>
+        <section>
+            <kendo-treeview childrenfield="items" kendotreeviewexpandable="" kendotreeviewhierarchybinding="" kendotreeviewselectable="" textfield="text" class="k-widget k-treeview" role="tree" aria-multiselectable="false">
```
gyoshev: Try to remove Angular-specific markup from the test case, as it isn't necessary for styling and will be confusing for future implementation of components for other frameworks.

Unused markup includes any directives, ng-* classes (apart from validation), unused `data-` attributes.

packages/material/scss/treeview/_theme.scss
```diff
@@ -0,0 +1,31 @@
...
+        }
+        .k-in {
+            &.k-state-selected {
+                color: $accent;
+                background-color: inherit;
+            }
+        }
+
+        .k-icon {
+            font-size: $treeview-icon-size;
```
gyoshev: This rule should not be necessary at this time, given the [comment on icons here](https://github.com/telerik/kendo/issues/8250#issuecomment-377214034). If this is intentionally added only for the expand/collapse icons, the selector should be

```scss
.k-top,
.k-mid,
.k-bot {
   > .k-icon { ... }
}
```

(otherwise all icons in the TreeView will become large)

packages/default/scss/treeview/_layout.scss
```diff
@@ -1,7 +1,8 @@
-@include exports("treeview/layout") {
-
-    $treeview-indent: 16px;
+$treeview-indent: 16px !default;
+$treeview-item-space-y: $button-padding-y !default;
+$treeview-item-space-x: $button-padding-x !default;
```
gyoshev: In that particular case, `padding-*` seems more accurate than `space-*`, because this is the padding of an item. `spacer` / `gap` / `offset` is better when there are multiple ways to achieve a given styling, e.g. when describing the space between items or the offset of an icon, where `padding` would be too limiting.

PR #152, opened 2018-05-01
 feat: add panelbar for material theme

elena-gancheva:


elena-gancheva:
@joneff @gyoshev I have concerns about handling borders in panelbar and `.k-group` items in it when expanded. The magic rule here -- https://github.com/telerik/kendo-themes/blob/c4a6ce67f91a621c1bd8ce0d05b09a897b3de3f5/packages/material/scss/panelbar/_theme.scss#L7-L13.

PR #163, opened 2018-05-04
Column menu for Angular Grid

gyoshev:


gyoshev:
It seems there is a problem with 0e5e9be, I'll check it before merging.

Issue #159, opened 2018-05-03
DropDownTree issues SASS


Dimitar-Goshev:
#### Wrong items deletion issue - seems like one bug in different reproduction steps

- [x] ~~Wrong items deletion with checkboxes and checkAll enabled (only SASS themes):~~
  1. ~~Use Multiple selection demo.~~
  2. ~~Expand the dropdown and click on Check all~~
  3. ~~Click on x button of Furniture token~~
  - ~~**Expected** - Furniture is unchecked and its token is removed~~
  - ~~**Actual** - Furniture remains checked, its token is not removed, the next item is unchecked.~~ ~~Consecutive clicks on x button remove next tokens and items, but Furniture remains.~~

- [x] ~~Wrong items deletion with checkboxes (only SASS themes):~~
  1. ~~Use Keyboard navigation demo.~~
  2. ~~Expand the dropdown and check Furniture.~~
  3. ~~Expand the dropdown and check Tables.~~
  4. ~~Expand the dropdown and check Sofas.~~
  5. ~~Click on x button of Sofas token~~
  - ~~**Expected** - Sofas is unchecked and its token is removed~~
  - ~~**Actual** - Sofas remains checked, its token is not removed. Instead, Tables is removed.~~

- [x] ~~Wrong items deletion with checkboxes (only SASS themes):~~
  1. ~~Use MVVM demo.~~
  2. ~~Expand the dropdown and check Furniture.~~
  3. ~~Expand the dropdown and check Tables.~~
  4. ~~Expand the dropdown and check Sofas.~~
  5. ~~Click on x button of Sofas token~~
  - ~~**Expected** - Sofas is unchecked and its token is removed~~
  - ~~**Actual** - Both Sofas and Tables are removed.~~

#### Other issues
- [x] ~~Clear button overlaps with toggle icon in single selection mode (LTR and RTL):~~
![image](https://user-images.githubusercontent.com/6614183/39584429-15748bac-4efb-11e8-88fd-674d2b4fc64f.png)
- [x] ~~Check all checkbox should be above root toggle arrows by design.~~
- [x] Vertically not centered template images in tokens:
![image](https://user-images.githubusercontent.com/6614183/39585370-51bc363a-4efd-11e8-8e75-14683be23e10.png)
- [x] Wrong distance between template image and text (probably the image should be to the left to match space above and below it):
[...]

PR #165, opened 2018-05-08
chore(dropdowntree): fixing issues #159

mvgmvg:
related https://github.com/telerik/kendo-themes/issues/159
related https://github.com/telerik/kendo/issues/8252

- fix clear button position
- improve check-all appearance
- fix right filter padding in RTL
- add RTL support for styles checkboxes and radiobuttons


packages/default/scss/dropdowntree/_layout.scss
```diff
@@ -5,19 +5,17 @@ $dropdowntree-popup-spacer-y: $padding-y-lg !default;
 
     // Dropdowntree popup
     .k-popup {
+
         .k-check-all {
             margin: $dropdowntree-popup-spacer-y $dropdowntree-popup-spacer-x 0;
         }
-        .k-check-all + .k-treeview {
+
+        .k-treeview {
```
gyoshev: This way every treeview inside a popup will get the dropdowntree popup spacing, which is incorrect. The previous selector addressed that to target the dropdowntree specifically.

mvgmvg: There are scenarios without check all, so the previous selector did not meet our expectations. Now we added a more specific class name to the k-popup, so I used it for a better selector.

mvgmvg: The dropdowntree is a new widget, so it should not be with the "fix" prefix.

gyoshev: Indeed, the DropDownTree is not a new widget. However, this PR fixes several other issues. Namely, these two issues should be marked as fixes, as they affect all components, not only the DropDownTree:

- fix right filter padding in RTL
- add RTL support for styles checkboxes and radiobuttons

It seems that the new selector is not available in the tests.

PR #202, opened 2018-05-30
fix(dialog): dialog title cuts off letters

elena-gancheva:
see telerik/kendo-theme-default#871


tests/visual/dialog.html
```diff
@@ -30,7 +30,7 @@
              aria-labelledby="titlebar-id">
 
             <div class="k-window-titlebar k-dialog-titlebar k-header" id="titlebar-id">
-                <span class="k-window-title k-dialog-title">Title</span>
+                <span class="k-window-title k-dialog-title">Title dialog type</span>
```
gyoshev: The text "dialog type" is misguiding, as it has nothing to do with the actual test case. Please add comments what it actually tests, like so:

```html
<span class="k-window-title k-dialog-title">
    Title
    <!-- descendants --> g y p
    <!-- ascendant --> t d l
</span>
````

This will allow future maintainers to understand why the text is that way, and prevent accidental deletion of the markup.

File: packages/default/scss/window/_layout.scss [hunk 2]
```diff
@@ -38,9 +39,9 @@
 
     // Title bar title text
     .k-window-title {
-        margin: 0;
+        margin: (-$window-title-lineheight / 8) 0;
```
gyoshev: This evaluates to `0.1625em` in the default theme. Why? Is this perhaps connected to the `$window-titlebar-padding-y`?

elena-gancheva: We have updated the `$window-title-lineheight` from 1 to 1.3. We have to add negative margins at top and bottom close to the updated higher line-height style (0.3)

elena-gancheva:
[Do not merge] Check visual tests for bootstrap. It seems that we have changed line-height for bootstrap content in the dialog, not only in its title.

PR #214, opened 2018-06-15
fix(pivotgrid): pivotgrid stretches beyond container in ie 11

joneff:
Closes https://github.com/telerik/kendo-theme-default/issues/647

At the end it's an ugly hack, but works correctly in IE 11 without breaking anything else.

### DIFF
--- a/packages/default/scss/pivotgrid/_layout.scss
+++ b/packages/default/scss/pivotgrid/_layout.scss
@@ -129,7 +129,7 @@
 
 
 
-@include exports("pivotgrid/configorator/layout") {
+@include exports("pivotgrid/configurator/layout") {
 
 
     .k-fieldselector {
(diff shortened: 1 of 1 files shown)


packages/default/scss/pivotgrid/_layout.scss
```diff
@@ -205,3 +205,21 @@
...
+
+    // Pivotgrid is stretched beyond container in IE 11
+    // see https://github.com/telerik/kendo-theme-default/issues/647
+    .k-ie11 {
+        .k-pivot-layout {
+            width: 100%;
+            table-layout: fixed;
+        }
+        .k-pivot-layout > tbody > tr > td:first-child {
+            width: 280px;
```
gyoshev: Oddly specific 🙂

joneff: I know, but it was the only I thought of

PR #216, opened 2018-06-19
chore(tooltip): use correct variables for tooltip

PreslavKozovski:


elena-gancheva:
@joneff @PreslavKozovski could you please provide naming that indicates that there is theme styles for notification component too? May be extract correct variables for tooltip at another pr? :)

packages/default/scss/notification/_layout.scss
```diff
@@ -1,16 +1,15 @@
...
-    $alert-padding-x: 1em !default;
-
     // Base
     .k-notification {
-        padding: 0;
         @include border-radius();
+        padding: 0;
     }
     .k-notification-wrap {
+        padding: $notification-padding-y calc(#{$notification-padding-x} + #{$icon-size} + #{$icon-spacing}) $notification-padding-y $notification-padding-x;
```
joneff: the calc thing can be extracted as a local variable, say `$notification-padding-end` and be used in both directions

PR #229, opened 2018-07-16
feat(chat): add styles for chat toolbar

joneff:


packages/default/scss/chat/_layout.scss
```diff
@@ -414,6 +414,61 @@ $message-group-spacing: calc( #{$chat-avatar-size} + #{$chat-avatar-spacing} );
...
+        }
+        .k-scroll-button-left {
+            left: 0;
+        }
+        .k-scroll-button-right {
+            right: 0;
+        }
+
+        .k-button {
+            border-width: 0;
```
elena-gancheva: Should these be here --`_layout.scss`? Border, background and color? My be move them into `_themes.scss`? :)

joneff: currently, per design, we want all buttons to be like that: to have no border, background and inherit color

packages/default/scss/chat/_layout.scss
```diff
@@ -414,6 +414,61 @@ $message-group-spacing: calc( #{$chat-avatar-size} + #{$chat-avatar-spacing} );
...
+        padding: $chat-toolbar-padding-y $chat-toolbar-padding-x;
+        width: 100%;
+        box-sizing: border-box;
+        overflow: hidden;
+        position: relative;
+        flex: 0 0 auto;
+
+        .k-button-list {
+            // margin-bottom: -20px;
+            // padding-bottom: 20px;
```
gyoshev: Remove the commented out code, or add a comment why these lines are commented out.

joneff: will do

Issue #231, opened 2018-07-17
Tooltip in Material Theme shows redundant box-shadow


elena-gancheva:
## I'm submitting a...
* Bug report

## Current behavior
Tooltip in Material Theme renders redun

## Expected behavior
There should not be rendered box-shadow from popup-container element

## Minimal reproduce of the issue
https://plnkr.co/edit/Qqr0SeZ1i15Ke205wfvv?p=preview

## Environment
* Kendo Ui for Angular

Package versions:
Tooltip - v1.0.0-dev.201807120905
Kendo Themes - v1.1.3

elena-gancheva:
The issue has been fixed and is available in the [development builds](https://www.telerik.com/kendo-angular-ui/components/installation/development-builds/) of the themе.

PR #233, opened 2018-07-17
fix(tooltip): tooltip in Material theme showing extra padding

elena-gancheva:
see telerik/kendo-themes#231


elena-gancheva:
@gyoshev, @joneff IMHO, we could think of removing the popup default padding in themes, so that other components like Tooltip, Dropdown lists and etc, don't need to override it -- may be we could consider of removing the popup's padding and add it only to components which need it (which are less than these that dont need the padding). We should discuss it @danielkaradachki

gyoshev:
Merged with https://github.com/telerik/kendo-themes/pull/232.
Refactoring will take place in another PR.

Issue #235, opened 2018-07-23
Implement SASS Material theme for Kendo UI widgets


Dimitar-Goshev:
### The following issue shows progress on implementing SASS Material theme for Kendo UI widgets
- [x] Charts
- [x] Linear Gauge
- [x] Radial Gauge
- [x] ListView
- [x] [Grid]
   - [x] Pager
- [ ] TreeList
- [ ] QR Code
- [ ] Barcode
- [ ] Map
- [x] MaskedTextBox
- [x] NumericTextBox
- [x] [Calendar](https://github.com/telerik/kendo-themes/issues/71)
- [x] DatePicker
- [x] TimePicker
- [x] DateTimePicker
- [ ] Pivot
- [x] ScrollView
- [ ] MediaPlayer
- [ ] Diagram
- [ ] Sortable
- [x] [TabStrip](https://github.com/telerik/kendo-themes/issues/24)
- [x] [Splitter](https://github.com/telerik/kendo-themes/issues/47)
- [x] Panelbar
- [x] [Autocomplete](https://github.com/telerik/kendo-themes/issues/43)
- [x] [Dropdownlist](https://github.com/telerik/kendo-themes/issues/40)
- [ ] DropDownTree 
- [x] [ComboBox](https://github.com/telerik/kendo-themes/issues/44)
- [x] [MultiSelect](https://github.com/telerik/kendo-themes/issues/45)
- [x] Upload
- [x] [Button](https://github.com/telerik/kendo-themes/issues/26)
- [x] ButtonGroup
- [x] SplitButton
- [x] DropDownButton
- [x] [Menu](https://github.com/telerik/kendo-themes/issues/78)
- [x] [Toolbar](https://github.com/telerik/kendo-themes/issues/87)
- [x] [Popup](https://github.com/telerik/kendo-themes/issues/51)
- [x] [ContextMenu](https://github.com/telerik/kendo-themes/issues/83)
- [x] [Dialog](https://github.com/telerik/kendo-themes/issues/31)
- [ ] Loader
- [ ] ProgressBar
- [x] [Tooltip](https://github.com/telerik/kendo-themes/issues/49)
- [x] [Window](https://github.com/telerik/kendo-themes/issues/85)
- [x] [Notification](https://github.com/telerik/kendo-themes/issues/181)
- [x] TreeView
- [ ] Spreadsheet
- [ ] Editor
- [ ] Gantt
- [ ] Scheduler
- [x] Switch
- [x] Color Picker
[...]

PR #236, opened 2018-07-23
feat(treelist): add treelist for Material theme

PreslavKozovski:
Part of https://github.com/telerik/kendo-themes/issues/235 and https://github.com/telerik/kendo/issues/8250


elena-gancheva:
@gyoshev, @joneff While implementing Treelist for Material Theme, we found that in Kendo jQuery we have different rendering in the filtering menu than from Kendo Angular. In Kendo jQuery the filtering menu item is not displaying **by design** for Sass themes.

**Current rendering in jQuery:**
image

**Design preview:** 
image

**Current rendering in Kendo Angular:**
image

In jQuery rendering there are missing two classes in the filtering menu buttons: `k-action-buttons` and `k-button-group`. If we add them, the design is as desired for all Sass Themes, **BUT** we regress on **Less themes** and they broke the layout styles.

PR #241, opened 2018-07-26
feat(map): add Map for Material theme

PreslavKozovski:
Part of https://github.com/telerik/kendo-themes/issues/235 and https://github.com/telerik/kendo/issues/8250


packages/material/scss/map.scss
```diff
@@ -0,0 +1,14 @@
+@import "variables";
+@import "mixins";
+
+
+// Dependencies
+@import "common/all";
+@import "map/images/markers";
+@import "button";
+//@import "scroller";
```
PreslavKozovski: @joneff could you please explain why do we need the "scroller" in the Bootstrap theme? And do we need it here?

PR #244, opened 2018-07-31
feat(progressbar): add Progressbar for Material theme

PreslavKozovski:


joneff:
@gyoshev In addition to adding material theme, I've also extracted progressbar variables and refactored the animation.

PR #246, opened 2018-08-01
fix(animations): react animation position and display properties

kspeyanski:
To be published together with [this PR](https://github.com/telerik/kendo-react-private/pull/309)

### DIFF
--- a/packages/default/scss/common/_animations.scss
+++ b/packages/default/scss/common/_animations.scss
@@ -8,6 +8,11 @@
         &-fixed {
             position: fixed;
         }
+
+        &-relative {
+            position: relative;
+            display: inline-block;
+        }
     }
 
 }


packages/default/scss/common/_animations.scss
```diff
@@ -1,9 +1,10 @@
 @include exports('animation/container') {
 
     .k-animation-container {
-        position: absolute;
+        position: relative;
```
joneff: Why do you need the position changed?

kspeyanski: We use to wrap all child elements of a component e.g. Month views in Calendar in a single `k-animation-container`. With stacked animations as is the case with `Reveal`, `Push` and `Zoom` we animate multiple child elements in a single container -> [example](https://www.telerik.com/kendo-react-ui/components/animation/api/PushProps/). 
Not having `relative` or `absolute` position does not allow us to animate 2 or more elements in the same _box_.
From our experience we're overriding the position property in most of our use cases to not brake the child rendering when inside a complex component.
Please take into account that we're using an additional package for animation -> [`react-transition-group`](https://github.com/reactjs/react-transition-group) which also applies some css rules we're not in control of.
As far as i'm concerned the animation styles are only being used by the react suite, so it should not brake anything outside it. Please correct me if i'm wrong

Issue #223, opened 2018-07-03
CSP improvement for font loading (avoid data:)


andrewb273:
## I'm submitting a...
* Suggestion for improvement

## Current behavior
Fonts are loaded via `data:` which requires an additional CSP exception. Specifically I think this is `WebComponentsIcons`.

## Expected behavior
Alongside this behavior which I understand is required for Angular, it'd be great to provide a build option or even an alternate output option (e.g. `all-csp.css` & adding the font manually) so that the `data:` font CSP rule doesn't need to be added.

## Minimal reproduction of the problem with instructions
1. Include `all.css` and perhaps a grid
2. Load page with CSP only allowing fonts from `self`
3. See the CSP exception in the dev console and the lack of fonts loading

## What is the motivation or use case for changing the behavior?
Tighter security via CSP

## Environment
Product:
* Kendo Ui for React (but it's likely for all)

Package versions:
```
"@progress/kendo-theme-default": "^2.54.0"
```

Browser:
- Chrome (desktop) version Version 67.0.3396.99 (Official Build) (64-bit)
- Firefox version 60.0.2 (64-bit)

System:
- TypeScript version: 2.9.1
- Node version: v8.9.0
- Platform: Windows 7

gyoshev:
Thanks for providing a workaround for the problem!

I think we can easily add a variable that, if set, will allow loading the font from a path:

```scss
$icon-font-path: "~@progress/kendo-theme-default/scss/common/font/WebComponentsIcons.ttf";
```

If not set, the default font will be embedded.

andrewb273:
@gyoshev so this would be an actual CSS variable? Are you thinking that it'd take precedence over the base64 font spec, but if null it'd fallback to the base64? The key thing is to not have the base64 in the content loaded to the browser, and the CSS variable solution sounds as though it might still include that base64 in the output. Though I could be misinterpreting? Key thing is that base64 isn't in the content that the browser loads.

gyoshev:
@andrewb273 I meant a SCSS variable, which allows  controls the CSS output. This will allow you to compile the theme without a data URL &mdash; if the variable is set to a path, the SASS processor will not embed the base64 string, but use the path instead.

[...]

PR #250, opened 2018-08-08
feat(icons): allow icon font to be loaded from URL

gyoshev:
This functionality allows the theme to be compiled without a data: URI, which makes it work in CSP scenarios.

See https://github.com/telerik/kendo-themes/issues/223


packages/default/scss/common/_icons.scss
```diff
@@ -3,10 +3,11 @@
 @include exports( "common/icons" ) {
 
     @font-face {
+        $icon-font-url: map-get( $data-uris, 'WebComponentsIcons.ttf' ) !default;
```
joneff: Just a minor comment, This could be exported two lines up, yes?

PR #256, opened 2018-08-17
docs: note that the Roboto font need to be loaded

gyoshev:
To use the Material theme as shown on the site, the Roboto font needs to be loaded. This update shows how to do that.

@joneff it is unexpected that the `$font-family` variable does not set the font to the `.k-widget` class, do you think that it should? It is either that or we should fix the misleading docs:

image


joneff:
We should definitely include it in `.k-widget` or provide styles for `body` that include it. Otherwise it's indeed misleading.

PR #260, opened 2018-08-19
Improve test pages

joneff:
In a nutshell:

* remove data- attributes
* remove aria- attributes
* remove href, title attributes
* remove empty attributes

The resulting code is much cleaner and contains only the attributes we need.


joneff:
@gyoshev Should we change `#test-area` width in checkboxes page? It looks kind of cramped in Bootstrap skin.

PR #263, opened 2018-08-20
test: add angular notification visual tests

elena-gancheva:


packages/material/scss/popup/_theme.scss
```diff
@@ -47,8 +47,10 @@ $grouping-border: $grouping-header-bg !default;
         }
     }
 
-    // Tooltip popup container padding
-    .k-animation-container-shown.k-tooltip-wrapper {
+    // Tooltip and Notification popup container padding
```
elena-gancheva: This will be removed after we refactor(to remove) the Popup default padding

packages/default/scss/notification/_layout.scss
```diff
@@ -28,13 +28,19 @@
...
             line-height: inherit;
 
             &::before {
                 font-size: $icon-size;
             }
         }
+
+        > .k-notification-content {
+            display: inline-block;
+            vertical-align: middle;
```
gyoshev: It appears that this `vertical-align` value is adequate for single line content, but doesn't look that good with multiline content:

image

There's no design for that in kendo-ux at the moment.

packages/material/scss/popup/_theme.scss
```diff
@@ -47,8 +47,10 @@ $grouping-border: $grouping-header-bg !default;
...
+    .k-animation-container-shown.k-tooltip-wrapper,
+    .k-animation-container-shown {
+        .k-notification &,
```
gyoshev: Seems like this targets the `.k-notification .k-animation-container-shown` class, which seems to be missing in the markup. Perhaps the rule is not needed, after all?

tests/visual/notification.html
```diff
@@ -92,13 +107,288 @@
...
+
+        <!-- angular notifications -->
+        <section class="dark">
+            <div class="k-animation-container k-animation-container-fixed k-animation-container-shown">
+                <div class="k-popup k-popup-transparent">
+                    <div class="k-widget k-notification k-notification-info">
+                        <div class="k-notification-wrap">
+                            <span class="k-icon k-i-info"></span>
+                            <div class="k-notification-content">
+                                Hi! I am SUCCESS notification
```
gyoshev: I'd suggest writing the framework flavor in the notification-content. This allows the screenshot to be much easier to understand, e.g. "Success notification in Angular" and "Closable info notification in jQuery"

joneff: Agreed on this one

packages/default/scss/notification/_layout.scss
elena-gancheva: Checked with @MartinTsekov, the notification type icon will be always aligned vertically with the first row (vertical-align: top) of notification's content and with the 'close' button if it exists (this is not handled well in jquery if there is multiline notification). We also consider to make `.k-notification-wrap` `padding-top: 8px`, which does not breaks any rendering in jQuery, so that the notification has more space whithin. What do you think, @gyoshev, @joneff ?

PR #285, opened 2018-09-03
test(editor): add editor test

joneff:


joneff:
@gyoshev it seems that 1024 px is a bit narrow to fit all buttons in bootstrap theme. I've decreased the size of the dropdown, I will try so further and probably reduce the padding of the test area, so we can get those last pixels in

gyoshev:
How about having several editors with different sets of tools? No need to expand the editing area that much either, given that it won't be styled.

joneff:
Done

Issue #252, opened 2018-08-16
Grid improvements.


MilenaCh:
- [x] [#359] The position of drag and drop indicator in the group header is not centered. - https://www.screencast.com/t/3KKlUdks
- [x] [#359] Drag clue has wrong dimensions - https://www.screencast.com/t/3KKlUdks (see the right appearance in the design)
- [x] [#453] Drag clue icon and text are now aligned (see the right appearance in the design)
![indicator](https://user-images.githubusercontent.com/5861835/47294592-3add3400-d616-11e8-94a5-7136f121f901.png)
- [x] [#453] Drag clue cursor icon is not correct (see the right appearance in the design) - should be "move" icon instead of "pointer" icon
- [x] [#454] Sort indexes are too close to the sort indicator. - https://www.screencast.com/t/yzBP6MvEMJ
- [x] ~~There is no hover effect on selected cells and rows - valid for all SASS themes.~~ Follow: #433
- [x] [#287] Vertical borders are missing in multicolumn headers:
https://www.screencast.com/t/GWcRKgcwT6Cc
- [x] [#286] Focus style should be applied when focus some cell via keyboard navigation. 
http://kendobuild-windows.openstack.progress.com/kendo-ui/grid/keyboard-navigation - the user is not able to see which is the currently focused cell.
- [x] [#297] Filter icon change its orientation on click:  https://www.screencast.com/t/vcQKnJbU
- [x] ~~Grid editing command cell should have one normal primary button and one "flat-looking" primary:
https://github.com/telerik/kendo-ux/blob/master/themes/material/grid.psd
implemented:~~ Follow this item: https://github.com/telerik/kendo/issues/8873
<img width="902" alt="screen shot 2018-08-29 at 11 24 25 am" src="https://user-images.githubusercontent.com/16577622/44775439-48c68980-ab7e-11e8-8c6a-f63c19975c8c.png">
by design: 
<img width="1127" alt="screen shot 2018-08-29 at 11 24 36 am" src="https://user-images.githubusercontent.com/16577622/44775452-51b75b00-ab7e-11e8-8d81-e07d26635296.png">

revov:
I thought this issue is a good placeholder to report an interesting "bug":

Nested grids (grids placed in the details section of a row in a parent grid) don't have horizontal dividers (bottom borders) in the last master row.

Check how in this example nested grids have horizontal separators for all rows except the last one:
https://stackblitz.com/edit/angular-izci5w?file=index.html

[...]

PR #287, opened 2018-09-03
fix(grid): Multi-column headers left border

PreslavKozovski:
Fixes an issue from #252


packages/material/scss/grid/_layout.scss
```diff
@@ -51,6 +52,10 @@ $grid-group-indicator-gap: 8px !default;
             font-weight: 700;
         }
 
+        thead > tr:not(:only-child) > th {
```
elena-gancheva: Do we have another option for selecting the desired element?

PreslavKozovski: I tested different approaches, for now, this is the only possible selector that provides the desired outcome.

Anyway, I am going to change the logic of the borders in the material grid today in order to maintain the alignment of the header with the table. Thus, I might change the above selector, however, I do believe that I will use it.

PR #292, opened 2018-09-05
fix focused state in the rows states test

PreslavKozovski:


joneff:
I've change the layout, so the glitch won't happen again.

Issue #300, opened 2018-09-10
MultiColumnComboBox grid-header has incorrect appearance with RTL


mparvanov:
- [x] The grid-header text should be aligned to the right like with less themes 
- [x] The header scroll element should be positioned to the left.

## Current behavior
![image](https://user-images.githubusercontent.com/7202037/45230467-a327be80-b2d1-11e8-8665-7dd351905f8e.png)

## Expected behavior
![image](https://user-images.githubusercontent.com/7202037/45230402-64920400-b2d1-11e8-8be6-5f8956e8dd81.png)

PR #303, opened 2018-09-10
chore(dropdowngrid): improve appearance of sass themes in rtl

joneff:
fixes https://github.com/telerik/kendo-themes/issues/300


packages/default/scss/dropdowngrid/_layout.scss
```diff
@@ -129,11 +133,20 @@
...
+        }
+        &.k-dropdowngrid-popup .k-grid-header .k-header,
+        .k-dropdowngrid-popup .k-grid-header .k-header,
+        .k-grid-list > .k-item > .k-cell {
+            border-left-width: 0;
             border-right-width: $grid-cell-vertical-border-width;
             text-align: right;
         }
+        &.k-dropdowngrid-popup .k-grid-header .k-header:first-child,
+        .k-dropdowngrid-popup .k-grid-header .k-header:first-child,
```
gyoshev: Is this likely to happen, having both `.k-rtl .k-dropdowngrid-popup` _and_ `.k-dropdowngrid-popup.k-rtl` ? Other than that, :shipit:

joneff: I am not sure if we always append `k-rtl` to popups. it might be coming from the body. Hence the double selector. After R3, I will take some time to investigate.

Issue #207, opened 2018-06-08
Filter icons overlap header's text.


MilenaCh:
Navigate to: kendo-ui/grid/filter-multi-checkboxes

Result: https://www.screencast.com/t/bmsxbaKVe

PR #313, opened 2018-09-17
Grid: Filter icons overlap header's text

PreslavKozovski:
related to: #207


joneff:
@PreslavKozovski  the related PR is already merged. test to see if we need to still apply your additional fixes.

PreslavKozovski:
@joneff The issue persists in the jQuery Grid, and these changing are fixing it.

About the angular Grid, with the above code, the padding changes will target the `th.k-filterable` instead of the `th.k-filterable > .k-link`, however, based on my tests, this will not result in any visual changes.

Issue #310, opened 2018-09-12
No RTL for ProgressBar with Default, Bootstrap and Material themes


Dimitar-Goshev:
## Current behavior
ProgressBar progress is LTR

## Expected behavior
ProgressBar progress is LTR

## Minimal reproduction of the problem with instructions
1. Go to https://demos.telerik.com/kendo-ui/progressbar/right-to-left-support
2. Switch theme to Default 2
3. Click on the button below the ProgressBar

## Environment
Product:
* Kendo Ui for jQuery

Browser:
- Chrome (desktop) version 69

PR #317, opened 2018-09-18
Progressbar rtl

PreslavKozovski:
fixes: #310


gyoshev:
The Material Guidelines have a nice document for [implementing RTL support](https://material.io/design/usability/bidirectionality.html). Progress bars fall under the "passage of time" section, and shouldn't be reversed.

PreslavKozovski:
Are we going to follow these guidelines for the other SASS themes? 

If yes, should we remove the "RTL" styles from the LESS themes? By doing this, we will be consistent, however, it will cause a breaking change.

joneff:
If I may, can't we just flip the proggress bar for rtl? And then in media player flip it back?

Actually, if we decide yo use flexbox entirely, we may getaway with having to provide the logic for the various mods entirely

joneff:
If you add `dir="rtl"` to this example what it does is flipping, actually rotating 180 degrees:

```css
.mat-progress-bar[dir=rtl],
[dir=rtl] .mat-progress-bar {
    transform: rotateY(180deg);
}
```

Source: https://github.com/angular/material2/blob/8ed0129906874b9c452a6fb5495b12a5cd18110b/src/lib/progress-bar/progress-bar.scss#L77.

So I guess it's quite a legit solution to simply flip when in RTL

PR #329, opened 2018-10-01
Misaligned Pager link items

PreslavKozovski:
fixes: telerik/kendo-theme-bootstrap#327


PreslavKozovski:
It seems that the issue is no longer replicable. Should we close this PR or we are going to merge the changed test?

joneff:
That's not entirely true, as I can see a couple of issues:

1) if you change the font size of the grid, none of the components in the pager changes the respective font size;
2) even if you explicitly change the font-size of the pager, then the combo will not change its font size, which leads to misalignment
3) Finally, the combo and pagination in bootstrap should be equally tall and they do differ a bit

PR #331, opened 2018-10-01
Fix the direction of the icons in the RTL Pager

PreslavKozovski:
fixes: telerik/kendo-theme-default#774


tests/visual/pager-rtl.html
```diff
@@ -0,0 +1,53 @@
...
+                    </span>
+                </span>
+                items per page
+            </span>
+
+            <a href="#" class="k-pager-refresh k-link" title="Refresh"><span class="k-icon k-i-reload"></span></a>
+
+            <span class="k-pager-info k-label">1 - 20 of 91 items</span>
+        </div>
+
```
elena-gancheva: trim some whitespacing of the html, please :)

tests/visual/pager-rtl.html
```diff
@@ -0,0 +1,53 @@
+<!doctype html>
```
elena-gancheva: I would suggest having both `ltr` and `rtl` pager visual tests in 'pager.html', displayed after a title for the current mode. This way, it will be more convenient to track/see changes in visual tests when needed and reduce png count

PreslavKozovski: I have done this for other controls (see #317) and it appeared that I don't like mixing `ltr` & `rtl` in a single html file :). 

I will add this test to the "pager.html" after #329

PR #336, opened 2018-10-02
Extract tabstrip variables

joneff:
Changes:

* extract tabstrip variables
* make material tabstrip as tall as design

One interesting change is that by [design](https://github.com/telerik/kendo-ux/blob/master/themes/default/layout.psd), in default theme tabstrip has border, but until now we have not included it.


packages/default/scss/tabstrip/_layout.scss
```diff
@@ -1,17 +1,28 @@
...
+        border-style: solid;
+        display: flex;
+        flex-direction: column;
+    }
+
     .k-tabstrip {
         border-width: 0;
+        border-color: transparent;
+        color: inherit;
         background-color: transparent;
```
elena-gancheva: Shouldn't `border-color`, `color` and `background-color` props be in `_theme.scss`?

packages/default/scss/tabstrip/_layout.scss
```diff
@@ -1,17 +1,28 @@
...
-        > .k-tabstrip-items {
-            border: 0 solid transparent;
+        .k-tabstrip-wrapper > & {
+            flex: 1 1 auto;
+        }
+
+        @at-root .k-tabstrip-items {
+            border-width: 0;
+            border-style: solid;
             border-color: inherit;
```
elena-gancheva: Same question here? :)

packages/default/scss/tabstrip/_layout.scss
```diff
@@ -218,12 +233,73 @@ $tabstrip-tab-gap: $nav-item-margin !default;
...
+            }
+        }
+
+    }
+
+}
+
+
+
+
```
elena-gancheva: whitespaces :)

joneff: Yes, whitespaces. It makes code easier to read. I usually leave 4 spaces between major sections of css, namely between `@include exports`  usages.

packages/default/scss/tabstrip/_layout.scss
joneff: `k-tabstrip` is also `k-widget`, which adds all three: border size and color, background color and text color. We do not want a border around the tabstrip (or if we need to, we can use `k-tabstrip-wrapper`). That's why I am removing it in the layout file.

Note: we can always get away with removing tabstrip wrapper, we just need to fix a thing here and there.

packages/default/scss/tabstrip/_layout.scss
joneff: Here is a bit different. I am initially resetting the border to 0. Then in the tabstrip directional variants I am only setting the border-width that's relevant. That's applied to all themes with a variables, so that's why it's in the layout file.

packages/default/scss/tabstrip/_layout.scss
```diff
@@ -120,15 +135,15 @@ $tabstrip-tab-gap: $nav-item-margin !default;
     // Tabstrip variants
     .k-tabstrip-top {
         > .k-tabstrip-items {
-            border-bottom-width: 1px;
+            border-bottom-width: $tabstrip-border-width;
```
joneff: @elena-gancheva that the said relevant border width.

PR #338, opened 2018-10-02
chore(build): add build scripts for api gen and watch

joneff:


package.json
```diff
@@ -19,6 +19,8 @@
   "scripts": {
     "build": "lerna run build",
     "lint": "lerna run lint",
+    "api": "lerna run api",
+    "watch": "lerna run --paralel watch",
```
gyoshev: Shouldn't this be a double-l parallel? 😅

joneff: fixed

PR #344, opened 2018-10-04
chore: remove kendo-common-tasks as top level depenency

joneff:
It seems that kendo-common-tasks is not needed on the top most level (it's still needed and depended upon from the themes).

Removing it from our pipeline should slight decrease our build time


ipeshev:
No need to merge this , since I'm working on something similar that will include this as well , look at kendo-themes-build-pack branch

PR #346, opened 2018-10-04
wip: kendo themes build pack

ipeshev:


ipeshev:
The top level /build folder will be moved in the new themes-build as well

packages/default/scss/postcss.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
gyoshev: All themes need to share the same PostCSS config. Perhaps the config can be shared by exporting it from the themes-build package?

ipeshev: Will try to do it like that .

ipeshev:
@joneff how exactly do you see swatches implemented , currently those are like files in scss/swatches folder, for each webpack entry is created and then passed as target for build. Do you have something different in mind ?

PR #348, opened 2018-10-04
Multiselect close button area

PreslavKozovski:
closes: telerik/kendo-theme-default#620


packages/material/scss/multiselect/_layout.scss
```diff
@@ -7,6 +7,10 @@
         .k-button {
             text-transform: none;
             margin: ($input-padding-y / 2) ($input-padding-y / 2) 0 0;
+
+            .k-select {
+                align-self: unset;
```
venelinpetrov: Is this really needed, the theme imports "default", where you already defined the same rule?

PR #350, opened 2018-10-04
chore(build): use addons for firefox instead of apt-get

joneff:
Quite often the travis cannot acquire firefox for the builds (examples [here](https://travis-ci.org/telerik/kendo-themes/builds/437313450), [here](https://travis-ci.org/telerik/kendo-themes/builds/437313413) and [here](https://travis-ci.org/telerik/kendo-themes/builds/437312033)) which leads to build failure.

That's why I switched to using addon instead of apt-get, which gets the build from mozilla site, instead of the apt-get way.

Naturally, there were some changes... actually two, but enough to impact image tests severly:

Anti aliasing is different and as a result, even in the most basic scenario, like barcode is difference:
image

More complex scenarios with lots of components on the page have even more differences.

But the biggest issue is the scollbar which apparently differs between the normal and ubuntu version of firefox and that makes the screenshots slightly smaller.

@gyoshev thoughts?


gyoshev:
Does the problem occur over a longer period (rather a single day)? If the problem isn't an intermittent issue with the specific package repository, the change seems good.

PR #361, opened 2018-10-11
calendar: remove views height differences

PreslavKozovski:


elena-gancheva:
Could you please provide some info on why is needed this change? :)

PreslavKozovski:
@elena-gancheva Yep, :)
https://www.screencast.com/t/UDAIoDtXm

packages/default/scss/calendar/_layout.scss
```diff
@@ -149,7 +149,7 @@ $calendar-view-gap: 32px !default;
             .k-content {
                 margin: 0;
                 width: 100%;
-                height: 14.167em;
+                height: 15em;
```
joneff: We are substituting one magic number for another. This should be calculated

packages/material/scss/calendar/_layout.scss
```diff
@@ -52,7 +52,7 @@ $calendar-view-gap: 48px !default;
                 6 * $calendar-month-cell-size,
                 $calendar-weekdays-height,
                 -3 * $calendar-meta-cell-size,
-                .5
+                .45
```
joneff: This should be reverted. The previous logic is precise and correct.

PreslavKozovski: I do believe it is neither precise nor correct. 

As you can see in the below video:
https://www.screencast.com/t/Cy8w7P30CwWI

The height of the "k-month" table is 276px. On the other hand, the height of the "k-meta-view" is 240px. This makes me believe that we should add 36px in padding (2*18px), however, we add 2*20px.

PreslavKozovski:
@joneff ,

I changed the height with min-height because I believe this property should be used in this scenario. 

About the calculations of this magic number, if we decide to do calculations, this height/min-height is not needed at all. However, we will have to do calculations similar to those of the Material theme for every view of the Calendar for every theme. 

Please, let me know what you think about this, and how we should proceed.

joneff:
This PR is superseded by #390 and telerik/kendo#9066

Issue #351, opened 2018-10-08
Combobox toggle icon misplaced & Numeric rulers misplaced


ipeshev:
## I'm submitting a...

* Bug report 

## Current behavior
Combobox toggle icon misplaced & Numeric rulers misplaced when used as cell editors.
![25180210-182a-4c2f-8ed4-f6a75b2844f3_style issues](https://user-images.githubusercontent.com/4143177/46596751-40029500-cae6-11e8-8d45-116429f7db5c.jpg)

## Expected behavior
The clear icon should not be over toggle arrow.

## Minimal reproduction of the problem with instructions
1.Open this [stackblitz sample](https://stackblitz.com/edit/kendoui-material?embed=1&file=src/app/components/grid/grid.component.html)
2.Try to edit an row.
## Environment
Product:

* Kendo Ui for jQuery
* Kendo Ui for Angular
* Kendo Ui for React
* Kendo Ui for Vue

Package versions:

```

```

Browser:

- Chrome (desktop) version XX
- Chrome (Android) version XX
- Chrome (iOS) version XX
- Firefox version XX
- Safari (desktop) version XX
- Safari (iOS) version XX
- IE version XX
- Edge version XX

System:
- TypeScript version: XX 
- Node version: XX  
- Platform:

PR #364, opened 2018-10-12
fix(dropdowns): misalignments in editable row of grid in Material Theme 

elena-gancheva:
see telerik/kendo-themes#351
see telerik/kendo-angular#1715


tests/visual/grid.html
```diff
@@ -82,6 +82,192 @@
 
         </div>
 
+        <div class="section">
+            <kendo-grid class="k-widget k-grid k-grid-no-scrollbar" style="width: 800px;">
```
gyoshev: I think that a better place for these tests would be the grid-editing test, as the grid visual test is getting pretty crowded.

PR #366, opened 2018-10-15
add kendo-theme-tasks

joneff:
Fixes telerik/kendo#8952
see telerik/kendo-themes#378


joneff:
@gyoshev @ipeshev by using gulp instead of webpack, aggressive caching of node_modules and with a bit of trickery (custom sass loader, compile by convention) I was able to cut down build time significantly.

Yes, I am still missing twbs-compat, but that will come as well.

I also need to double check with themebuilder integration.

ipeshev:
In dist folder all.js files are missing do we need them ?

ipeshev:
The gulp code is looking good but the black magic with lib imports may backfire at some point , when some plugin is updated.
Is this commited with --no-verify option ?

ipeshev:
Aggressive caching in travis may be is not the best options, from our previous experience. I prefer not to think about the freshness of my node_modules all the time.

joneff:
@ipeshev I have not integrated lint, yet.

ipeshev:
@joneff  it is fast (bow) . Shouldn't dist/all.css files be minified ? Still missing twbs-compat.css in packages/default/  folder. Do we still need webpack entry points ( packages/**/build/all.js  ) ?

ipeshev:
It will be good to have root level tasks for serving demos from which snapshots are derived.

gyoshev:
> Shouldn't dist/all.css files be minified ?

The output should be piped at least through CSS nano (or similar)

> Still missing twbs-compat.css in packages/default/ folder.
> Do we still need webpack entry points ( packages/**/build/all.js )

I think that both of these are not necessary for the output package, nor are they documented.

joneff:
> The output should be piped at least through CSS nano (or similar)

@gyoshev, could we consider add `.min` suffix i.e. have one unminified version and one minified?

@ipeshev, I am going to revert some of the changes you made, since we need "import-dependency" scripts for derivative themes.

joneff:
@gyoshev @ipeshev @elena-gancheva latest changes are in

current cli:
```js
kendotheme build // builds theme
kendotheme build --watch // builds theme and watches for changes
kendotheme build --swatches // builds swatches, if any
kendotheme lint // runs sass-lint
kendotheme api // runs sass-doc
```

That's just v.0.0.1. If we agree on most things, can we move it forward, so I can test it with themebuilder and material theme?

elena-gancheva:
The current cli changes are good. We should test the themebuilder behavior.
Can these changes affect the lerna managing of the themes package somehow?

joneff:
They shouldn't affect lerna in any way

Issue #375, opened 2018-10-17
Filtermenu arrows are not positioned correctly


PreslavKozovski:
Filtermenu arrows are not positioned correctly

![image](https://user-images.githubusercontent.com/17339722/47087508-dd677280-d224-11e8-9995-0af499a5f204.png)

PR #377, opened 2018-10-17
Filtermenu arrows not positioned correctly in the jQuery Grid

PreslavKozovski:
fixes: #375


packages/material/scss/common/_forms.scss
```diff
@@ -60,7 +60,6 @@ $forms-invalid-color: inherit !default;
     }
 
     // Expanded
-    .k-state-border-up .k-select .k-i-arrow-60-down,
```
joneff: That lines needs to be there. Instead, scope the styles:

```scss
.k-state-border-up > .k-select .k-i-arrow-60-down,
.k-state-border-down > .k-select .k-i-arrow-60-down {
...
}
```

Issue #378, opened 2018-10-18
Building version 2.57.1 of the Default theme leads to a SASS compiler error


dtopalov:
Error: Base-level rules cannot contain the parent-selector-referencing character '&'.

![image](https://user-images.githubusercontent.com/7533599/47149220-925c6680-d2db-11e8-857d-08c7ddd94f26.png)

Seems related to:

[https://github.com/sass/sass/issues/1873](https://github.com/sass/sass/issues/1873)

joneff:
The issue is regarding https://github.com/telerik/kendo-themes/blob/1e693f729c261ca986baa2f44397f0708687d34a/packages/default/scss/notification/_layout.scss#L75 and it's a valid issue. The `&` shouldn't be there indeed

joneff:
The issue is regarding https://github.com/telerik/kendo-themes/blob/1e693f729c261ca986baa2f44397f0708687d34a/packages/default/scss/notification/_layout.scss#L75 and it's a valid issue. The `&` shouldn't be there indeed

elena-gancheva:
The issue has been fixed and is available in the [development builds](https://www.telerik.com/kendo-angular-ui/components/installation/development-builds/) of the themе.

elena-gancheva:
Shipped with `2.58.0` version of the theme.

PR #379, opened 2018-10-18
fix(notification): base-level rules for RTL throwing SASS compiler error

elena-gancheva:
see telerik/kendo-themes#378


packages/default/scss/notification/_layout.scss
```diff
@@ -72,31 +72,27 @@
...
 
     // RTL
-    &[dir='rtl'] .k-notification-wrap,
-    .k-rtl .k-notification-wrap {
-        padding-left: $notification-padding-end;
-        padding-right: $notification-padding-x;
+    .k-notification,
+    .k-notification-container {
+        &[dir='rtl'],
+        .k-rtl & {
```
gyoshev: I think these selectors can be simplified a bit, since the `.k-rtl` class is always applied to an enclosing element (typically the `<html>` / `<body>` tags). Thus instead of generating 4 rules for each nested selector, the enclosing styles could be

```scss
.k-notification-container[dir='rtl'],
.k-rtl .k-notification {
   ...
}
```

This should work because .k-rtl is used only in jQuery widgets that don't have a notification container, otherwise the dir=rtl is set to only the k-notification-container class.

PR #382, opened 2018-10-23
test(calendar): improve calendar test

joneff:
Improve calendar test by adding more views. Needed for kendo-jquery-multiview-calendar styles.


elena-gancheva:
**Note:** the selected row lines are positioned static in test cases, because they are handled by javascript code, so the misalignment in visual previews is expected and it is not a styling bug.
Reference image:
image

Issue #380, opened 2018-10-19
Notification content does not wrap on small screens


gyoshev:
## I'm submitting a...

* Bug report 

## Current behavior

When the browser is resized, the notification content does not wrap, and the user cannot read it.

![image](https://user-images.githubusercontent.com/90405/47205488-baf16880-d38e-11e8-9f86-f51a09385198.png)

## Expected behavior

Wrapping the content seems like a good solution.

![image](https://user-images.githubusercontent.com/90405/47205521-d3fa1980-d38e-11e8-9e43-bba25d02486a.png)

## Minimal reproduction of the problem with instructions

Open the [templates example in Stackblitz](https://www.telerik.com/kendo-angular-ui/components/notification/content/#toc-using-templates) and resize the output area.

## What is the motivation or use case for changing the behavior?

Notification usability on small screens

## Environment
Product:

* Kendo Ui for Angular

gyoshev:
Workarounds until this issue is fixed:

- Use a template and wrap the notification content in a container with `white-space: normal` set.
- Add the following CSS rule to the page: `.k-notification-content { white-space: normal }`

elena-gancheva:
The fix is available in the [development builds](https://www.telerik.com/kendo-angular-ui/components/installation/development-builds/) of the themе.

elena-gancheva:
Shipped with `2.58.0` version of the theme.

PR #383, opened 2018-10-23
fix(notification): notification content does not wrap on small screens

elena-gancheva:
see telerik/kendo-themes#380


elena-gancheva:
This PR hits the update of Firefox to version 63.0.0, which is causing the long list of updated visual previews.

PR #387, opened 2018-10-25
Scrollview rendering improvements

PreslavKozovski:
As discussed, I am changing these class names in order to achieve consistency. 

related to: telerik/kendo-angular-scrollview/pull/122


packages/bootstrap/scss/scrollview/_layout.scss
```diff
@@ -1 +1,3 @@
-@import "~@progress/kendo-theme-default/scss/scrollview/layout";
\ No newline at end of file
+$scrollview-pager-item-border-width: 0 !default;
```
joneff: this, along with the rest of the variables, should be extracted in the `_variables.scss`

packages/bootstrap/scss/_variables.scss
```diff
@@ -964,6 +964,26 @@ $progressbar-indeterminate-gradient: null !default;
...
 
+// Scrollview
+
+$scrollview-pager-offset: 20px !default;
+$scrollview-pager-item-spacing: 20px !default;
+$scrollview-pager-item-border-width: 0 !default;
+
+$scrollview-pagebutton-size: 8px !default;
+$scrollview-pagebutton-bg: rgba($carousel-indicator-active-bg, .5) !default;
+$scrollview-pagebutton-border: null !default;
```
elena-gancheva: Why setting border to `null`, instead of `0 !default`? I think we set defaults to null only for `background-color`, `color`, etc. Is this variable used for border-color?

packages/bootstrap/scss/_variables.scss
```diff
@@ -964,6 +964,26 @@ $progressbar-indeterminate-gradient: null !default;
...
+$scrollview-pagebutton-primary-bg: null !default;
+$scrollview-pagebutton-primary-border: null !default;
```
elena-gancheva: same here

packages/bootstrap/scss/_variables.scss
PreslavKozovski: As far as I know, when `null` is assigned to a property in the `.scss` files, the same line/rule is not transferred to the `css` file. And this is making our production files smaller.

However, I'm not 100% sure about that :D @joneff could you please confirm or deny this?

gyoshev: Here's a confirmation:
image

Note that both values have their use-cases: `null` can reduce output, and `0` can override existing borders.

elena-gancheva: yes, the question here is do we need to override existing border on the pagebutton?

PreslavKozovski: We do not need to override existing borders in the Bootstrap theme. Actually, we do not use these variables at all in this theme. However, I added them in order to have consistency with the other `_variables.scss` files. 

Do you think this approach is okay or we should remove the unneeded variables, and if we need them in the future - add them?

elena-gancheva: Imho, setting to `null` is better approach.

packages/bootstrap/scss/_variables.scss
elena-gancheva: ignore

PR #388, opened 2018-10-25
feat(scheduler): add angular scheduler layout styles

elena-gancheva:
TODO:
- when smaller screen resolutions, we should consider how to display  view switcher buttons.

I suggest having single button (when on small screens) with a dropdown, containing the different views. What do you think? @tsvetomir @gyoshev @joneff @lkarakoleva


joneff:
The responsive behaviour is another task. As discussed, we'll have both class based styles and media queries that "inherit" the said classes.

PR #395, opened 2018-10-30
fix(scheduler): borders glitches

elena-gancheva:


packages/default/scss/scheduler/_theme.scss
```diff
@@ -124,6 +124,13 @@ $scheduler-event-text: #fff !default;
         border-color: $primary-button-border;
     }
 
+    // Edit dialog
+    .k-scheduler-edit-dialog {
+        .k-dialog-buttongroup {
+            border-color: $border-color;
```
gyoshev: I'd expect these to be styled by the dialog action button styles. Might be a good idea to verify why this style is necessary.

PR #420, opened 2018-11-02
Use correct colors in material panelbar and calendar

joneff:
Needed for #376.


PreslavKozovski:
Based on the design the text color should be different, and the widget should have a border.

On the other hand, I see that our design is using an archived material design guidelines (and I was not able to find new guidelines for the "Expansion panels"). Having said that, if the current behavior is the expected one, I believe we should update our UX repo.

Could you please confirm that these borders/colors are desired?

joneff:
@PreslavKozovski the guidelines are in https://github.com/telerik/kendo-ux/blob/master/themes/material/layout.psd

image

The changes are as follows:

1) Panelbar needs to use it's own specified color and not an external one. That's needed for themebuilder
2) Calendar selected hover colors is calculated to be darker in light themes and lighter in dark themes. So while the current color may not be the correct one (although it is), the important bit is that it's now calculated correctly. That's also needed for themebuilder

PR #429, opened 2018-11-09
multiview calendar polishing

joneff:
targets telerik/kendo#9079


matanasov:
@joneff you can merge this

PR #435, opened 2018-11-12
extract DateTime variables

### LINKED ISSUE #408: Provide variable for time-picker header padding
Currently the variable `$time-header-padding` is needed in theme-bootstrap. Move it to theme-default when #720 is done.

PreslavKozovski:
related to: #408

### DIFF
--- a/packages/bootstrap/scss/_variables.scss
+++ b/packages/bootstrap/scss/_variables.scss
@@ -979,6 +979,15 @@ $window-shadow: null !default;
 $window-focused-shadow: null !default;
 
 
+// DateTime
+
+$time-list-highlight-color: $border-color !default;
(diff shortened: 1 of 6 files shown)


packages/bootstrap/scss/_variables.scss
```diff
@@ -966,6 +966,15 @@ $progressbar-indeterminate-gradient: null !default;
 $progressbar-chunk-border: $bg-color !default;
 
 
+// DateTime
+
+$time-list-highlight-color: $border-color !default;
+$time-list-focused-bg: rgba(0, 0, 0, .02) !default;
+$time-list-title-focus: #000 !default;
```
elena-gancheva: [suggestion]: use `black` instead

Issue #409, opened 2017-10-13
Buttons are not displayed correctly on IE on small window size


svetq:
Buttons are not displayed correctly on IE on small window size.

Check the following sample plunker for reproduction:
[http://plnkr.co/edit/Gnng47LUYP4ZMHFVbfDW?p=preview](http://plnkr.co/edit/Gnng47LUYP4ZMHFVbfDW?p=preview)

Workaround - add the following style:
.k-button { display: inline-block; }

PR #436, opened 2018-11-12
button not displayed correctly on IE on small window sizes

PreslavKozovski:
fixes: #409

### DIFF
--- a/packages/default/scss/button/_layout.scss
+++ b/packages/default/scss/button/_layout.scss
@@ -379,4 +379,14 @@
 
     }
 
+    .k-ie9,
+    .k-ie10,
+    .k-ie11 {
+
+        .k-button,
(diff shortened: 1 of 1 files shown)


packages/default/scss/button/_layout.scss
```diff
@@ -80,6 +80,15 @@
             }
         }
 
+        .k-ie9 &,
```
joneff: Flexbox issues should be resolved like this

```suggestion
        .k-no-flexbox {
             .k-button {
                display: inline-block;
            }
        }
```

In addition, you could move the selector at the end of the file and add some handling for buton-group as well.

PreslavKozovski:
In a chat discussion, we decided to keep the "k-ie9/10/11" classes.

PR #439, opened 2018-11-12
update PivotGrid implementation according to the latest design

### LINKED ISSUE #400: Pivotgrid - update implementation according to the latest design
Update pivotgrid according to the latest [design](https://github.com/telerik/kendo-ux/blob/master/themes/default/pivotgrid-jquery.psd)

PreslavKozovski:
related to: #400


elena-gancheva:
Seems like some `rtl` spacings are missing, see the visual previews:
image

PreslavKozovski:
The orange color of the sorted columns is not achievable with the current rendering. Thus, I would suggest postponing this change when a major rendering improvement is needed for the PivotGrid.

PreslavKozovski:
@elena-gancheva The PivotGrid does not support RTL, for the time being. 

So, do we need to add these styles if the RTL mode is not officially supported?

Issue #433, opened 2018-11-09
Add hover effect on the selected cells and rows for the Grid


PreslavKozovski:
Add hover effect on the selected cells and rows for the Grid.

PR #448, opened 2018-11-16
Grid: add hover + selected rows styles

PreslavKozovski:
closes: #433


PreslavKozovski:
What do you think of changing the logic apply the colors to the TD elements? I tested different scenarios, and I was not able to find any regressions.

joneff:
The logic will work OK if the colors are semi transparent. However, for the time being we do not have such themes, so the logic will do.

The correct way to do it is to have hover-selected colors.

PR #454, opened 2018-11-19
Grid: sort indexes are too close to the indicator

PreslavKozovski:
part of: #252


packages/default/scss/_variables.scss
```diff
@@ -295,6 +296,17 @@ $header-selected-border: $selected-border !default;
 $header-selected-gradient: null !default;
 
 
+// Icons
+$icon-size: 16px !default;
```
elena-gancheva: [suggestion] We can calculate these values one from another, this way if the user customize `$icon-size`, it will affect `$icon-size-lg` as well?

PR #458, opened 2018-11-20
Switch improvements

PreslavKozovski:


packages/bootstrap/scss/_variables.scss
```diff
@@ -986,6 +986,33 @@ $scrollview-navigation-default-opacity: .7 !default;
...
+$switch-handle-border-radius: $switch-size !default;
+$switch-handle-border-width: 1px !default;
+
+$swtich-on-handle-left: null !default;
+$swtich-off-handle-left: null !default;
+
+$switch-label-display: inline !default;
+$switch-label-spacing: 8px !default;
+
+$switch-occupied-space: add-three($switch-container-border-width, $switch-container-padding-x, $switch-handle-size);
```
PreslavKozovski: @elena-gancheva @joneff 
The layout is pretty much finalized. Could you check these three variables, and let me know if you see a better way to these calculations?

PreslavKozovski:
Merge this PR together with https://github.com/telerik/kendo-angular-inputs/issues/294 and https://github.com/telerik/kendo-react-private/issues/474

jivanova:
- [ ] Implement switch labels positioning and dimension calculations. Switch labels will be moved outside of the k-switch-handle.

packages/default/scss/switch/_layout.scss
```diff
@@ -1,99 +1,131 @@
...
+        @include border-radius( $switch-handle-border-radius );
         width: $switch-handle-size;
         height: $switch-handle-size;
-        display: inline-block;
-        margin: 0 ($switch-padding * 3) 0 $switch-padding;
-        border-width: $switch-border-width;
+        border-width: $switch-handle-border-width;
         border-style: solid;
+        display: inline-block;
         vertical-align: middle;
```
elena-gancheva: not needed

packages/default/scss/switch/_theme.scss
```diff
@@ -47,73 +15,124 @@ $switch-active-shadow: 0 0 0 2px rgba(0, 0, 0, .06) !default;
...
         &.k-state-focused {
-            outline: none;
-            box-shadow: $switch-focus-shadow;
+            @include box-shadow( $switch-off-focused-shadow );
         }
 
-        &:active {
-            box-shadow: $switch-active-shadow;
+        .k-switch-label-on {
+            color: transparent;
```
elena-gancheva: do we need to assign a variable here for customization?

packages/default/scss/switch/_layout.scss
```diff
@@ -1,99 +1,131 @@
...
     .k-switch {
+        @include border-radius( $switch-border-radius );
         cursor: pointer;
+        width: $switch-size;
+        background: none;
         border: 0;
-        display: inline-block;
-        font-size: $font-size-xs;
+        outline: 0;
+        box-shadow: none;
```
elena-gancheva: Shouldn't this be in `_theme.scss`?

packages/default/scss/switch/_layout.scss
```diff
@@ -1,99 +1,131 @@
...
-        display: block;
-        text-align: center;
+        display: $switch-label-display;
+        width: $switch-label-width;
         position: absolute;
-        text-transform: uppercase;
+        top: 50%;
+        transform: translateY(-50%);
+        text-transform: $switch-label-text-transform;
         text-shadow: none;
```
elena-gancheva: `_theme.scss`

packages/default/scss/switch/_layout.scss
PreslavKozovski: When removed, the handle is no longer aligned correctly.

packages/default/scss/switch/_theme.scss
PreslavKozovski: No, we don't, this is the color of the label in the opposite state. For example, when the state is "OFF" we always want that the "ON" label is not present(transparent);

tsvetomir:
If I understand correctly, these changes will render the theme unusable with older versions of the components. It should be marked as a **breaking change** and released as a major version.

PR #490, opened 2018-12-13
docs: create initial documentation for kendo-theme-tasks package

joneff:
That's a crude and unrefined version of the initial documentation for kendo-theme-tasks.

I do feel that we could make theme-tasks to work with pretty much everything trough configuration, since it's powerful enough. The two main features being simple and efficient `~` importer and embed assets.


tsvetomir:
A section explaining what Swatches are and how they're designed to be used would complete the README.

gyoshev:
Are the local CONTRIBUTING.md files [picked up by Github](https://blog.github.com/2012-09-17-contributing-guidelines/)? If not, probably a root-level file will reduce the repetition in the themes, and the tasks can have a local one that is specific to them.

joneff:
@tsvetomir , we already have a short "article" in the wiki -- https://github.com/telerik/kendo-themes/wiki/Swatches. When it's informative enough we could mention in it the readme.

After discussion with @gyoshev, we decided to keep the contributing files, so they are distributed with each package.

Issue #518, opened 2018-12-22
Chart: Set background-color in sass


SchneiderMK:
## I'm submitting a...

* Suggestion for improvement  

## Current behavior

I like to set the background-color of the charts at a global place.
I can not use the global variables $bg-color or $base-bg. In my case the base colors are different.
I know that I can use <kendo-chart-area background="..."></kendo-chart-area> to set the background-color of the chart. But I must do this then every time.

Is there a global styling option available?

## Expected behavior

Possibility to set the background-color of a chart via css-class or sass-variable.

elena-gancheva:
This functionality has been implemented and is available in the [development builds](https://www.telerik.com/kendo-angular-ui/components/installation/development-builds/) of the themеs.

joneff:
The improvement is available both with

* @progress/kendo-theme-default@2.63.0
* @progress/kendo-theme-bootstrap@2.18.0
* @progress/kendo-theme-material@1.6.0

and 

* @progress/kendo-theme-default@3.0.0+
* @progress/kendo-theme-bootstrap@3.0.0+
* @progress/kendo-theme-material@2.0.0+

PR #521, opened 2018-12-31
Chart: adding component specific background color variable

TeyaVes:
see telerik/kendo-themes#518


elena-gancheva:
The commit message tag should be `feat(charts): ...`, so that we can output a new `dev` version of the themes, which will be including the improvement for customizing charts `background-color: $chart-bg`. You can use `interactive rebase` to change the commit messages and remove the ones which are in between

PR #539, opened 2019-01-09
Fix swatches and add them to npm package

joneff:
Previously, swatches only built the first swatch, due to heavy optimizations on our side and the way gulp works.

I've extended the code to allow building without optimizations, which does take some more time.

### DIFF
--- a/packages/theme-tasks/lib/tasks/sass.js
+++ b/packages/theme-tasks/lib/tasks/sass.js
@@ -11,7 +11,6 @@ const nodeSassPackageImporter = require("../utils/nodesass-packageimporter");
 const postcss = require("gulp-postcss");
 const autoprefixer = require("autoprefixer");
 const calc = require("postcss-calc");
-const browserSync = require("browser-sync").create();
 const sassdoc = require('sassdoc');
 
 const postcssPlugins = [
@@ -27,12 +26,6 @@ const sassOptions = {
     outputStyle: "compressed",
     importer: nodeSassPackageImporter
 };
-const browserSyncOptions = {
-    server: {
-        baseDir: "./"
-    },
-    open: false
-};
 
 
 // #region core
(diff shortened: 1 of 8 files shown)


packages/theme-tasks/lib/tasks/sass.js
```diff
@@ -40,10 +38,17 @@ function build(glob) {
     let _glob = glob || paths.sass.src;
 
     return gulp.src(_glob)
-        .pipe(sass(sassOptions).on("error", sass.logError))
+        .pipe(sass(fastSassOptions).on("error", sass.logError))
+        .pipe(postcss(postcssPlugins))
+        .pipe(gulp.dest(paths.sass.dist));
+}
+function buildSlow(glob) {
```
tsvetomir: How about `buildAll` or similar? Slow or fast is relative :)

packages/theme-tasks/lib/utils/nodesass-packageimporter.js
```diff
@@ -37,4 +31,30 @@ module.exports = function(url, prev/*, done*/) {
         file: file
     };
 
+}
+
+function slowPackageImporter(url) {
```
tsvetomir: May be refactored into a function that returns the needed importer to avoid duplication:

```js
const packageImporter = (options = { cache: false }) => url => {
    if (!url.startsWith("~")) {
        return null;
    }

    let file = path.resolve( path.join(
        process.cwd(),
        "node_modules/",
        url.slice(1)
    ) );

    if (options.cache && imported.has(file)) {
        return EMPTY_IMPORT;
    }

    imported.add(file);

    return { file };
}
```

packages/theme-tasks/lib/tasks/sass.js
joneff: It's actually really noticeable. fast takes about 3 to 5 seconds. slow takes about 30.

I will never the less think on new names

joneff: applied changes

packages/theme-tasks/lib/utils/nodesass-packageimporter.js
joneff: changes aaplied

Issue #540, opened 2019-01-09
Wrong selected style color for Grid's context menu.


MilenaCh:
1. Navigate to kendo-ui/grid/multicolumnheaders demo page
2. Open context menu and select some of the options in it.

Result:
https://www.screencast.com/t/LDpqSo6TL5g

PR #550, opened 2019-01-14
Grid: column settings dropdown styles

TeyaVes:
Related to: https://github.com/telerik/kendo-themes/issues/540

* As discussed with @lkarakoleva , the styles of the items in the column menu dropdown should be the same as the items in the DropDownList (https://demos.telerik.com/kendo-ui/dropdownlist/index)


packages/bootstrap/scss/grid/_theme.scss
```diff
@@ -133,4 +133,31 @@
     .k-column-list-wrapper {
         border-color: inherit;
     }
+
+    .k-column-menu {
```
joneff: `.k-column-menu` is the popup. The styles below should be applied here.

Considering the original styles of the menu, perhaps we should rework it for the default theme:

```scss
.k-menu {} // styles for menu
.k-context-menu {} // styles for context menu
.k-column-menu {} // styles for column menu
```

OR, we should switch to using context menu there, because that's what we want. @PreslavKozovski could you assist for that?

joneff:
For the time being, we are merging this. At a later stage, we'll need to change the menu to be context menu in order not to override styles

Issue #404, opened 2017-12-05
No visual indication applied to selected Scheduler event


Dimitar-Goshev:
It is not possible to tell which event is selected, as there is no indication to make it stand out.
Go to [Scheduler / Keyboard support](https://demos.telerik.com/kendo-ui/scheduler/selection) with Default v2/ Bootstrap v4
Click on an event - current appearance:
![image](https://user-images.githubusercontent.com/6614183/33605088-e9eecca4-d9c0-11e7-968e-daad13c5ac77.png)

Expected appearance - highlight similar to LESS themes:
![image](https://user-images.githubusercontent.com/6614183/33605114-05eca778-d9c1-11e7-9694-4ae782187e21.png)

TeyaVes:
Updated designs for focused items in the SASS themes: https://github.com/telerik/kendo-ux-private/issues/9

PR #559, opened 2019-01-17
Scheduler: selected events style

TeyaVes:
Related to: https://github.com/telerik/kendo-themes/issues/560 & https://github.com/telerik/kendo-themes/issues/404

Default theme - normal state- light overlay on all background colors
                        - selected state- no overlay

Bootstrap theme - selected state- inner box-shadow

Material theme - selected state- outer box-shadow + light overlay on dark backgrounds & dark overlay on light backgrounds


packages/default/scss/_variables.scss
```diff
@@ -1148,6 +1148,18 @@ $switch-off-handle-hovered-border: $switch-on-handle-hovered-border !default;
 $switch-off-handle-hovered-gradient: null !default;
 
 
+// Scheduler
+$modal-header-border-color: #cccccc;
+$modal-inner-padding: 1em;
```
PreslavKozovski: I see that these two variables are extracted from the `theme` / `layout` files, however, I do not see where are we using them. Could you elaborate on do we need these variables or we can remove them?

packages/material/scss/scheduler/_theme.scss
```diff
@@ -45,18 +42,37 @@ $scheduler-event-text: $primary-contrast !default;
...
             // TODO: $scheduler-content-area-bg
             background-color: $primary-contrast;
         }
-    }
-    .k-event.k-state-selected {
-        @include appearance( selected-node );
+
+        &.k-state-selected {
+            @include appearance( selected-node );
+            background-color: $scheduler-event-bg;
```
PreslavKozovski: Removing both lines(`@include app...` and `backgorund-col...`) will result in the same.

packages/default/scss/scheduler/_theme.scss
```diff
@@ -117,13 +110,28 @@ $scheduler-event-text: #fff !default;
...
+        &::before {
+            display: block;
+            left: 3px;
+            opacity: 0.35;
+            background-color: $scheduler-event-text;
+        }
+
+        &.k-state-selected {
+            color: $scheduler-event-selected-text;
+            @include appearance( selected-node );
```
PreslavKozovski: Remove this line. Removing it will not change the outcome.

PreslavKozovski:
As far as I see, you are using the `$scheduler-event-border-radius` variable only in the `Default` theme. Please, correct me if I am wrong. 

If this is the case, for the other themes, set the value to `null`

Issue #566, opened 2019-01-22
Switch layout is broken when used within a k-form element


dtopalov:
Within **.k-form**:

![image](https://user-images.githubusercontent.com/7533599/51542831-d32cc400-1e64-11e9-8917-40f233275fca.png)
![image](https://user-images.githubusercontent.com/7533599/51542853-dfb11c80-1e64-11e9-8bb9-b66b593869e6.png)

Within a regular FORM element (with no k-form class):

![image](https://user-images.githubusercontent.com/7533599/51542883-f5bedd00-1e64-11e9-9238-f22c24a4df0f.png)
![image](https://user-images.githubusercontent.com/7533599/51542893-fbb4be00-1e64-11e9-828b-5eb321ee6d73.png)

Runnable demo:

[https://stackblitz.com/edit/angular-wbz1be-kxvxb3?file=app/app.component.ts](https://stackblitz.com/edit/angular-wbz1be-kxvxb3?file=app/app.component.ts)

kspeyanski:
We've also received a bug report for the React Switch. A quick fix is to apply the 'k-display-inline-flex' class to the root element of the switch.

PR #575, opened 2019-01-24
Switch: layout in a k-form element

PreslavKozovski:
targets: https://github.com/telerik/kendo-themes/issues/566


packages/default/scss/input/_layout.scss
```diff
@@ -430,7 +430,7 @@ $checkbox-size: $icon-size !default;
     .k-form {
         input,
         label:not(.k-checkbox-label):not(.k-radio-label),
-        .k-widget:not(.k-calendar) {
+        .k-widget:not(.k-calendar):not(.k-switch) {
```
elena-gancheva: Do we have another option of selecting this element, without using `:not`?

PreslavKozovski: The only possible other solution is to make this selector deeper 
https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/switch/_layout.scss#L10
However, this will conflict with the approach of styling our widgets(using `.k-widget.k-switch` for this one.

Having said that, using the `:not` selector seems like a better solution in this case.

PR #581, opened 2019-01-29
feat: add support for RTL of Kendo Angular Treeview

svetq:


elena-gancheva:
Great, we have a valid visual test! :+1: 
Two things to consider:
1. Rebase the branch to remove not-related to this featire commits;
2. There is a bug in the direction of the RTL arrows (or missing RTL styles)
See the image below:
image

Issue #578, opened 2019-01-25
Grid column menu icon is not positioned correctly in multi column headers in IE and Edge


matanasov:
Open this demo in IE / Edge and notice the column menu icon location

http://kendobuild-windows.openstack.progress.com/kendo-ui/grid/multicolumnheaders

![icon](https://user-images.githubusercontent.com/5861835/51750198-fa300380-20b9-11e9-8b05-071585b28b8e.png)

PR #585, opened 2019-01-30
Grid: column menu icon position in IE and Edge

TeyaVes:
Related to: https://github.com/telerik/kendo-themes/issues/578


TeyaVes:
In IE and Edge, the absolute position of the of the column menu icon in the table cell is not working and the easiest solution I found would be to float the element to the right as it is in the LESS themes only in Edge and IE

Issue #574, opened 2019-01-24
Displaced columns and not centered headings in Scheduler Agenda view with Material


Dimitar-Goshev:
## I'm submitting a...
* Bug report 

## Current behavior
![image](https://user-images.githubusercontent.com/6614183/51678587-6d1b7a80-1fe5-11e9-81c5-67a5f013e02a.png)

## Expected behavior
Columns are aligned and headings are centered
![image](https://user-images.githubusercontent.com/6614183/51678690-ad7af880-1fe5-11e9-8fe5-f6dfc6b71960.png)

## Minimal reproduction of the problem with instructions
1. Go to https://demos.telerik.com/kendo-ui/scheduler/resources-grouping-vertical
1. Switch to Material theme and choose Agenda View

## Environment
* Kendo Ui for jQuery - 2019.1.115

Browser:
- Chrome 71
- IE 11
- Firefox 62

Issue #554, opened 2019-01-16
Misaligned cells in DAY&WEEK view


ipeshev:
## I'm submitting a...

* Bug report 

## Current behavior
![problem](https://user-images.githubusercontent.com/4143177/51248708-b2aec680-1999-11e9-9035-b681f7558980.png)

## Minimal reproduction of the problem with instructions
1.Open this demo https://www.telerik.com/kendo-angular-ui-develop/components/scheduler/
2.Choose material theme
3.On bootstrap theme it is not working at all

## Environment
Product:

* Kendo Ui for jQuery
* Kendo Ui for Angular
* Kendo Ui for React
* Kendo Ui for Vue

Package versions:

```

```

Browser:

- Chrome (desktop) version XX
- Chrome (Android) version XX
- Chrome (iOS) version XX
- Firefox version XX
- Safari (desktop) version XX
- Safari (iOS) version XX
- IE version XX
- Edge version XX

System:
- TypeScript version: XX 
- Node version: XX  
- Platform:

mparvanov:
The cells are aligned in Kendo jQuery Day and Week views.

joneff:
The issue is related only to material + kendo-angular and the issue is fixed in the develop branch

PR #594, opened 2019-02-04
Scheduler: columns layout in material theme

TeyaVes:
related to: https://github.com/telerik/kendo-themes/issues/574 & https://github.com/telerik/kendo-themes/issues/554

In the material theme, the font-size of the cells in the header is smaller than the ones in the body, which is why the width in em units was producing different width size for the columns in the header and the columns in the body


TeyaVes:
> The fix works for the jQuery Scheduler, however, the Angular one still looks misaligned. Could you check this?

The fix should be working for Angular scheduler as well, although the visual tests look broken- this is because there is an inline style that sets a fixed width of the first 'all day' cell, which is relevant for the default theme, but not for the bootstrap and the material one.

Issue #438, opened 2018-11-12
Add outer group header to selects


lkarakoleva:
Since the limitation caused by the virtualization does not apply for angular and react, we could have outer group header:

Default theme: https://github.com/telerik/kendo-ux/blob/master/themes/default/dropdowns.psd
Material theme: https://github.com/telerik/kendo-ux/blob/master/themes/material/dropdowns.psd

Top group header should be bold to separate it from the list items.

tapopov:
This is available in the [development builds](https://www.telerik.com/kendo-angular-ui/components/installation/development-builds/) of the themе.

tapopov:
Available in @progress/kendo-theme-default@3.3.0

PR #601, opened 2019-02-06
feat(dropdowns): add grouping for angular

tapopov:
Related to https://github.com/telerik/kendo-themes/issues/438


packages/default/scss/popup/_theme.scss
```diff
@@ -20,7 +21,7 @@
         background: $grouping-header-bg;
     }
     .k-popup .k-list .k-item > .k-group::before {
-        border-color: $grouping-header-bg $grouping-header-bg transparent transparent;
+        border-color: $grouping-border;
```
elena-gancheva: This doesn't look right. Previous styles are applied that the element will have border 1px everywhere, but with different color (`border-top/right/bottom/left`). Please, check it.

Issue #332, opened 2018-10-02
.k-button.k-flat:hover.k-state-active resets the button color


joneff:
`.k-button.k-flat:hover.k-state-active` resets the button color and even when the button is of type primary, it gets a grayish tint:

![image](https://user-images.githubusercontent.com/8048/46334918-7b0c5080-c62e-11e8-9315-e5a508bec90a.png)

PR #610, opened 2019-02-07
Button: .k-button.k-flat:hover.k-state-active corrected color

TeyaVes:
related to: https://github.com/telerik/kendo-themes/issues/332


packages/bootstrap/scss/button/_theme.scss
```diff
@@ -201,7 +201,12 @@
     }
     .k-button.k-primary.k-flat,
     .k-button.k-primary.k-bare {
-        color: $accent;
+
+        &,
+        &:hover,
```
PreslavKozovski: I believe we should add
```
&.k-state-hover
```
to the selectors here and below.

packages/material/scss/button/_theme.scss
```diff
@@ -47,7 +47,11 @@ $button-disabled-text: rgba(0, 0, 0, .38) !default;
     .k-button.k-flat.k-primary,
     .k-button.k-bare.k-primary,
     .k-button.k-outline.k-primary {
-        color: $primary-button-bg;
+
+        &,
+        &:hover {
```
joneff: still missing `k-state-hover`

Issue #353, opened 2018-10-08
Scrolling buttons of bottom tabstrip are not correctly aligned


joneff:
Related to #336

![image](https://user-images.githubusercontent.com/8048/46608332-1b6de380-cb0d-11e8-9a32-188c331e2c73.png)

mparvanov:
The scroll buttons position is incorrect in IE: 
![image](https://user-images.githubusercontent.com/7202037/53474727-89905400-3a76-11e9-98d6-c85d5c07a7c7.png)

PR #613, opened 2019-02-11
Tabstrip: bottom scrolling buttons

TeyaVes:
related to: https://github.com/telerik/kendo-themes/issues/353


joneff:
@TeyaVes do we need `bottom: 0` both in the layout and theme?

TeyaVes:
> @TeyaVes do we need `bottom: 0` both in the layout and theme?

It's actually not needed and I've removed it.

Issue #617, opened 2019-02-15
Incorrect current time marker arrow


danielkaradachki:
example: https://dojo.telerik.com/@SiliconSoul/IQiSEGon

The border color in the arrow rules:
https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/scheduler/_layout.scss#L495

is overridden by:
https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/scheduler/_layout.scss#L18

danielkaradachki:
Also incorrect arrow position in the timeline view due to missing relative position on k-scheduler-header-warp:

        .k-scheduler-header-wrap {
            position: relative;
        }

PR #619, opened 2019-02-18
Current Time

danielkaradachki:
https://github.com/telerik/kendo-themes/issues/617

Works, but don't like it very much. 
An arrow class that sets the common styles should be used. I can add it for angular but the same should be done for jQuery:

    .k-current-time-arrow {
        width: 0;
        height: 0;
        background: transparent;  
        border: 4px solid transparent;
    }

Also, I am not sure if the border color shouldn't be moved to the _theme.scss file.


tests/visual/scheduler.html
```diff
@@ -639,6 +641,181 @@
             </div>
         </div>
 
+        <div id="scheduler" style="height: 600px;" data-role="scheduler" class="k-widget k-scheduler k-floatwrap" role="grid" aria-multiselectable="true">
```
elena-gancheva: You can clear the `role` and `aria` labels and attributes from the test markup, if you are not testing them

packages/default/scss/scheduler/_layout.scss
```diff
@@ -17,6 +17,43 @@
...
+            background: red;
+            position: absolute;
+        }
+
+        .k-current-time-arrow-down {
+            width: 0;
+            height: 0;
+            background: transparent;
+            border-bottom: 4px solid  transparent;
+            border-top: 4px solid #f00;
```
elena-gancheva: Consider this color to be a variable or check if there is a existing variable for it

packages/default/scss/scheduler/_layout.scss
```diff
@@ -17,6 +17,43 @@
...
+        }
+
+        .k-current-time-arrow-left {
+            width: 0;
+            height: 0;
+            background: transparent;
+            border-bottom: 4px solid  transparent;
+            border-top: 4px solid transparent;
+            border-left: 4px solid transparent;
+            border-right: 4px solid #f00;
```
elena-gancheva: same here

tests/visual/scheduler.html
elena-gancheva: Provide info as a comment that this is an Angular scheduler test, please

PR #633, opened 2019-02-26
Scheduler RTL layout

tsvetomir:
Event arrows are pointing in the opposite direction in RTL.

image


TeyaVes:
Just a few things I noticed:
- the kendo-themes/tests/visual/scheduler-angular-rtl-month.html is currently showing normal direction, not rtl

Also, I think we should improve a couple of things in the RTL scheduler and maybe open a separate issue for these:
- we should remove the padding-right from _kendo-scheduler .k-event, .k-event_ selector and add padding-left instead. I think that this space is saved for the close button (or at least it's supposed to) and since the close button will be on the left side, the padding should be there as well.
- we might need to move the overlay in the default theme to the left so that it creates a darker border on the right side, instead of the left side:

---currently---


---ideally---

PR #639, opened 2019-03-01
fix(scheduler): icon padding in RTL

tsvetomir:
* Added tests in `tests/visual/1024px` to test media queries
* Fixed Scheduler issue in RTL on small screens


tsvetomir:
Well, maybe except for the Material theme :/

TeyaVes:
> Well, maybe except for the Material theme :/

The material theme should be okay now :)

PR #653, opened 2019-03-08
ColorPicker: wrapper hover styles not applied

dimitar-pechev:
* Applied the `.k-state-hover` also to `.k-picker-wrap:hover` - no point in using js to achieve the hover effect.
* While on the ColorPicker topic: the styles of the `.k-item` elements inside the ColorPalette are inconsistent when the palette is used as a standalone component and when it's used in the ColorPicker component (inside a popup), as you can see below. The `cursor: pointer` is applied from the popup styles (`.k-popup .k-item`). Perhaps we should apply that style for the standalone palette as well. Raising the question here, as there's no point in opening a separate PR for this in my opinion.

image

image

### DIFF
--- a/packages/bootstrap/scss/colorpicker/_theme.scss
+++ b/packages/bootstrap/scss/colorpicker/_theme.scss
@@ -14,6 +14,7 @@
         }
 
         // Hovered state
+        .k-picker-wrap:hover,
         .k-state-hover {
             @include appearance( hovered-button );
         }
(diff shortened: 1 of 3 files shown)


elena-gancheva:
@dimitar-pechev I suggest to apply the styles for hover states on the `k-item` for both:
`
.k-colorpicker,
.k-colorpalette {
    .k-item {
        cursor: pointer;
    }
}
` and remove the cascading from the `.k-popup` class

dimitar-pechev:
Applied the style just to `.k-colorpalette .k-item` as it seems to be sufficient. The rather strange styling for `.k-item`s in the popup is located [here](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/popup/_layout.scss#L25).

PR #666, opened 2019-03-21
Validator: layout changes

PreslavKozovski:
targets: https://github.com/telerik/kendo-themes/issues/667


packages/material/scss/validator/_theme.scss
```diff
@@ -0,0 +1,3 @@
+@include exports("validator/theme") {
```
PreslavKozovski: @joneff 
Is exporting an empty module okay?

joneff: sure

Issue #682, opened 2019-04-08
Angular Toolbar is displaying wrong button color in Material


elena-gancheva:
The Toolbar in Kendo UI for Angular is displaying wrong color of control type buttons - `button` and `toggle button` in Material theme.

See the image below:
![image](https://user-images.githubusercontent.com/8842674/55706515-0d304f80-59ea-11e9-8b9c-e19140f6ed4c.png)

Reproducible on both demos:
https://www.telerik.com/kendo-angular-ui-develop/components/toolbar/control-types/

alpkarli:
It's same in the buttons outside the toolbar. Can you also fix it?

PR #684, opened 2019-04-09
Fixes for toolbar and splitter

joneff:
* flex splitter does not collapse under IE 11
* add styles for flexible spacer (needed for PDF viewer)


elena-gancheva:
TODO: see https://github.com/telerik/kendo-themes/issues/682

PR #693, opened 2019-04-15
Added cursor pointer on elements with k-link class

TeyaVes:
elements with k-link class act as links and should be styled like links similar to the ones in the LESS themes

related to: https://github.com/telerik/blazor/issues/88


PreslavKozovski:
For consistency, I believe we should add the same change in the `Less` themes.

PR #694, opened 2019-04-15
feat: add styles for PDF Viewer widget

joneff:
Part of telerik/kendo#9500


joneff:
> Do we have RTL support?

With the latest addition of separator items, the toolbar is truly RTL. The PDF viewer itself is a drawing canvas. Drawing, does not seem susceptible to RTL.

To sum up: the toolbar is, the rest, not so sure, but it doesn't really fall within our expertise.

Issue #678, opened 2019-04-02
Add styles for Spreadsheet k-spreadsheet-has-comment:after cell indicator


mparvanov:
Currently the new Spreadsheet Comments functionality applies a small red square on the top right corner of the cells that has comments in the Less themes. The same approach need to be applied in the SASS themes for consistency.

![image](https://user-images.githubusercontent.com/7202037/55409675-1d1fdd80-556b-11e9-972b-6f9f1cd57994.png)

Less styles: https://github.com/telerik/kendo/pull/9438/commits/0239af718be9566b83ecb1a659b8812c40165864

Related to telerik/kendo#9355

PR #697, opened 2019-04-17
Spreadsheet: add comment indicator sass styles

Juveniel:
targets: #678 

related to: telerik/kendo#9355


packages/bootstrap/scss/spreadsheet/_theme.scss
```diff
@@ -157,3 +167,23 @@
...
 
 }
+
+
+
+
+@include exports("spreadsheet/rtl/theme") {
+
+    .k-rtl .k-spreadsheet,
+    .k-spreadsheet[dir="rtl"] {
```
PreslavKozovski: If the Spreadsheet can be a child of an "RTL" element, add this selector:

```
[dir="rtl"] .k-spreadsheet
```

Is this an okay scenario?

packages/bootstrap/scss/spreadsheet/_theme.scss
```diff
@@ -157,3 +167,23 @@
...
+
+        .k-dirty {
+            border-color: $error $error transparent transparent;
```
PreslavKozovski: The same as 
https://github.com/telerik/kendo/pull/9584#discussion_r277264174

(there are similar sections below)

packages/default/scss/spreadsheet/_layout.scss
```diff
@@ -517,6 +518,25 @@
...
+
+    // Comments
+    .k-spreadsheet-has-comment::after {
+        content: "";
+        display: block;
+        position: absolute;
+        top: 0;
+        right: 0;
+        left: auto;
+        border-width: 3px;
```
joneff: As a minor comment here is what the main what the structure should be:

```scss
selector {
    // mixins, if any:
    // extends, if any:

    // content, if any:
    content: "";

    // margin, padding, width, height in that order
    // if we need to list directional properties, we start from top and move clockwise
    // min max follow the original value in that order
    margin: 10px;
    padding-top: 2px;
    padding-right: 1px;
    padding-left: 0;
    width: 100%;
    min-width: 1px;
    max-width: 33px;
    height: auto;

    // border; individual props are: width, style, color in that order
    // do not use 1px solid for shorthand
    border-width: 1px;
    border-style: dashed;

    // decoration: color, background
    color: red;
    background: gold;

    // font: size, line-height, family, weight, style, align, white-spacing, word wrap
    font-size: 12px;
    font-weight: bold;

    // display and float, flex / grid props afterwards
    display: flex;
    float: left;

    // position, z-index, offsets (directional way preffered)
    position: absolute;
    z-index: 30;

    // The rest of the props are not as common
}
```

Issue #641, opened 2019-03-05
Error: ExtendCompound when building with Dart Sass- Invalid scss files 


Bjego:
When I try to compile the scss files from all.scss it fails with 
"Error: compound selectors may no longer be extended."

OS: Windows 7.
Package.json: 
[package.json.txt](https://github.com/telerik/kendo-themes/files/2930260/package.json.txt)

Here are the commands I use to build the scss file:
```

c:\Git\kendoIssue>npm i
c:\Git\kendoIssue>.\node_modules\.bin\sass .\node_modules\@progress\kendo-theme-default\scss\all.scss .\kendo.css
Error: compound selectors may no longer be extended.
Consider `@extend .k-button, ::before` instead.
See http://bit.ly/ExtendCompound for details.

   ╷
99 │                 @extend .k-button::before;
   │                         ^^^^^^^^^^^^^^^^^
   ╵
  node_modules\@progress\kendo-theme-default\scss\toolbar\_layout.scss 99:25  root stylesheet
```

After fixing this one file. It fails on other places as well

blemaire:
Just had the same issue, it appears it has been deprecated for a while and now removed:
https://sass-lang.com/documentation/at-rules/extend#disallowed-selectors

What the changes of getting this fixed in the next kendo theme release?

cluetjen:
Same here. As soon as you install sass node package, angular cli will fail with this error. Our problem is, that we're using ng-packagr and ng-packagr >= 5.x will install sass and so, it will break angular cli. Our workaround for now: Use ng-packagr 4.x

s.a. 
https://github.com/telerik/kendo-themes/issues/551
https://github.com/ng-packagr/ng-packagr/issues/1253

blemaire:
@cluetjen have you had any issues with downgrading to ng-packager v4?

cluetjen:
@blemaire - no, but I fear it will break for angular 8, see https://github.com/ng-packagr/ng-packagr/blob/master/CHANGELOG.md . What's the reason for your thumbs down on my comment?

blemaire:
> What's the reason for your thumbs down on my comment?

@cluetjen apologies for the thumb down it wasn't meant for your comment but rather to express how I felt about having to downgrade to ng-packagr v4. I removed it to prevent this confusion.

Thanks for your reply about the effect of downgrading

joneff:
As of dart-sass v 1.14.3, the `::before` pseudo element is treated as such and that particular problem does not appear.

[...]

PR #705, opened 2019-04-24
Fixed invalid syntax on compound selector

blemaire:


joneff:
@blemaire consider the following snippet:

```scss
// input
.button { color: red; }
.flat { color: blue; }
.button.flat {  color: green; }

.compound { @extend .button.flat; }

// output
.button { color: red; }
.flat { color: blue; }
.button.flat, .compound { color: green;}
```
Which is precisely what we want to achieve.

Simply changing the selectors will make the code compile but will not yield the same result. A much more thorough refactoring is required and we are considering it for a later stage. Probably when we drop older versions of angular. The said refactoring will be a breaking change so we'll aptly bump the major version as well.

We are aware of the issue with dartsass (#641) and for now we stick to compiling with libsass.

/cc @Bjego, @cluetjen

blemaire:
I understand how this PR does not yield the same result, could we come up with a solution involving [Extend/inheritance](https://sass-lang.com/guide#topic-7) for now.

this place is a good candidate to fix most of the compound selectors: [kendo-theme-default-button]5https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/button/_layout.scss#L206)

A `%k-button__k-flat' selector could be added and then used in the places that currently use `.k-button.k-flat` as a compound selector: so `@extend .k-button.k-flat;` would become `@extend %k-button__k-flat;`

joneff:
That might patch things up for now, yes. Of course we need to add a patch for `.k-button.k-primary.k-flat` as well. There are some 175 imports, mostly `.k-button.k-flat`, but we should check them all.

Also, I am not quite keen on BEM style syntax, as we don't have it anywhere else.

Lastly, patching things up, while solving the compiling issue, does not solve the bigger issue, namely how do we deal with difference in behaviour and should we consider updating our modifier strategy, say `.k-button-primary` instead of `.k-button.k-primary` to sort of future proof for this particular behaviour.

blemaire:
Ahah, the name was just an example, I'm only getting started with kendo so not quite used to naming conventions. I did not realise there was that many compound selectors, sounds like hell to migrate 😢 
Closing this fornow, let's carry on on the related issue #641

PR #708, opened 2019-04-24
Calendar minimal paddings and alignment

TeyaVes:


packages/default/scss/calendar/_layout.scss
```diff
@@ -598,6 +602,11 @@
                 }
             }
 
+            > :first-child,
```
joneff: The buttons have classes, we should use those!

For now I am letting the PR ahead

PR #723, opened 2019-05-09
test(colorpicker): add more tests for angular colorpicker

elena-gancheva:


tests/visual/colorpicker.html
```diff
@@ -232,8 +232,73 @@
...
+
+            <section>
+                <p>Angular colorpicker with opened colorpalette</p>
+                <kendo-colorpicker class="k-widget k-colorpicker">
+                    <span class="k-picker-wrap k-state-default">
+                        <span class="k-selected-color" style="background-color: rgb(78, 174, 197);"></span>
+                        <span class="k-select"><span class="k-icon k-i-arrow-s"></span></span>
+                    </span>
+                </kendo-colorpicker>
+                <kendo-popup class="k-animation-container k-animation-container-shown" style="left: 522px; top: 680px;">
```
PreslavKozovski: I'd suggest removing the `top` and `left` styles and use a static position for the popup. By doing this, the Bootstrap and Material tests would also look okay

Issue #709, opened 2019-04-25
Wrong icon margin in column menu in grid in RTL


TeyaVes:
## I'm submitting a...
Bug Report

Wrong margin in the column menu icon within the grid in RTL

## Current behavior
https://dojo.telerik.com/iGIRotoj/11

currently, the horizontal margin is:   margin: -2px 4px 0 -4px;

## Expected behavior
in RTL the horizontal margin should be vice-versa: -2px -4px 0 4px;

## Environment
Product:
* Kendo Ui for jQuery
* Kendo Ui for Angular
* Kendo Ui for React
* Kendo Ui for Vue

PR #731, opened 2019-05-13
Grid column menu icons in rtl

TeyaVes:
targets: https://github.com/telerik/kendo-themes/issues/709

related to: https://github.com/telerik/kendo/pull/9678


tests/visual/grid-column-menu-left.html
```diff
@@ -0,0 +1,262 @@
+<!doctype html>
```
PreslavKozovski: Change the file name to "...-rtl", not "left"

tests/visual/grid-column-menu-left.html
```diff
@@ -0,0 +1,262 @@
...
+<html lang="en" class="k-typography k-no-animations">
+<head>
+    <title>Grid Column Menu</title>
```
PreslavKozovski: Change the title

packages/default/scss/menu/_layout.scss
```diff
@@ -116,6 +116,14 @@ $image-spacing: 8px !default;
         .k-item {
             display: block;
             border-color: inherit;
+
+            .k-icon {
+                margin-right: $image-spacing;
```
PreslavKozovski: If the `margin-right` style is inside "menu" file. Why the relevant RTL styles are in the "grid" file?

TeyaVes: The margin-left and right in the 'grid' file is resetting the margins in RTL mode for the column menu within the grid, which are set as:

    .k-columnmenu-item {
        padding: $nav-item-padding-y $nav-item-padding-x;
        cursor: pointer;

        > .k-icon {
            margin-right: $icon-spacing;
        }
    }
  
While the styles set in the 'menu' file are targeting the menu outside of the grid.

PR #744, opened 2019-05-21
chore: set specific http server version

KeiBlag:
Related to the following issue on Windows: https://github.com/indexzero/http-server/issues/525


thornjad:
This issue is fixed in http-server version 0.12.0

Issue #750, opened 2019-05-23
MultiSelect's readonly mode displays unexpected clearButton 


martintabakov:
## I'm submitting a...

* Bug report 

## Current behavior
In Kendo Multiselect in Bootstrap 4 Theme: if readonly mode and clearButton are enabled, the clear button is visible at the right of the input field. If clicked, the selected tag will be removed.

## Expected behavior
The clearButton shouldn't be visible.

## Minimal reproduction of the problem with instructions
The Behavior is reproduced in this Dojo example: https://dojo.telerik.com/UsiXAZuv. Simply run and click the clear button.

## Environment
Product:

* Kendo Ui for jQuery
Package versions: 2019.2.514

PR #753, opened 2019-05-23
Multiselect: hide clear button when readonly

TeyaVes:
targets: https://github.com/telerik/kendo-themes/issues/750


packages/default/scss/common/_forms.scss
```diff
@@ -219,14 +219,15 @@ $forms-invalid-color: $error !default;
             }
         }
     }
-    .k-multiselect-wrap:hover,
```
joneff: this selector is needed for angular / react.

The correct way would be to have a k-readonly class

Issue #751, opened 2019-05-23
[ButtonGroup] Incorrect margins when is used in Grid's toolbar


mbechev:
If a button is before (or after) the ButtonGroup in Grid's toolbar, there is no margin between them.
Also, gaps are showing between buttons that are in ButtonGroup (same appear when  ButtonGroup is used in Toolbar Component).

## Expected behavior
Margin (4px) should be added between the ButtonGroup and any other buttons. 

There should be no gaps between buttons in ButtonGroup

## Minimal reproduction of the problem with instructions
Check the following stackblitz example:
[https://stackblitz.com/edit/angular-nud4dl-tchh9x?file=app/app.component.ts](https://stackblitz.com/edit/angular-nud4dl-tchh9x?file=app/app.component.ts)

Screenshot:
![buttons-group](https://user-images.githubusercontent.com/33104481/58252151-16913380-7d6e-11e9-91b6-d47a88c992d1.png)

## Environment
Product:
* Kendo UI for Angular

Browser:
- Chrome (desktop) [ 74.0.3]
- Firefox [66.0.5]
- IE [11]
- Edge [latest]
- Opera [latest]

joneff:
fixed by #766

PR #766, opened 2019-06-04
Fix margin between buttons in grid toolbar

KeiBlag:
Related to https://github.com/telerik/kendo-themes/issues/751


elena-gancheva:
Great! 👍 remove the inner chore commits

packages/default/scss/grid/_layout.scss
```diff
@@ -376,8 +376,15 @@
             vertical-align: middle;
         }
         .k-button + .k-button {
```
joneff: Initially, we were supposed to have only buttons in the toolbar, hence the selector. It should now be `* + *`. Of course, since now we can have toolbar and button groups, that should be immediate descendant (child) selector or `>`. The end form of the selector should be `> * + *`

packages/default/scss/grid/_layout.scss
```diff
@@ -376,8 +376,15 @@
...
+            margin-left: 0;
+        }
+        .k-button + .k-button-group .k-button:first-of-type {
+            margin-left: $grid-toolbar-button-spacing;
+        }
```
joneff: The is not needed.

packages/default/scss/grid/_layout.scss
```diff
@@ -376,8 +376,15 @@
...
+        .k-button-group + .k-button {
```
joneff: The same way we can have buttons after button groups, we can have button groups after button groups; toolbars after buttons; comboboxes after toolbars etc.

packages/default/scss/grid/_layout.scss
KeiBlag: Fixed.

packages/default/scss/grid/_layout.scss
KeiBlag: Fixed.

packages/material/scss/grid/_layout.scss
```diff
@@ -34,6 +34,18 @@ $grid-group-drag-clue-opacity: .9 !default;
             box-shadow: none;
         }
 
+        .k-grid-toolbar {
```
joneff: this is not needed. It's already loaded from line 19. You need to change `$grid-toolbar-button-spacing` to 0.

Issue #701, opened 2019-04-18
Material theme hover style issue with multiselect running in dialog


DanielHTpg:
**Describe the bug**
When using a kendo angular kendo-multiselect control in a kendo-dialog the hover state styles of the multi select items get overriden by  ".k-window-content .k-button" selector that removes the background color with "background: none !important;". Then in a hover state, the item will have no background.

I've tested the hover state in a dialog with the default theme and there it works as expected.

**To Reproduce**

  1. Open StackBlitz https://stackblitz.com/edit/angular-multiselect-theme-error
  2. Hover over the multi select item "Basketball"
  3. Observe incorrect hover color

**Expected behavior**
Hovered items have proper background color.

**Screenshots**
The red circle is an item that I'm currently hovering over.
![image](https://user-images.githubusercontent.com/24475714/56379983-4732fa00-6211-11e9-9250-5fe58b3e73e3.png)

**Browser**
 - OS: Windows 10
 - Browser: Chrome
 - @progress/kendo-theme-material: 2.4.1

TeyaVes:
reproducible in Kendo jQuery as well: http://dojo.telerik.com/OxAMOKoS/2

KeiBlag:
Fixed with #768

PR #768, opened 2019-06-05
Fix hover background color of multiselect control in a dialog window in Material theme

KeiBlag:
Related to https://github.com/telerik/kendo-themes/issues/701


elena-gancheva:
Change the commit message `fix: scope selectors` to something that actually describes the bug

Issue #771, opened 2019-06-10
Double border in editor toolbar


ivelina-penkova:
In Angular and React there is a border of the editor component and another border of the toolbar component which results in double border in all themes:
<img width="857" alt="Screen Shot 2019-06-10 at 10 58 51 AM" src="https://user-images.githubusercontent.com/12846286/59180928-ccca8a80-8b6e-11e9-8961-39d7ac5e31d3.png">

There should be border of the whole component and only bottom border of the toolbar component.

cc: @joneff @elena-gancheva @kspeyanski

joneff:
Fixed by #773

PR #773, opened 2019-06-10
Fix editor toolbar double border in Bootstrap theme

KeiBlag:
Related to https://github.com/telerik/kendo-themes/issues/771


packages/bootstrap/scss/editor/_layout.scss
```diff
@@ -1 +1,6 @@
-@import "~@progress/kendo-theme-default/scss/editor/layout";
\ No newline at end of file
+@import "~@progress/kendo-theme-default/scss/editor/layout";
+
+// Base
+.k-editor .k-toolbar {
+    border-width: 0 0 1px 0;
```
elena-gancheva: Check if you can use a variable here on the `1px` value

Issue #357, opened 2018-10-09
Dialog missing .k-overlay with opacity: 0.5


svetq:
## I'm submitting a...
* Bug report 

## Current behavior
When using a window above, which we want to display a Dialog, utilizing the appendTo property of the open method of the dialogService, the class .k-overlay has opacity: 0 because of the following css rule:

> .k-window .k-overlay {
>    position: absolute;
>    opacity: 0;
> }

## Minimal reproduction of the problem with instructions
Check the following example:
[https://stackblitz.com/edit/angular-uozsuy-rkd3cl?file=app%2Fcustom-window.component.ts](https://stackblitz.com/edit/angular-uozsuy-rkd3cl?file=app%2Fcustom-window.component.ts)

## Environment
Product:

* Kendo Ui for Angular

## Workaround:

add the following style:
>    .k-window .k-overlay{
>      opacity: 0.5;
>    }

joneff:
The reasoning behind the styles is that if we have a dialog inside a dialog, we don't wan't the next dialog to apply any more overlaying.

Imagine the case with dialog with overlay and two nested dialogs -- you would get a 87.5 % combined opacity, which would mostly make the rest of the page not visible.

Since we can not possibly know the intentions of the developer, the workaround with component styles is the only legitimate solution.

Question: is the issue only triggered upon appendTo inside a window and is the scenario in question going to work with the same if the appendTo was in the body, per se?

tsvetomir:
Appending to the body [would work](https://stackblitz.com/edit/angular-uozsuy-zx9tzy?file=app/custom-window.component.ts), as long as we get the z-index right.

Currently, the Window will increment it's z-index for each new instance, but this does not apply to the Dialog. That's probably something we should fix first.

elena-gancheva:
Adding @Raisolution to the discussion and for now, we will not merge the fix in the themes.

Raisolution:
I am OK with not merging this too. The Dialog and Window components has always been like the chicken and the egg - each time one or the other should come first depending on the current use case. I am currently considering to implement something like a `zIndexFn` input for the Window, which will serve as a function overriding the default z-index logic and setting it explicitly. If the z-index of the Dialog is also completely extracted from the themes and moved to a component logic instead, such cases will be handled easily by the developer, without CSS overrides. What do you think about this approach?

tsvetomir:
[...]

PR #782, opened 2019-06-13
Fix: background-opacity-of-dialog-in-window

KeiBlag:
Related to https://github.com/telerik/kendo-themes/issues/357


packages/default/scss/window/_layout.scss
```diff
@@ -15,6 +15,10 @@
             position: absolute;
             opacity: 0;
         }
+
```
joneff: As per @tsvetomir's suggestion, remove the next lines and comment out the above lines.

KeiBlag: Fixed.

Issue #692, opened 2019-04-12
Shadow for frozen columns is too wide


tsvetomir:
## I'm submitting a...
* Bug report 

## Current behavior
The [design spec](https://github.com/telerik/kendo-ux/blob/master/themes/material/grid.psd) for the Material theme specifies a shadow over the non-locked columns:

![image](https://user-images.githubusercontent.com/247917/56047288-c5912700-5d4d-11e9-975b-09a597fbb8a9.png)

In reality the shadow it's around 10px wider:

![image](https://user-images.githubusercontent.com/247917/56049143-7f3dc700-5d51-11e9-902e-542d8947f0dc.png)

## Expected behavior
The shadow should be of the specified size.

## Minimal reproduction of the problem with instructions
1. Open [the Kendo UI for Angular](https://www.telerik.com/kendo-angular-ui/components/grid/columns/locked/) or [Kendo UI for jQuery](https://demos.telerik.com/kendo-ui/grid/frozen-columns) locked (frozen) column examples.
2. Set the theme to Material

KeiBlag:
Fixed with https://github.com/telerik/kendo-themes/pull/791

PR #791, opened 2019-06-17
Fix: Grid Shadow Width for Frozen Columns

KeiBlag:
Related to https://github.com/telerik/kendo-themes/issues/692


packages/material/scss/grid/_theme.scss
```diff
@@ -6,6 +6,7 @@
     $grid-edit-cell-bg: #f5f5f5 !default;
     $grid-edit-cell-shadow: 0 0 5px 3px rgba(0, 0, 0, .12) !default;
     $grid-locked-column-shadow: inset 18px 0 5px -15px rgba(0, 0, 0, .14) !default;
+    $grid-locked-column-shadow-init: inset 18px 0 4px -16px rgba(0, 0, 0, .14) !default;
```
elena-gancheva: why are we introducing a new variable here?

Issue #801, opened 2019-06-25
ColorPicker component with gradient view has missing margins in RTL


elena-gancheva:
When in RTL mode, the Colorpicker component with a gradient view has missing margins.

See the image:
![image](https://user-images.githubusercontent.com/8842674/60094797-b66b2400-9755-11e9-90b6-56db489eaa1d.png)

## I'm submitting a...
* Bug report

## Environment
Product:
* Kendo Ui for Angular

Browser:

- Chrome (desktop) version XX
- Chrome (Android) version XX
- Chrome (iOS) version XX
- Firefox version XX
- Safari (desktop) version XX
- Safari (iOS) version XX
- IE version XX
- Edge version XX

PR #804, opened 2019-06-27
Fix(colorpicker): missing margins in RTL 

KeiBlag:
Related to https://github.com/telerik/kendo-themes/issues/801


elena-gancheva:
Move the colorpicker RTL test in a separate `colorpicker-rtl.html` file, so that the current one will be not this long and cut from the screenshot

packages/default/scss/slider/_layout.scss
```diff
@@ -256,6 +256,10 @@ $slider-draghandle-active-scale: 1 !default;
         }
     }
 
+    [dir="rtl"].k-slider-vertical .k-slider-track {
```
elena-gancheva: The selector and the applied styles should be different. The Colorpicker in RTL for the Material theme is not looking as desired, please change that

KeiBlag: OK

packages/default/scss/colorpicker/_layout.scss
```diff
@@ -230,6 +231,11 @@
             }
             &.k-alpha-slider {
                 margin-right: $colorpicker-margin / 2;
+
+                &[dir="rtl"] {
```
elena-gancheva: The `[dir="rtl"]` style selector should be applied one the same level as `k-flatcolorpicker` and **not** on every slider element. Then all the elements that need alignment under RTL mode, should be nested in it. 

```suggestion
             [dir="rtl"] .k-flatcolorpicker {
                 // styles of sliders go here
             }   
```

See the rendering: 
image

Check here, for exampe https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/notification/_layout.scss#L58

packages/default/scss/colorpicker/_layout.scss
```diff
@@ -282,6 +285,19 @@
             text-align: center;
         }
     }
+
+    [dir="rtl"].k-flatcolorpicker {
```
TeyaVes: What if the rtl direction is set to the whole page and not to the colorpicker widget itself? We also have a class .k-rtl which we use for targeting as well

elena-gancheva: The `k-rtl` class is added to the styles. Since the jquery colorpicker is with different layout than the angular one, we should add feature for styling the jquery colorpicker in RTL mode. This PR includes angular colorpicker rlt styles + added visual tests for both suites. The logged item is here https://github.com/telerik/kendo-themes/issues/808

Issue #817, opened 2019-07-15
Buttons overlap in Grid when in editing mode


keiblag:
## I'm submitting a...

* Bug report 

## Current behavior
When the Grid component is in editing mode and the width of the column that holds the command buttons is not big enough, they overlap.

![grid](https://user-images.githubusercontent.com/10484594/61212509-2cd2c480-a70b-11e9-8f50-31255a0c7722.PNG)

## Expected behavior
The buttons shouldn't overlap.

## Minimal reproduction of the problem with instructions
https://www.telerik.com/kendo-angular-ui/components/grid/editing/editing-reactive-forms/

  1. Click button Add new
  2. In column command see the buttons overlap

## Environment
Product:

* Kendo Ui for Angular

TeyaVes:
duplicate of https://github.com/telerik/kendo-themes/issues/239

joneff:
Fixed by #1226

PR #820, opened 2019-07-16
[WIP] Fix buttons overlap in Grid editing mode

keiblag:
Targets https://github.com/telerik/kendo-themes/issues/817


KeiBlag:
If the column is reduced in size (width) and cannot hold the two buttons, they are set in two rows without space between them. This applies to all themes. 
grid-default


I suggest we change the rule in the default theme that removes the margin: 
```CSS
.k-command-cell > .k-button {
      margin-top: $grid-form-component-offset;
      margin-bottom: $grid-form-component-offset;
}

@joneff  please advise. 
cc @elena-gancheva @TeyaVes

joneff:
Superseded by #1226

PR #842, opened 2019-08-06
Filter: initial commit with styles

TeyaVes:
targets: https://github.com/telerik/kendo/issues/9887


joneff:
@TeyaVes markup can be greatly improved, but the rest is fine.

PR #861, opened 2019-08-27
Angular toolbar splitter

PetarMetodiev:
In this PR:
 - Fix issue where a separator in the toolbar was receiving additional `margin-left: 8px` in angular specific rendering.
 - Styling of separator now takes into account updated angular rendering.


packages/default/scss/toolbar/_layout.scss
```diff
@@ -158,6 +158,11 @@
             &:not(:last-child) {
                 margin-right: $toolbar-padding-x;
             }
+
+            .k-separator {
+                margin: 0;
```
joneff: do not reset top and bottom margin. use `margin-left` and `margin-right` instead

tests/visual/toolbar.html
```diff
@@ -665,6 +665,82 @@
             </div>
         </div>
 
+        <div class="section" style="clear: both;">
```
joneff: remove all inline styles and attributes that are not needed for the test

PR #863, opened 2019-09-02
Timeline: style fixes

TeyaVes:
targets: https://github.com/telerik/kendo/issues/9874


packages/default/scss/card/_layout.scss
```diff
@@ -164,6 +164,11 @@
             border-color: inherit;
         }
 
+        .k-button {
+            max-width: 100%;
```
PreslavKozovski: Add a test for this bug

PR #871, opened 2019-09-05
feat(progressbar): add styles for progress label position

joneff:
Note: ~~currently, IE 11 is not so supported. I will add a follow up commit / branch to handle it.~~ Works in IE as well

There is no implicit position for the label. Instead, it relies on explicit label position.

/cc @Raisolution


tests/visual/progressbar.html
```diff
@@ -161,22 +362,52 @@
                 </ul>
             </div>
 
-            <span></span>
-            <span>LTR</span>
-            <span>RTL</span>
-            <span>Indeterminate</span>
```
PreslavKozovski: Why did you remove the Indeterminate progressbar?

joneff: re-added

Issue #896, opened 2019-09-18
buttongroup in k-button-group class has incorrect spacing on selection


PetarMetodiev:
## I'm submitting a bug

* Bug report 

## Current behavior
See image
![image](https://user-images.githubusercontent.com/6378306/65146252-063f4b80-da24-11e9-8751-32a2ce854ad3.png)

## Environment
Product:
* Kendo Ui for Angular

PR #898, opened 2019-09-19
fix(toolbar): buttongroup style in responsive menu

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/896


tests/visual/toolbar.html
```diff
@@ -587,6 +587,45 @@
             </div>
         </div>
 
+        <div class="section" style="float: left">
```
joneff: we shouldn't really use `float` styles.

PR #904, opened 2019-09-20
Speedup gulp

joneff:
Gulp initialization takes too much time due to dependency loading. By refactoring the code to flatten gulp tasks as much as possible, we can achieve faster gulp startup and ultimately faster tasks execution.

While I've changed things, it does not change how things work or compile, so no fix / feat.


packages/bootstrap/package.json
```diff
@@ -19,14 +19,19 @@
...
     "lint": "gulp lint",
     "api": "gulp api",
     "api-check": "npm run api && git diff --exit-code --quiet -- docs/ || (echo -e '\\033[0;31mERROR: API docs are out of date' && exit 1)",
-    "build": "gulp sass",
-    "dart": "gulp dart",
-    "watch": "gulp sass:watch",
     "embed-assets": "gulp assets",
-    "swatches": "gulp swatches",
     "test": "npm run lint && npm run build && npm run api-check",
     "prepublishOnly": "bash ./build/embed-dependencies 'bootstrap' '@progress/kendo-theme-default' && npm run build && npm run swatches",
```
tsvetomir: I'd use a gulp task that lists "sass" and "sass:swatches" as dependencies instead of calling them indirectly.

joneff: agreed

joneff:
I've optimized task so that related tasks are run in conjunction, not one after another.

There is a new prepublish task that lints, builds themes in sass and dart and checks for api errors. That increases push time, yes, but ensures that the code will most likely work when deployed to travis.

PR #906, opened 2019-09-24
Housekeeping

joneff:
* simplify foreach.sh (which we use in lieu of lerna locally) to use --if-present
* update npmignore
* remove redundant root scripts
* use default sasslint config location (`.sasslintrc`)

This is just a housekeeping PR. There is nothing actually changed. Extracted from #904, which got quite big.


build/foreach.sh
```diff
@@ -5,27 +5,8 @@ set -e
...
-    for theme in default bootstrap material
-    do
-        cd $dir/packages/$theme
-        npm run $1
-    done
-fi
+for pkg in default bootstrap material theme-tasks
+do
+    cd $dir/packages/$pkg
+    npm run $1 --if-present
```
tsvetomir: Nice touch with `--is-present`, we used empty scripts for that purpose elsewhere.

PR #909, opened 2019-09-24
Flatten theme files for dist

### LINKED ISSUE #423: compilation of SCSS theme takes too much time
## I'm submitting a...

* Suggestion for improvement

## Current behavior

Building the SCSS theme takes around 30 seconds.

## Minimal reproduction of the problem with instructions
Create a start up Angular 7 project and load just the Kendo Default theme in angular.json as follows:

```
"styles": [
    "node_modules/@progress/kendo-theme-default/scss/all.scss"
]
```

run ng build --prod

## Environment
Product:
* Kendo Ui for Angular

joneff:
Note: Obviously, this needs to be rebased after #904 is merged. Pay attention to the last two commits.

It's been an issue for a while now, compiling the themes outside of the scope of our development takes too much time. That's why I've added functionality to inline / flatten / squash / whatever the themes in a single mostly self-contained file that can be included.

The results are sub 1s compile time.

Since we are not using `_index.scss` convention anywhere, we could actually have the current `_all.scss` files renamed to `_index.scss` and the have `_all.scss` generated. We could even do it for components.

~~I am open to suggestions how to name the tasks. `inline-sass-files` doesn't sound right. Perhaps "flatten"?~~ Renamed task to `flatten-sass-files`.

* fixes #423 and probably others

/cc @gyoshev thoughts?

### DIFF
--- a/packages/theme-tasks/gulpfile.js
+++ b/packages/theme-tasks/gulpfile.js
@@ -16,6 +16,7 @@ const Fiber = require("fibers");
 const sassdoc = require("sassdoc");
 const eslint = require("gulp-eslint");
 const sasslint = require("gulp-sass-lint");
+const baka = require("@joneff/baka");
 const sassImporterFactory = require("./lib/sassimporter");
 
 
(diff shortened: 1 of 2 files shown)


gyoshev:
“Flatten” sound good (maybe better than “baka” :wink:). Congrats on the improved compile time!

PR #911, opened 2019-09-26
fix(grid): page scroll on column chooser item click

danielkaradachki:
https://github.com/telerik/kendo-angular/issues/2476


joneff:
That's an issue (required relatively positioned wrapper) hopefully we won't have when we switch to the new styles for checkbox / radio

Issue #859, opened 2019-08-26
Blazor Menu is not styled correctly in Bootstrap theme.


matanasov:
It seems like some additional padding is added to the top and bottom of the sub-menu:

![MenuBootstrap](https://user-images.githubusercontent.com/5861835/63698721-1b132f80-c828-11e9-8471-47b0fb8d2fec.png)

joneff:
There is different rendering between kendo-jquery and the rest of the suites, hence the difference in appearance. I guess we could tweak kendo-jquery rendering to target the issue

PR #916, opened 2019-09-30
Menu: submenu double padding in blazor bootstrap theme

TeyaVes:
targets: https://github.com/telerik/kendo-themes/issues/859


TeyaVes:
> The padding must remain, since it's part of bootstrap framework. See https://getbootstrap.com/docs/4.3/components/navbar/#supported-content. Consider using `margin-top` with negative values for nested popups

I changed the .k-popup to .k-menu-popup and in this way we'll target the menu within a menu popup only and remove the double paddings from there. @joneff , what do you think about that?

Issue #140, opened 2018-04-24
Broken appearance when multiple sliders with labels are placed on top of each other


danielkaradachki:
Example: https://dojo.telerik.com/@SiliconSoul/IYUloGaZ

gyoshev:
This might require a few changes in the slider layout and rendering to allow it to expand when there are ticks. Here is [an example](https://dojo.telerik.com/iHENENEB) with preliminary version of the styles that could allow this.

JoomFX:
This one seems to be fixed already.
[Here's how it looks](https://dojo.telerik.com/@ivalkov/IWIjobaj) with the current theme's version.

joneff:
With less themes, yes. sass themes remain https://dojo.telerik.com/@joneff/Adotaviy

The issue is related to `.k-slider-horizontal .k-tick` having top margin. What happens is that it has a height of 100% and then additional top margin which in terms in height that's always more than 100% and float issues appear.

PR #925, opened 2019-10-03
fix(slider): multiple display-block sliders appearance

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/140


packages/default/scss/slider/_layout.scss
```diff
@@ -344,7 +343,8 @@ $slider-draghandle-active-scale: 1 !default;
         }
 
         .k-slider-horizontal .k-slider-buttons & {
-            padding-top: 0;
+            padding: 2px 0;
+            box-sizing: border-box;
         }
```
joneff: It's quite possible that the issue is related to slider without buttons as well

Issue #929, opened 2019-10-04
Disabled Textarea still has hover effect


akloeber:
**Describe the bug**
A Textarea which is disabled still has a hover effect.

**To reproduce**
1. Go to https://stackblitz.com/edit/angular-gh8hud-jp1yno
2. Move cursor above textarea

**Expected behavior**
There should not be a hover effect.

**Affected package (please provide)**
* theme-default

**Affected suites (please provide)**
* Kendo UI for Angular
* _maybe others_

**Affected browsers (please provide)**
- Chrome: [for example, 77]
- Safari: [for example, 12]

**Additional context**
For other TextBoxes the hover effect is suppressed via `pointer-events: none`:
```css
.k-textbox:disabled, .k-textbox[disabled], .k-textbox.k-state-disabled {
  pointer-events: none;
  /* ... */
}
```
Unfortunately this style gets not applied here as the Textarea only has class `k-textarea`. Something like this is missing:
```css
.k-textarea:disabled, .k-textarea[disabled], .k-textarea.k-state-disabled {
  pointer-events: none;
}
```

dmanova:
Verified - the cursor/pointer is not changed on hover. Only few themes ( that are not used in Kendo UI for Angular, e.g metro, flat, highcontrast) change border color on hover.

PR #934, opened 2019-10-08
Textarea: add disabled styles

TeyaVes:
targets: https://github.com/telerik/kendo-themes/issues/929


TeyaVes:
> Add a visual test :)

I can't really add a test for that, as I'm disabling the pointer events on the disabled textarea and I can't demonstrate it with static html

Issue #663, opened 2019-03-15
Double borders in RTL in the agenda view


danielkaradachki:
Example: https://stackblitz.com/edit/angular-biqhns-mzmjw9?file=app/app.component.ts

PR #939, opened 2019-10-09
fix(scheduler): remove double border

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/663

Note: If the fix works fine, without side effects, we will discuss the _`!important` vs ugly-selector_ approach.


packages/default/scss/scheduler/_layout.scss
```diff
@@ -811,6 +811,11 @@
             td.k-first {
                 border-left-width: 0;
             }
+
+            // sass-lint:disable no-important
```
joneff: This will disable sass-lint for the entire file. Use `// sass-lint: disable-line no-important` for the specific line

packages/default/scss/scheduler/_layout.scss
```diff
@@ -1046,6 +1051,11 @@
                         border-right-width: 1px;
                     }
                 }
+
+                // sass-lint:disable no-important
+                .k-scheduler-datecolumn {
+                    border-right-width: 0 !important;
```
joneff: same as above

joneff:
We need to add a test for that:

image

Issue #836, opened 2019-07-31
margin-left is set to MultiSelect tags in Material theme


IvanDanchev:
* Regression report 
Introduced in 2019.2.514.

## Current behavior
**margin-left** causes selected items on second + rows in the MultiSelect to appear unaligned to the items in the first row.

## Reproduction
[Dojo ](https://dojo.telerik.com/EPUPudun) example. [Screenshot](https://www.screencast.com/t/OWzfiQZboyAf).

## What is the motivation or use case for changing the behavior?
Regression

## Environment
* Kendo Ui for jQuery

Browser:
- Chrome (desktop) version XX
- Firefox version XX
- Edge version XX

PR #962, opened 2019-10-15
Multiselect: tags margin in Material

PreslavKozovski:
related to: https://github.com/telerik/kendo-themes/issues/836


packages/material/scss/multiselect/_layout.scss
```diff
@@ -13,21 +13,27 @@
             }
 
             & + .k-button {
-                margin: ($input-padding-y / 2) 0 0 ($input-padding-y / 2);
+                margin: ($input-padding-y / 2) ($input-padding-y / 2) 0 0;
```
joneff: This rule simply duplicates the above one. If that's the fix, you need to remove the rule.

packages/material/scss/multiselect/_layout.scss
```diff
@@ -13,21 +13,27 @@
...
 
             .k-button {
                 margin-left: ($input-padding-y / 2);
                 margin-right: 0;
 
+                & + .k-button {
+                    margin-left: ($input-padding-y / 2);
+                    margin-right: 0;
+                }
+
```
joneff: likewise -- the exact same rule is specified above.

Issue #844, opened 2019-08-07
Migrate k-drag-clue styles from LESS themes


Stamo-Gochev:
## I'm submitting a...
* Suggestion for improvement

## Current behavior
The `.k-drag-clue` does not have `position: absolute` and additional properties like `z-index` in the [SASS themes](https://github.com/telerik/kendo-themes/blob/720e51b7ba90c792e069ee796e28fd2022b203e0/packages/default/scss/grid/_layout.scss#L416). The properties are available in the [LESS themes](https://github.com/telerik/kendo-themes/blob/720e51b7ba90c792e069ee796e28fd2022b203e0/packages/default/scss/grid/_layout.scss#L471-L474) and these styles are currently inlined in [ng2](https://github.com/telerik/kendo-angular-grid/blob/654981b9c5b24599df00bcc52cfe974bfe7a66a7/src/dragdrop/drag-hint.service.ts#L24) and [react](https://github.com/telerik/kendo-react-private/blob/4d9746a2fcb0a1d2df3915f4c0865056cdecd27c/packages/grid/src/drag/DragClue.tsx#L29). This is necessary when grouping a grid and the drag clue is shown over the group panel.

## Expected behavior
The necessary styles are added to the SASS themes

## Environment
Product:
* Kendo Ui for jQuery
* Kendo Ui for Angular
* Kendo Ui for React
* Kendo Ui for Vue

Package versions:
**kendo-theme-default**: 4.1.0, but applies to bootstrap and material

Browser: All

PR #963, opened 2019-10-15
Dragclue: position and z-index styles

PreslavKozovski:
targets: https://github.com/telerik/kendo-themes/issues/844


PreslavKozovski:
The appearance of the Dragclue is a different issue, I believe we will address it when the new design is ready - https://github.com/telerik/kendo-ux-private/issues/47

PR #984, opened 2019-10-21
refactor(combobox): extract common styles

### LINKED ISSUE #982: Extract all Combobox styles from common logic
**Is your feature request related to a problem? Please describe.**
Currently, the styles responsible for `layout`/`theme` sizings, colors, border-widths, etc. for the Combobox component are separated in global `common` files.

Next step, extract all variables for the Combobox component, that are including appearances and depending on other styles.

**Describe the solution you would like to see implemented**
Extract those styles from the common files and apply them in the corresponding for the component style folders. This refactoring must **not** regress the current design and behavior.

**Additional context**
We can achieve decoupled styles and more flexibility when customizing the styles for a given component. This enhancement will be needed for creating a consistent and flexible styleguide for creating forms with Kendo components.

Related to: https://github.com/telerik/k2/issues/150

JoomFX:
closes #982


packages/material/scss/combobox/_theme.scss
```diff
@@ -1,3 +1,5 @@
 @include exports( "combobox/theme" ) {
-
+    .k-combobox {
```
elena-gancheva: Here we can import the combobox `_theme.scss` file from the Default theme, where the `background-color: transparent;` property is already set.  If some styles need to be specific for the Material theme for the combobox, the exports namespace here should be `combobox/theme/material`

PR #1015, opened 2019-10-28
refactor(numerictextbox): extract common styles

silviyaboteva:
closes: https://github.com/telerik/kendo-themes/issues/1003

### DIFF
--- a/packages/bootstrap/scss/_variables.scss
+++ b/packages/bootstrap/scss/_variables.scss
@@ -565,6 +565,39 @@ $autocomplete-focused-shadow: $input-focused-shadow !default;
 // NumericTextBox
 $numerictextbox-spinners-focused-shadow: null !default;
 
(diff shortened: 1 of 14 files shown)


silviyaboteva:
After a quick discussion with @joneff, there are two cases of applying the `k-state-hover`, `k-state-focused` and `k-state-disabled`. The first one is on `k-numerictextbox` span element and the other is on `k-numeric-wrap`. Should we unify this somehow? 
There are existing examples on both approaches in inputs.html file (where the classes are attached to the element with `k-numerictextbox` class), and in grid-editing.html (on `k-numeric-wrap`).

Another thing that is worth mentioning is that currently, the height of the numeric textbox wrapper (and other components like DropDown, ComboBox, DropdownTree etc.) in the Material theme is 35px, but I believe that it must be 36px instead. What can we do about this?

@elena-gancheva @joneff @JoomFX

Issue #989, opened 2019-10-22
Card component styles additions


elena-gancheva:
Currently, there are missing styles for full customization of card component and some of the existing styles should be refactored so that they have complete consistency with the component's api.

**Styles for refactoring**
- `k-card-state-{error/info/warning/success}` - those are not states of the card. This class should be renamed to something like `k-card-{success/error/warning/info/default}`
- card action buttons layout - if set to `stretched`, the styles are depending on having a class `k-card-action` on each element within the `k-card-actions` class. This is not flexible, if the user has configured the card's actions with a template and custom classes.
- `k-hr` should be renamed to `k-card-separator` - separate card content with a `hr` or styles, and should be `vertical` or `horizontal`, depending on the layout of the card

**Styles for adding**
- `k-card-thumbnail` (avatar, thumbnail, circled image for the card header, should be aligned with the card title and subtitle within a `k-hbo` container with `flex-shrink: 0`; Should set `background-image` property with the proper sizings for it.
- `k-card-actions-{start/center/end/stretched}` - the position of the card action buttons, aligned horizontally (same for `k-card-vertical-actions`)
- `k-card-footer` - this class is missing. Should be same as header, body with paddings and border.

@joneff @Juveniel

Juveniel:
I agree with all of the above.

Suggestion: a divider is currently available in the themes as `.k-hr`, so we could move this bullet to the refactoring column and just create a `.k-card-separator` class for the cards that extends it?

elena-gancheva:
Updated, thanks :)

elena-gancheva:
Related to https://github.com/telerik/kendo-themes/issues/176 - discussion about card-deck, card-list and widths of the card in a container

PR #1017, opened 2019-10-28
Chore(cards): add additional options

TeyaVes:
targets: https://github.com/telerik/kendo-themes/issues/989


elena-gancheva:
1. `k-card-horizontal` and `k-card-vertical` are switched `flex-direction: row/col;`
2. card actions – `k-card-actions-cente`r is missing `justify-content: center;` (for both horizontal and vertical alignment of the card actions)
3. vertical card styles are missing
4. helper classes for different card configurations
5. **To add** – `k-card-thumbnail` class, set default background color style (**To consider** – if the thumbnail is a `image`, the card component will render a `div` element with a source set through `background-image`, so if we apply styles for the image position within this element, they will no apply on the `k-card-thumbnail` container. Those styles should be applied to the inner element)
6. **To add** – `k-card-thumbnail-round` – should make `border-radius: 50%;` (configuration of the component – can be a circle or a square shape)
7. **To add [under consideration]** - `k-card-thumbnail-{type}` where type is `image`, `initials`, `icon` (should apply styles for vertical and horizontal alignment of the content in the thumbnail

PR #1033, opened 2019-10-30
refactor(inputs): replace appearance mixin with fill and box-shadow

silviyaboteva:
closes: https://github.com/telerik/kendo-themes/issues/1032


packages/bootstrap/scss/input/_theme.scss
```diff
@@ -4,40 +4,50 @@
...
         &.k-invalid,
         &.k-state-invalid,
         &.ng-invalid.ng-touched,
         &.ng-invalid.ng-dirty {
-            color: $error;
-            border-color: $invalid-border;
+            @include fill (
+                $color: $error,
+                $border: $invalid-border
+            );
```
joneff: We should definitely have variables for that, but at a later point

packages/bootstrap/scss/input/_theme.scss
```diff
@@ -4,40 +4,50 @@
...
         }
 
         // Selection
         &::selection {
-            background-color: $accent;
-            color: $accent-contrast;
+            @include fill (
+                $color: $accent-contrast,
+                $bg: $accent
+            );
```
joneff: Same applies for selection, as well

Issue #1023, opened 2019-10-29
Sorted locked column does not have background in the header cell.


Xizario:
Replicates https://www.telerik.com/kendo-react-ui/components/grid/
**Sort** by **ID** column.

Works in **Default** and **Material** themes. 
Broken in **Bootstrap**.
<img width="234" alt="Screenshot 2019-10-29 at 8 39 23" src="https://user-images.githubusercontent.com/9458606/67743548-a75bf180-fa27-11e9-9f7c-3b7395dc1e9e.png">

PR #1036, opened 2019-10-31
fix(grid): sorted and selected locked columns background color issues

silviyaboteva:
closes: https://github.com/telerik/kendo-themes/issues/1023
closes: https://github.com/telerik/kendo-react/issues/372

_Additional note:_ I applied the selector also in Default and Material themes as if in future we decide to put the following rule:
```
col.k-sorted,
th.k-sorted {
    background-color: $grid-sorted-bg;
}   
```
inside `.k-grid`, to prevent the same inexpected behavior.


silviyaboteva:
@joneff quick question: Does the second fix is ok (to apply colors with no opacity) and extract them as variables? Please ping me if there is a better solution to this problem.

Issue #908, opened 2019-09-24
Add additional how-to topic for DartSass compilation


svetq:
**Describe the issue or request**
The current compilation of the project should include postcss-calc in the start up pipeline so that the values computed with calc() are proper.

joneff:
Related to #846, #834

joneff:
We've provided documentation for:

* [Compiling themes](https://github.com/telerik/kendo-themes/wiki/Compiling-themes)
* [Customizing themes](https://github.com/telerik/kendo-themes/wiki/Customizing-themes)
* [Including themes in React projects](https://github.com/telerik/kendo-themes/wiki/Including-themes-in-React-projects)
* [Including themes in Angular projects](https://github.com/telerik/kendo-themes/wiki/Including-themes-in-Angular-projects)

PR #1045, opened 2019-11-06
fix: calc(8px - 0) invalid property value~

IKKI2000:


joneff:
Take a look at https://github.com/telerik/kendo-themes/issues/908, we can not simply set the value in question to `0px`, since somebody can override it again to be `0`.

Instead, you should include postcss-calc in yout build process to ensure that the value is dropped from the final css. If you are using gulp to build your styles, you can take a look at https://github.com/telerik/kendo-themes/blob/develop/packages/theme-tasks/gulpfile.js how to include it. If you are using webpack, it does differ a bit, but is still doable.

We'll have documentation on how exactly to do it with webpack, gulp and parcel.

PR #1046, opened 2019-11-06
chore: trigger image tests

joneff:
On November 5th, 2019, image tests were not present for a while, so by merging PR's I've introduced several changes in the baseline images. This PR aims to address said changes.


joneff:
We'll need to revisit material screenshots after the refactoring is completed

PR #1049, opened 2019-11-12
fix(treelist): add left/right borders to multicolumn headers

silviyaboteva:
https://github.com/telerik/kendo-react-private/pull/825


silviyaboteva:
@joneff since the treelist rendering in React is different than the original one (introducing only one table instead of two), we hit some problems with the styles. Could we think of some general solution? What do you think?

Issue #680, opened 2019-04-04
Dataviz styles failing when building with Bootstrap 4.3.1 and Kendo Boostrap 4.2.0


jcastillo-ait-asia:
## I'm submitting a...

* Bug report

## Current behavior

Module build failed at

```
ERROR in ./src/styles.scss
Module build failed (from /media/user/external/git/uber-spa/node_modules/mini-css-extract-plugin/dist/loader.js):
ModuleBuildError: Module build failed (from /media/user/external/git/uber-spa/node_modules/sass-loader/lib/loader.js):

            top: calc( #{($padding-y-sm*2) + ($line-height*$font-size)} + 1px);
                                                                             ^
      Incompatible units: 'px' and 'rem'.
      in /media/user/external/git/uber-spa/node_modules/@progress/kendo-theme-bootstrap/modules/@progress/kendo-theme-default/scss/dataviz/_layout.scss (line 350, column 79)
    at runLoaders (/media/user/external/git/uber-spa/node_modules/webpack/lib/NormalModule.js:301:20)
    at /media/user/external/git/uber-spa/node_modules/loader-runner/lib/LoaderRunner.js:367:11
    at /media/user/external/git/uber-spa/node_modules/loader-runner/lib/LoaderRunner.js:233:18
```

## Environment
```
Angular CLI: 7.3.7
Node: 11.13.0
OS: linux x64
Angular: 7.2.11
... animations, common, compiler, compiler-cli, core, elements
... forms, http, language-service, platform-browser
... platform-browser-dynamic, router, service-worker

Package                            Version
------------------------------------------------------------
@angular-devkit/architect          0.13.7
@angular-devkit/build-angular      0.13.7
@angular-devkit/build-ng-packagr   0.13.7
@angular-devkit/build-optimizer    0.13.7
@angular-devkit/build-webpack      0.13.7
@angular-devkit/core               7.3.7
@angular-devkit/schematics         7.3.7
@angular/cdk                       7.3.6
@angular/cli                       7.3.7
@angular/flex-layout               7.0.0-beta.24
@angular/material                  7.3.6
@angular/pwa                       0.13.7
@ngtools/json-schema               1.1.0
@ngtools/webpack                   7.3.7
@schematics/angular                7.3.7
@schematics/update                 0.13.7
ng-packagr                         4.7.1
rxjs                               6.4.0
typescript                         3.2.4
webpack                            4.29.0
```

Kendo dependencies:
```
"bootstrap": "4.3.1",
[...]

PR #1060, opened 2019-11-18
use html calc method to calculate treemap-title sizes

feitzi:
Should fix the error telerik/kendo-themes#680

Instead of using calculation the values for top and left with scss use the browser calc method.  
This is required if you use a `rem` or `em` font-size value


joneff:
@feitzi , can you rebase your branch from latest develop and change the name of the commit to `fix(treemap): use html calc method to calculate treemap-title-sizes`. We use commitzen to extract the change log based on the commit messages

packages/default/scss/dataviz/_layout.scss
```diff
@@ -347,10 +347,10 @@ $chart-title-font-size: 1.143em !default;
             right: 0;
         }
         .k-treemap-title + .k-treemap-wrap {
-            top: calc( #{($padding-y-sm*2) + ($line-height*$font-size)} + 1px);
+            top: calc( #{($padding-y-sm*2)} + (#{$line-height}*#{$font-size}) + 1px);
```
joneff:
```suggestion
            top: calc( #{$padding-y-sm * 2} + (#{$line-height} * #{$font-size}) + 1px);
```

packages/default/scss/dataviz/_layout.scss
```diff
@@ -347,10 +347,10 @@ $chart-title-font-size: 1.143em !default;
...
         }
         .k-treemap-title-vertical + .k-treemap-wrap {
-            left: calc( #{($padding-y-sm*2) + ($line-height*$font-size)} + 1px);
+            left: calc( #{($padding-y-sm*2)} + (#{$line-height}*#{$font-size}) + 1px);
```
joneff:
```suggestion
            left: calc( #{$padding-y-sm * 2} + (#{$line-height} * #{$font-size}) + 1px);
```

feitzi: done

packages/default/scss/dataviz/_layout.scss
feitzi: done

feitzi:
I squashed the commits and added your change requests.

Issue #1034, opened 2019-10-30
Add Grid Pager NumericTextBox specific CSS


dtopalov:
**Is your feature request related to a problem? Please describe.**
The following Grid Pager enhancement requires a minor addition to the themes:

https://github.com/telerik/kendo-angular-grid/pull/858

**Describe the solution you would like to see implemented**
The following styles need to be applied to the NumericTextBox component that replaced the previously used regular Input:

<kendo-numerictextbox
            [style.margin]="'0 1ex'"
            [style.width]="'3em'"

The element can be targeted as ".k-pager-input .k-numerictextbox"

**Additional context**
Add any other context or screenshots about the feature request.

Adding any of the predefined input-related classes to the NumericTextBox does not yield the desired results, so we need to apply the `margin` and `width` rules preferably without inlining them.

PR #1062, opened 2019-11-18
refactor: add grid pager numerictextbox style

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/1034

### DIFF
--- a/packages/default/scss/grid/_layout.scss
+++ b/packages/default/scss/grid/_layout.scss
@@ -900,6 +900,13 @@
         font-size: inherit;
     }
 
+    .k-pager-input {
+        .k-numerictextbox {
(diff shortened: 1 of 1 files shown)


packages/default/scss/grid/_layout.scss
```diff
@@ -900,6 +900,13 @@
         font-size: inherit;
     }
 
+    .k-pager-input {
+        .k-numerictextbox {
+            margin: 0 1ex;
```
elena-gancheva: I think we should check if there is any variable for the numerictextbox which is responsible for the font-size/height of the element. If there is not, is it a good idea to introduce a variable?

Issue #679, opened 2019-04-03
Angular Scheduler edit-dialog Material theme improvements


elena-gancheva:
Several improvements for Kendo UI for Angular Scheduler edit-dialog in Material theme:

Checkout the the design [specifications](https://github.com/telerik/kendo-ux/blob/master/themes/material/dialogs-scheduler.psd).

- [ ] The width of the dialog content is larger
Current state: 
![image](https://user-images.githubusercontent.com/8842674/55476684-e657cf00-561f-11e9-934f-9b456fb0a4dd.png)

- [x] There are inconsistent spacings between the elements in the edit-form
Current state: 
![image](https://user-images.githubusercontent.com/8842674/55476879-5fefbd00-5620-11e9-8ece-aa2205ca19c4.png)

- [x] Dialog buttons are not aligned as the design specification
Current state: 
![image](https://user-images.githubusercontent.com/8842674/55476926-7dbd2200-5620-11e9-8199-b7751f56c427.png)

- [ ] We should consider small screens, for example adding scrollbars when the dialog's content is overflowing.

- [x] Some labels are not aligned
Current state: 
![image](https://user-images.githubusercontent.com/8842674/55477085-f2905c00-5620-11e9-93bf-f3ec38c45b90.png)

- [ ] Different classes for the different sizes of the dialog `k-dialog-sm, k-dialog-xs, etc` for the Material theme

@joneff @Raisolution @TeyaVes @PreslavKozovski

joneff:
* The components breaking on two lines can be targeted with specific styles. 
* The vertical misalignment of inputs and text can easily be targeted
* The sizing may prove a bit challenging (note: we must extend the sizing map slightly)

The rest is fine.

svetq:
The DropDowns located in the repeat "monthly" and "yearly" tabs are too narrow:
![image](https://user-images.githubusercontent.com/20015190/60576744-a91ded00-9d86-11e9-8133-d457155f8670.png)
and:
![image](https://user-images.githubusercontent.com/20015190/60576891-f5692d00-9d86-11e9-800d-ad6ab47925b9.png)

joneff:
Fixed by applying form guidelines in https://github.com/telerik/kendo-angular-scheduler/pull/321

PR #1066, opened 2019-11-19
fix(scheduler): add scrollbar to edit dialog

JoomFX:
related to https://github.com/telerik/kendo-themes/issues/679


tests/visual/scheduler-angular-edit-dialog.html
```diff
@@ -0,0 +1,233 @@
+<!doctype html>
+<html lang="en" class="k-typography">
+<head>
+    <title>Scheduler Angular Edit Dialog</title>
+    <meta charset="utf-8" />
+    <link rel="stylesheet" href="../../packages/default/dist/all.css" />
+    <link rel="stylesheet" href="assets/styles.css" />
+    <script src="assets/scripts.js"></script>
+
+    <style>
```
elena-gancheva: The test's screenshots are not looking okay. This may be caused of the positioning styles of the dialog and should be handled in the test's styles, just for the test purpose. Please, check this https://github.com/telerik/kendo-themes/blob/develop/tests/visual/dialog.html#L16

Issue #1079, opened 2019-11-27
Add Avatar styles


JoomFX:
**Is your feature request related to a problem? Please describe.**
Styles need to be added for the new [Avatar component](https://github.com/telerik/k2/issues/173).
Part of https://github.com/telerik/kendo-angular-layout/pull/274

**Describe the solution you would like to see implemented**
Avatar styles added to Kendo-Themes

**Describe the alternatives you have considered**
N/A

**Additional context**
N/A

PR #1080, opened 2019-11-27
feat(avatar): add component styles

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/1079


packages/default/scss/avatar/_theme.scss
```diff
@@ -0,0 +1,13 @@
+@include exports("avatar/theme") {
+
+    .k-avatar {
```
joneff: You don't really need this part.

PR #1082, opened 2019-11-27
chore(breadcrumb): resolve rendering and design changes

Juveniel:
targets: telerik/kendo/issues/10318


packages/material/scss/_variables.scss
```diff
@@ -1966,7 +1966,7 @@ $upload-focused-shadow: 0 0 0 2px rgba(0, 0, 0, .13) !default;
 
 
 // Breadcrumb
-$breadcrumb-padding-x: $padding-x-sm !default;
+$breadcrumb-spacer-x: $padding-x !default;
```
joneff:
```suggestion
$breadcrumb-padding-x: $breadcrumb-padding-x !default;
```

Also, add `$breadcrumb-padding-y` variable

packages/default/scss/breadcrumb/_layout.scss
```diff
@@ -2,7 +2,7 @@
 
     // Breadcrumb
     .k-breadcrumb {
-        padding-right: $breadcrumb-padding-x;
+        padding-right: $breadcrumb-spacer-x;
```
joneff: there should be padding all around not just a single direction

PR #1123, opened 2019-12-13
fixed grid search panel input in rtl mode

GhalamborM:


joneff:
@GhalamborM can you elaborate on what you are trying to achieve with this PR? I don't see any references issue. In addition, there are errors preventing the PR from merging.

joneff:
Since grid search does support RTL, and there is no activity in this thread, I am closing the PR.

PR #1131, opened 2019-12-20
Window: add focus styles on focus pseudo-class

TeyaVes:
targets: https://github.com/telerik/blazor/issues/612


joneff:
@TeyaVes we need to add test for that!

Issue #1114, opened 2019-12-09
[ButtonGroup] White border on inner buttons


PetarMetodiev:
**Describe the bug**
Inner buttons inside a buttongroup have a white border-bottom.

**Expected behavior**
No buttons inside a buttongroup should have white border bottom.

**Screenshots**
![ButtonGroup](https://user-images.githubusercontent.com/6378306/70436550-3c73f080-1a92-11ea-9a0f-bbe9663cea9a.png)

**Affected package (please provide)**
* theme-default

**Affected suites (please provide)**
* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Affected browsers (please provide)**
- Chrome

**Build system information (please provide)**
- OS: Windows

elena-gancheva:
Closing this issue, since it is related to the browsers difference in the sub-pixel precision.

PR #1137, opened 2019-12-24
feat(filemanager): add component styles

Juveniel:
Styles for the File Manager component:

spec: https://github.com/telerik/kendo/issues/10278
design: https://github.com/telerik/kendo-ux-private/issues/51

partof: https://github.com/telerik/kendo/issues/10191
targets: https://github.com/telerik/kendo/issues/10436


Juveniel:
@PreslavKozovski, @joneff 

1. The white space in the test pages is a result of browser pixel calculation ([related issue in the themes repository](https://github.com/telerik/kendo-themes/issues/1114).) Resolved by increasing <h> elements size in the test pages.

2. All titles are now unique.

3. Will fix after the Wiki is updated.

4. Actually, there is no Splitter widget initialized in the FileManager, but simple resizable handles. This has been discussed and confirmed with @ivelina-penkova .

5. Added ContextMenu tests. Also added Drag Hint test page. Note that some of the options available in the design are not yet implemented in the Filemanager, thus missing from the test page.

6. Styles for the resizable split bars are available in the Splitter.scss, thus added it as a dependency.

7. Reordered according to all.scss

Issue #816, opened 2019-07-11
Tooltip content is not wrapped in IE


elena-gancheva:
## I'm submitting a...
* Bug report 

## Current behavior
If tooltip's content is larger, the tooltip content container is not wrapped in IE. 

## Expected behavior
Max-width of the content should be provided in the styles.

## Minimal reproduction of the problem with instructions
https://stackblitz.com/edit/angular-atdm1d?file=app/app.component.ts

## Environment
Product:
* Kendo Ui for Angular

Browser:
- IE

joneff:
@JoomFX can you confirm that this is still an issue?

JoomFX:
@joneff yes, this is still an issue.

PR #1187, opened 2020-01-28
fix(tooltip): add max-width

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/816


packages/bootstrap/scss/_variables.scss
```diff
@@ -1526,6 +1526,7 @@ $notification-error-border: tint($error, 9) !default;
 // Tooltip
 $tooltip-padding-y: $padding-y !default;
 $tooltip-padding-x: $padding-x !default;
+$tooltip-max-width: 250px !default;
```
TeyaVes: For the Bootstrap theme, maybe we should keep the tooltip max-width of the Twitter Bootstrap, which is 200px

TeyaVes:
The Kendo tooltip (in jQuery, as I see in Angular as well) has a setting for width and height and inline styles are added to the k-tooltip-content. Meaning, that if the width of the tooltip is set to equal 500px, but its parent has a max-width of 250px, the tooltip will overflow and won't look okay. So I am not sure if this is the best way to fix the issue here. Wouldn't a specified width fix the issue in IE? I am not able to open the stackblitz from the targeted issue in IE to check.

@elena-gancheva , @PreslavKozovski , what do you guys think?

PreslavKozovski:
Adding max-width is not okay. This will break the jQuery Tooltip - https://docs.telerik.com/kendo-ui/api/javascript/ui/tooltip/configuration/width

JoomFX:
Because of the exact issue @TeyaVes mentioned above, we needed to modify the Angular Tooltip component so it adds `max-width` (apart from the `width`) inline style.
After discussing the situation it seems that it will be best if we look for another solution and not merge this PR.

PR #1188, opened 2020-01-28
fix(editor): use k-button styles instead of k-tool for table wizard popup

joneff:
related to https://github.com/telerik/kendo/issues/10589


Juveniel:
This change also requires to update the rendering of the jQuery Editor table wizard. Targeted with https://github.com/telerik/kendo/pull/10593

Issue #1191, opened 2020-01-29
Add window component visual tests


elena-gancheva:
There are missing scenarios of window component in the visual tests

PR #1198, opened 2020-01-30
Add window tests

inikolova:
closes https://github.com/telerik/kendo-themes/issues/1191


tests/visual/window-rtl.html
```diff
@@ -0,0 +1,96 @@
...
+            grid-template-columns: repeat(3, 1fr);
+            align-items: start;
+        }
+        /* test page overrides */
+         .k-window {
+            position: relative;
+        }
+    </style>
+</head>
+<body dir="rtl" class="k-rtl">
```
elena-gancheva: the `dir="rtl"` attribute should be removed, since jQuery are rendering only class `k-rtl` if RTL.

tests/visual/window-rtl.html
```diff
@@ -0,0 +1,96 @@
...
+
+    <div id="test-area" class="grid">
+        <div class="k-widget k-window k-window-wrapper">
```
elena-gancheva: There is missing `dir="rtl"` attribute in the tests. The `dir="rtl"` is rendered from Angular components on the host element to specify that the component will be styled corresponding to RTL mode.

elena-gancheva:
👍 Remove the inner screenshots from the visual tests with an interactive rebase

Issue #1055, opened 2019-11-15
The layout breaks when using a Grid and a Splitter within a .k-form element


svetq:
**Describe the bug**
The layout breaks when adding a Grid and a Splitter within a .k-form wrapping element.

**To reproduce**
Please check the following example:
https://stackblitz.com/edit/angular-skfppx-j9k7h6?file=app/app.component.ts

**Workaround**
add the following css:
     
> .k-form .k-widget.k-splitter {
            display: flex;
        }

https://stackblitz.com/edit/angular-skfppx-8miyuk?file=app/app.component.ts

**Affected package
* theme-default
* theme-bootstrap
* theme-material

**Affected suites
* Kendo UI for Angular

**Affected browsers
- Chrome: 78

danielkaradachki:
Seem to be the same as https://github.com/telerik/kendo-themes/issues/835

PR #1204, opened 2020-02-03
Fix grid in form layout

inikolova:
Fixes: [1055](https://github.com/telerik/kendo-themes/issues/1055)


packages/default/scss/input/_layout.scss
```diff
@@ -250,11 +250,20 @@
     }
 
     .k-form {
-        input:not(.k-checkbox):not(.k-radio),
-        .k-widget:not(.k-calendar):not(.k-switch) {
+        input:not(.k-checkbox):not(.k-radio) {
```
joneff: This will make **all descendant inputs** block, including native checkboxes and radio buttons, as well as inputs in our components (`.k-input` and hidden inputs). That, in terms, may toggle visibility for them as a side effect, depending on the actual setup.

If the idea is to have our components, namely `k-textbox` and `k-textarea` be block, then let's make them only. Or rather, make them flex.

PR #1216, opened 2020-02-05
Consolidate barcode and qrcode with dataviz styles

joneff:
Slightly reduces the imports count and build time


packages/default/scss/dataviz/_layout.scss
```diff
@@ -378,6 +387,19 @@ $chart-title-font-size: 1.143em !default;
...
+
+
+@include exports("dataviz/qrcode/layout") {
+
+    .k-qrcode {
+        display: inline-block;
+    }
+
+}
+
```
PreslavKozovski: Do we need so many empty rows?

Issue #1202, opened 2020-02-03
Extract DropDowns styles from common


elena-gancheva:
Currently some logic for DropDowns, such as RTL and positioning of the clear icon are implemented in common `_forms.scss` file. The styles should be extracted and refactored in the component's `layout` and `theme` logical files.

Additionally add some missing tests to cover all scenarios and states of the DropDowns.

Note: Check DateInputs implementation as well.

PR #1217, opened 2020-02-05
refactor: forms file

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/1202


packages/default/scss/autocomplete/_layout.scss
```diff
@@ -82,6 +82,13 @@
             }
         }
 
+        // Searchbar
+        .k-searchbar {
+            flex: 1;
```
joneff: flex should be behind the other flex properties

Issue #239, opened 2018-07-26
Command buttons overlap if the column does not have enough width.


PreslavKozovski:
## I'm submitting a...
* Bug report 

reported in 1178646

## Current behavior

Command buttons overlap if the column does not have enough width.

https://www.screencast.com/t/sJLhyNY6v
https://www.screencast.com/t/b6uVQ28B8p

## Expected behavior

https://www.screencast.com/t/RfLoICH5av

## Minimal reproduction of the problem with instructions

https://dojo.telerik.com/ePohucoL
https://dojo.telerik.com/UMAgeWOv

## Environment
Product:

* Kendo Ui for jQuery
Package versions:

```

```

Browser:

- Chrome (desktop) version XX

NikoPenev21:
Reported also in 1448153

PR #1226, opened 2020-02-06
Grid: prevent command buttons from overlapping

TeyaVes:
targets: #239


joneff:
@TeyaVes superb! We should have done this from the start.

PR #1236, opened 2020-02-10
 Rework functions and mixins with default value

### LINKED ISSUE #1222: Rework functions with default value
Functions with default value should be reworked to either not have a default value, or contain the default value in the same file. The same goes for functions that use global variables.

In the context of sass modules, recursive loading is not allowed, so in terms, the functions and mixins don't have access to global variables.

joneff:
Fixes  #1222 

Note: the changes in bootstrap typography, badges etc are expected, because I updated the variables to match the ones from bootstrap.


TeyaVes:
Just a question, which is not related to the modified code, but to the way we are calculating $button-calc-size. I noticed that in the three different themes, we are calculating it in a different way:

default
$button-calc-size: calc( #{$button-inner-calc-size} + **2px** ) !default;

bootstrap
$button-calc-size: calc( #{$button-inner-calc-size} + #{**$input-border-width** * 2} ) !default;

material
$button-calc-size: calc( #{$button-inner-calc-size} + #{**$button-border-width** * 2} ) !default;

My suggestion is in the default theme to change the 2px to  **$button-border-width** * 2 and in bootstrap replace the $input-border-width with  **$button-border-width**

In this way, we'll keep the same behavior as button-size() before.

Vanka, what do you think?

joneff:
I will use the full version: `button-line-height + (padding-y * 2) + (border-width * 2)`

joneff:
To reduce the size of the PR, I will refactor the mixins in a separate PR(s)

Issue #1231, opened 2020-02-07
Grid Pager icons are reversed in RTL mode


dtopalov:
**Describe the bug**
The prev/next/first/last icons in the Grid Pager are pointing in the opposite direction when in RTL mode.

**To reproduce**
1. Go to https://stackblitz.com/edit/angular-esfabd?file=app/app.component.ts
2. Both sets of icons next to the page number buttons are pointing in the wrong direction

**Expected behavior**
![image](https://user-images.githubusercontent.com/7533599/74013224-a1659d00-4994-11ea-972a-01a57aa30ed2.png)

Possible workaround:

https://stackblitz.com/edit/angular-esfabd-wddnrg?file=app/app.component.ts

**Affected package (please provide)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please provide)**
* Kendo UI for Angular

**Affected browsers (please provide)**
- Firefox: [for example, 67, lts]
- Chrome: [for example, 77]
- Edge version: [for example, 15]
- Safari: [for example, 12]
- IE version: [for example, 11]

**Build system information (please provide)**
- OS: [for example, Windows, Linux, Mac]
- Node version: [for example, 10.15.3]
- NPM version: [for example, 6.4.1]

**Additional context**
Add any other context about the bug.

PR #1240, opened 2020-02-10
fix(grid): reversed pager icons in RTL

inikolova:
Fixes https://github.com/telerik/kendo-themes/issues/1231


Xizario:
Since it works in React, I hope it is not broken after this pr.

silviyaboteva:
After testing the changes, I confirm that this fix doesn't break the React pager layout as we use the other icons.

Issue #1244, opened 2020-02-11
[ToolBar] border-right on popup menu button in RTL


PetarMetodiev:
**Describe the bug**
When in `RTL` mode in `<kendo-toolbar [overflow]="true">` a border on the right side of the button which opens the overflow popup appears.

**To reproduce**
https://stackblitz.com/edit/angular-bqrwnj

**Expected behavior**
The button should not have a border on the right(similar to when not in `RTL` mode)

**Affected package (please provide)**
* theme-default

**Affected suites (please provide)**
* Kendo UI for Angular

PR #1245, opened 2020-02-12
Fix toolbar button border

inikolova:
Fixes https://github.com/telerik/kendo-themes/issues/1244


packages/bootstrap/scss/toolbar/_layout.scss
```diff
@@ -1 +1,17 @@
 @import "~@progress/kendo-theme-default/scss/toolbar/_layout.scss";
+
+
+.k-toolbar {
```
JoomFX: Do we need this code? I mean should there be a border in the Bootstrap theme?
I couldn't find a design for it.

If there should be a border, then the code should be wrapped in a `@include exports("toolbar/layout/bootstrap") { }` block.

inikolova: I followed the design for the Editor toolbar

PR #1246, opened 2020-02-12
Extract pager variables

### LINKED ISSUE #1234: Pager cuts page number larger then 99 in Bootstrap and Material.
Pager cuts page number larger than 99 in Bootstrap and Material.
In the default theme, it cuts the page larger than 99.

image

As this is coming from the rendering, we should see if we need a new rendering for this.

A current workaround is to use CSS to make the input larger:

```
.k-pager-wrap .k-pager-input .k-textbox {
    width: 5em;
}
```

joneff:
* Extract pager variables for easier customization
* make responsive pager look like dropdown and dropdown items
* simplify styling
* fix #1234 -- Pager cuts page number larger then 99 in Bootstrap and Material


PreslavKozovski:
@silviyaboteva,

@ivelina-penkova said to me that she would add the "input" to the design in the near future.

joneff:
@silviyaboteva the font-size improvement will be handled at a later stage

Issue #1247, opened 2020-02-12
[Slider] Vertical slider labels are not aligned with the respective ticks


dtopalov:
**Describe the bug**
When the orientation of the Slider is vertical, the large ticks are not aligned with the labels due to the dynamically calculated height of the ticks (LI elements) and the fact that the labels are a child element that is vertically aligned to the top.

Related internal issue:

https://github.com/telerik/kendo-angular-inputs/issues/436

**To reproduce**
Create a vertically oriented slider, set min to 0, max to 5, largeStep to 1, and smallStep to 5

**Expected behavior**
The label should be aligned with the tick

**Screenshots**
![image](https://user-images.githubusercontent.com/7533599/74348237-0c5f0b80-4dbb-11ea-9d44-df4153678a46.png)

**Affected package (please provide)**
* theme-default
* theme-bootstrap
* theme-material
* theme-tasks

**Affected suites (please provide)**
* Kendo UI for Angular

**Affected browsers (please provide)**
- Firefox: [for example, 67, lts]
- Chrome: [for example, 77]
- Edge version: [for example, 15]
- Safari: [for example, 12]
- IE version: [for example, 11]

**Build system information (please provide)**
- OS: [for example, Windows, Linux, Mac]
- Node version: [for example, 10.15.3]
- NPM version: [for example, 6.4.1]

**Additional context**
Add any other context about the bug.

PR #1252, opened 2020-02-13
fix(slider): vertical labels alignment

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/1247


packages/default/scss/slider/_layout.scss
```diff
@@ -127,6 +127,11 @@
         .k-slider-bottomright .k-last { background-position: -152px 0; }
         .k-slider-bottomright .k-tick-large.k-last { background-position: -62px 0; }
 
+        .k-middle {
```
PreslavKozovski: Why `k-middle`? A belive `k-center` is better for a class that assigns `align-items: center;`

JoomFX: It is related to [this issue](https://github.com/telerik/kendo-angular-inputs/issues/436) where we were about to render new, `k-middle` class.

However, it turned out that we do not need a new class, we can use the existing `k-tick-large` instead (which I did).

PreslavKozovski: If we no longer need this class, I believe you should remove/modify the test.

Issue #1143, opened 2020-01-07
Upload button does not appear as button, unless hovered, in Material theme


Dimitar-Goshev:
**Describe the bug**
Since kendo-theme-material 3.4.0 Select files button does not appear as button, unless hovered.
The change is due to the rule that all inbuilt buttons should be flat buttons. But [by design](https://github.com/telerik/kendo-ux/blob/master/themes/material/PSD/upload.psd), the button in Material Upload is not flat.

**To reproduce**
1. Go to this [Dojo](https://dojo.telerik.com/@Dimitar-Goshev/oRUcivuL)

**Current behavior**
Select files does not look as a button as [here]https://github.com/telerik/kendo-themes/blob/305ab8355069ec7ae0f77c9ea8352c48662b9e20/tests/visual/output/material/upload.png
![kendo-themes_upload png at](https://user-images.githubusercontent.com/6614183/71884948-f6b65080-3141-11ea-956d-f30e2747c9fa.png)

**Expected behavior**
Select files resembles a button as [here](https://github.com/telerik/kendo-themes/blob/13fbdc8f77252978791a6a9d36ca13a5490e316e/tests/visual/output/material/upload.png)
![kendo-themes_upload png ](https://user-images.githubusercontent.com/6614183/71884652-6546de80-3141-11ea-9cd7-ad13824f0afb.png)

**Affected package (please provide)**
* theme-material

**Affected suites (please provide)**
* Kendo UI for jQuery

**Affected browsers (please provide)**
- Firefox: 71
- Chrome: 79
- Edge version: 44
- IE version: 11

silviyaboteva:
After a discussion with @ivelina-penkova, it seems that the designs for the material theme are not updated according the latest decision. The updated designs could be found [here](https://github.com/telerik/kendo-ux/blob/master/themes/material/PSD/upload.psd). The button should be a default (not primary).

PR #1281, opened 2020-02-18
fix(upload): update upload button in the Material theme

silviyaboteva:
fixes: https://github.com/telerik/kendo-themes/issues/1143


joneff:
buttons inside other components should look flat-ish, or more precisely without shadow. So the suggested fix does not really help. I am still thinking on a global solution that will handle those cases.

Leaving this one in WIP state for now

joneff:
As per discussion with designers, we shade the background of the dropzone just a bit

Issue #1221, opened 2020-02-06
Switch to `@include fill()`


joneff:
Switch to `@include fill()` instead of `@include appearance()`.

Needed for migrating to sass module system

- [x] action buttons (@JoomFX  #1295)
- [x] buttons (@JoomFX #1295)
- [x] dropdownlist (@JoomFX #1287)
- [x] multiselect (@JoomFX #1300)
- [x] gantt (@JoomFX #1304)
- [x] grid (@JoomFX #1311)
- [x] listbox (@JoomFX #1314)
- [x] pivotgrid (@JoomFX #1317)
- [x] popup (@JoomFX #1321)
- [x] scheduler (@JoomFX #1323)
- [x] splitter (@JoomFX #1326)
- [x] spreadsheet (@JoomFX #1330)
- [x] toolbar (@JoomFX #1336)
- [x] colorpicker (@JoomFX #1339)
- [x] base (@JoomFX #1343)
- [x] dataviz (@JoomFX #1344)
- [x] datetime (@JoomFX #1346)
- [x] editor (@JoomFX #1352)
- [x] map (@JoomFX #1354)
- [x] notification (@JoomFX #1357)

PR #1295, opened 2020-02-21
Use mixin fill for button and action-buttons

JoomFX:
Part of #1221


joneff:
Some screenshots are appearing with blank space. We'll target that later

Issue #785, opened 2019-06-14
Textboxes are moved to a new line in IE 11


svetq:
## I'm submitting a...
* Bug report 

## Current behavior
In IE11 each Kendo Textbox is moved to a new line by default.

## Minimal reproduction of the problem with instructions
Check the following example in IE 11:
https://stackblitz.com/edit/angular-m5egnj-9ezh6b?file=app/app.component.ts

## Workaround

Add the following style:
`    .k-textbox{
      display:inline-block;
    }`

https://stackblitz.com/edit/angular-m5egnj-c16ofe?file=app/app.component.ts

## Environment
Product:
* Kendo Ui for Angular

joneff:
Apparently, IE 11 has issues with input elements that are inline-flex. So this must be targeted specifically:

```scss
input.k-textbox {
    display: inline-block;
}
```

PR #1299, opened 2020-02-24
fix(input): update input display property

inikolova:
Fixes https://github.com/telerik/kendo-themes/issues/785

### DIFF
--- a/packages/default/scss/adaptive/_layout.scss
+++ b/packages/default/scss/adaptive/_layout.scss
@@ -161,12 +161,6 @@
                 white-space: nowrap;
             }
 
-            textarea.k-textbox {
-                width: 100%;
-                min-height: 4em;
-                resize: vertical;
-            }
-
             .k-edit-buttons {
                 border-width: 0;
             }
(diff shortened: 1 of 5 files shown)


PreslavKozovski:
Another possible solution is to target IE like this:
```
.k-ie11 {
    .k-textbox,
    .k-input.k-textbox,
    .k-textarea {
        ...
    }
}
```

What do you think about this?

inikolova:
@PreslavKozovski, this will not work with the Kendo UI for Angular components

PreslavKozovski:
@inikolova 
As far as I know, such classes can be added to our Angular components. For example - 

image

Having said that, during our next meeting, we can discuss what we are going to do with such cases. Do we want to scope IE styles, or are we going to implement styles that can work in all supported browsers?

joneff:
@PreslavKozovski my bad. I didn't explain the task properly. The behaviour in question was caused by IE not understanding input { display: inline-flex } properly. To target this, it needs to be inline-flex.

Since input.k-textbox and textarea.k-textarea cannot have child elements, we can safely have:

```scss
input.k-textbox,
textarea.k-textarea {
    display: inline-block
}
```

without affecting any browser :)

PR #1301, opened 2020-02-24
Allow adaptive rendering without height

PreslavKozovski:
part of: telerik/kendo/issues/10336


PreslavKozovski:
I will squash the commits and remove the visual previews after the review.

PreslavKozovski:
The "SM" Pager has different styles when placed in an adaptive Grid - https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/adaptive/_layout.scss#L478

Also, during my tests, I noticed that, at this stage , it is broken - https://github.com/telerik/kendo-themes/pull/1301/commits/dc98ca143d526e4c95e635a2f822e32ab397435c?short_path=bd5cdac#diff-bd5cdac7ca06a8430f7286a515a40037

I believe that if we had the above test in the past, we wouldn't have introduced this regression.

PR #1307, opened 2020-02-24
Improve textboxes

joneff:
I noticed that we have a myriad of searchbox implementations, so I am going to target those in a following PR.

Part of #1795


silviyaboteva (approved):
Great work @joneff! 
I noticed that you've fixed a little bug in the textarea that is reporoducible in the editor-table-wizard-accessibility test :)

PR #1311, opened 2020-02-25
fix(grid): use fill mixin

JoomFX:
part of https://github.com/telerik/kendo-themes/issues/1221


JoomFX:
In my opinion, the changes shown in the `grid-grouping.png` file (text becomes bold) are how it should be.
This same text is bold in both the Default and the Material themes.

Issue #1315, opened 2020-02-26
$calendar-today-color -> missing ! default


meriturva:
Is it possible to add here:

https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/calendar/_theme.scss#L5

default check on `$calendar-today-color`

PR #1316, opened 2020-02-26
refactor(calendar): add missing default flag

inikolova:
Related to https://github.com/telerik/kendo-themes/issues/1315


packages/default/scss/calendar/_theme.scss
```diff
@@ -2,7 +2,7 @@
 // color - mark the today date with a different text color
 // box - instead of changing the text color, show a box around the date
 $calendar-today-style: color !default;
-$calendar-today-color: $primary;
```
TeyaVes: I would also suggest moving these variables to the variables.scss file

PR #1320, opened 2020-02-27
feat(inputs): add focused box-shadow in bootstrap

TeyaVes:
fixes telerik/kendo-themes/issues/345


silviyaboteva:
Based on the visual tests I saw two things:
 - on tests/visual/output/bootstrap/dateinputs.png, the Angular focused widgets have two box-shadows applied (one over the whole widget and one only over the input without the button)
 - on tests/visual/output/bootstrap/inputs.png there is an additional red box-shadow around the invalid MasedTextBox widget

The rest seems good.

TeyaVes:
> Based on the visual tests I saw two things:
> 
> * on tests/visual/output/bootstrap/dateinputs.png, the Angular focused widgets have two box-shadows applied (one over the whole widget and one only over the input without the button)
> * on tests/visual/output/bootstrap/inputs.png there is an additional red box-shadow around the invalid MasedTextBox widget
> 
> The rest seems good.

1. In angular, k-state-focused is added both on the wrapping element and on the inner element, which is why this double box-shadow is happening. I'll sync with someone from their team whether one of the two classes should be removed or this should be handled in the themes

2. The masked textbox is in an invalid state and it's focused, this is why it has the red box-shadow

TeyaVes:
As discussed with @elena-gancheva , the inner .k-state-focused will be removed from the angular pickers, so I updated the tests.

I also added a new row in the inputs' test for invalid focused inputs to make it clearer

PR #1325, opened 2020-03-02
Badge styles

joneff:
* use only translate styles to position badge
* update badge styles to match design (default, material)
* conditionally set badge size to fixed or calculated

Needed for as a base for badge sizes and shapes.


inikolova:
Height of the badges in the Bootstrap theme should be 22px (as per the design)

joneff:
The height of bootstrap badges (in bootstrap) is `(75% * 1rem) + (2 * 3px) = 12px + 6px = 18px`. Ours is 20px because we add 1px for top and bottom border.

Also, the bootstrap badge is relative font-size i.e. if you put it in a container with bigger font size it will change size, where as default and material have fixed font-size.

PR #1334, opened 2020-03-05
docs: add third-party license notices

ipeshev:


ipeshev:
@joneff this should go into distrubuted packages, the text is for Kendo Angular and should be changed as well. Maybe it should be referenced in packages/*

ipeshev:
Closing since themes are opensource

PR #1342, opened 2020-03-08
Throw gulp errors

joneff:


packages/theme-tasks/gulpfile.js
```diff
@@ -81,7 +82,7 @@ function getArg(key) {
     return (index < 0) ? null : (!next || next[0] === "-") ? true : next; // eslint-disable-line no-nested-ternary
 }
 function _info(msg, ...args) { return logger.info(colors.gray(msg), ...args); }
-// function _error(msg, ...args) { return logger.info(colors.red(msg), ...args); }
+function _error(msg, ...args) { return logger.error(colors.gray(msg), ...args); }
```
silviyaboteva: I only have one question. Why the color for the error is grey(like the info) instead of red?

joneff: Because if everything is red it becomes ineligible.

PR #1362, opened 2020-03-13
Message box

joneff:
Add styles for message box


packages/default/scss/all.scss
```diff
@@ -22,6 +22,7 @@
 @import "ripple/_index.scss";
 // @import "table/_index.scss";
 // @import "typography/_index.scss";
+@import "messagebox/_index.scss";
```
PreslavKozovski: Add the styles in Bootstrap and Material.

packages/default/scss/messagebox/_layout.scss
```diff
@@ -0,0 +1,14 @@
+@include exports("message-box/layout") {
+
+    // Base
+    .k-message-box {
+        margin: $message-box-margin;
```
TeyaVes: There are some visual changes in the message box in the chat- there is additional margin at the bottom. I suppose this should be fixed

Issue #1359, opened 2020-03-12
Do not rely on .k-widget for styling


joneff:
The initial styles of `.k-widget` tend to do more damage than good and we need to remove them. To do so, we need to have a set of variables for each component that uses them e.g.:

```scss
$grid-border-width: 1px !default;
$grid-font-size: $font-size !default;
$grid-line-height: $line-height !default;

$grid-bg: $component-bg !default;
$grid-text: $component-text !default;
$grid-border: $component-border !default;
```

Note: unless otherwise needed, the each component should inherit from base component variables.

- [ ] TBA

After this is done, we can go on with #1072.

- [x] grid (@JoomFX #1361)
- [x] autocomplete (@JoomFX #1365)
- [x] adaptive (@JoomFX #1366)
- [x] breadcrumb (@JoomFX #1367)
- [ ] buttons (@JoomFX #1370)
- [x] calendar (@JoomFX #1373)
- [x] card (@JoomFX #1377)
- [x] chat (@JoomFX #1378)
- [x] colorpicker (@JoomFX #1381)
- [x] combobox (@JoomFX #1383)
- [x] dataviz (@JoomFX #1387)
- [x] datetime (@JoomFX #1389)
- [x] window (@JoomFX #1390)
- [x] drawer (@JoomFX #1392)
- [x] dropdownlist (@JoomFX #1393)
- [x] dropdowntree (@inikolova - extends dropdownlist)
- [x] editor (@inikolova  https://github.com/telerik/kendo-themes/pull/1402)
- [x] gantt (@inikolova https://github.com/telerik/kendo-themes/pull/1409)
- [x] listbox (@JoomFX #1453)
- [x] listview (@TeyaVes https://github.com/telerik/kendo-themes/pull/1400)
- [x] maskedtextbox (@JoomFX #1470)
- [x] mediaplayer (@JoomFX #1471)
- [x] menu (@JoomFX #1472)
- [x] multiselect (@JoomFX #1473)
- [x] notification (@JoomFX #1475)
- [x] numerictextbox (@JoomFX #1476)
- [x] pager (@JoomFX #1478)
- [x] panelbar (@JoomFX #1479)
- [x] pdf-viewer (@JoomFX #1485)
- [x] progressbar (@JoomFX #1496)
- [x] rating (@JoomFX #1497)
- [x] scheduler (@JoomFX #1498)
- [x] scrollview (@JoomFX #1510)
- [x] slider (@JoomFX #1511)
[...]

PR #1367, opened 2020-03-16
refactor(breadcrumb): extract k-widget styles

JoomFX:

### DIFF
--- a/packages/bootstrap/scss/_bootstrap-overrides.scss
+++ b/packages/bootstrap/scss/_bootstrap-overrides.scss
@@ -49,6 +49,10 @@ $theme-colors: () !default;
     $component-active-bg: $primary !default;
     $component-active-color: contrast-yiq($component-active-bg) !default;
 
+    $breadcrumb-padding-x: 0 !default;
+    $breadcrumb-padding-y: 0 !default;
+    $breadcrumb-bg: $component-bg !default;
+
     $card-bg: $component-bg !default;
     $card-color: $component-color !default;
     $card-border-color: rgba(contrast-yiq($card-bg, $gray-900), .125) !default;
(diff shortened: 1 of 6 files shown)


packages/bootstrap/scss/_bootstrap-overrides.scss
```diff
@@ -49,6 +49,8 @@ $theme-colors: () !default;
     $component-active-bg: $primary !default;
     $component-active-color: contrast-yiq($component-active-bg) !default;
 
+    $breadcrumb-bg: $component-bg !default;
```
joneff: That's not needed.

```suggestion
```

packages/default/scss/_variables.scss
```diff
@@ -1794,6 +1794,14 @@ $tooltip-error-border: $tooltip-error-bg !default;
 
 
 // Breadcrumb
+$breadcrumb-font-family: $font-family !default;
+$breadcrumb-font-size: $font-size !default;
+$breadcrumb-line-height: $line-height !default;
+
+$breadcrumb-bg: $component-bg !default;
+$breadcrumb-text: $component-text !default;
+$breadcrumb-border: $component-border !default;
```
joneff: If there is border color, we should also have border-width

packages/material/scss/_variables.scss
```diff
@@ -1838,6 +1838,15 @@ $tooltip-error-border: $tooltip-error-bg !default;
 
 
 // Breadcrumb
+$breadcrumb-border-width: 0 !default;
+$breadcrumb-font-family: $font-family !default;
+$breadcrumb-font-size: $font-size !default;
+$breadcrumb-line-height: $line-height !default;
```
joneff: it's good that we have those, but we are not using them ;)

PR #1374, opened 2020-03-17
Remove redundant validator styles

joneff:
The sass counterpart of telerik/kendo#10675:

* remove any `k-tooltip-validation` styles
* switch them with `k-tooltip-error` for now
* update tests


TeyaVes:
The tooltip width in the visual tests doesn't seem correct

PR #1383, opened 2020-03-19
refactor(combobox): extract k-widget styles

JoomFX:


JoomFX:
The `k-widget` theme styles are not applied to the `k-combobox` class on purpose.
I believe there is no need to apply them since `k-combobox` has `border-width: 0;` and `background: none;` and since the theme styles are already applied to the `k-dropdown-wrap` class.

PR #1402, opened 2020-03-25
refactor(editor): extract k-widget styles

inikolova:
Part of https://github.com/telerik/kendo-themes/issues/1359

### DIFF
--- a/packages/default/scss/editor/_layout.scss
+++ b/packages/default/scss/editor/_layout.scss
@@ -3,11 +3,20 @@
     // Base
     .k-editor {
         @include use-background-clip();
+        border-width: $editor-border-width;
+        border-style: solid;
+        box-sizing: border-box;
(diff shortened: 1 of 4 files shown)


packages/default/scss/editor/_layout.scss
```diff
@@ -3,6 +3,13 @@
     // Base
     .k-editor {
         @include use-background-clip();
+        border-width: $editor-border-width;
+        border-style: solid;
+        box-sizing: border-box;
+        outline: 0;
+        font-family: inherit;
```
JoomFX: It doesn't really matter, but for consistency sake we might want to have the `font-family` as a separate variable, as it is in [the other components](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/grid/_layout.scss#L43).

Issue #1406, opened 2020-03-26
Chip label overflows its content on smaller resolutions


elena-gancheva:
**Describe the bug**
On small resolutions, the chip label is overflowing its content.

**Expected behavior**
The chip label should not overflow its content, may be should set `text-overflow: ellipsis;`

**Screenshots**
![image](https://user-images.githubusercontent.com/8842674/77627369-7557a880-6f4f-11ea-9fa9-e69b919b7eec.png)

Check on the demos: https://www.telerik.com/kendo-angular-ui/components/buttons/

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for Angular

**Affected browsers (please remove the unneeded items)**
* All

PR #1410, opened 2020-03-26
fix(chip): label overflows its content on small resolutions

inikolova:
Fixes https://github.com/telerik/kendo-themes/issues/1406


packages/default/scss/chip/_layout.scss
```diff
@@ -90,7 +91,15 @@
...
     .k-chip-content {
+        min-width: 0;
+        overflow: hidden;
+        text-overflow: ellipsis;
         padding: 0 $chip-content-padding-x;
+        flex: 1;
+    }
+
+    .k-chip-label {
+        white-space: nowrap;
```
JoomFX: I think it will be better if we move the `white-space: nowrap` to `.k-chip-content` because we do not have `.k-chip-label` element when the user adds the label with content projection.
An example of what I mean can be seen in the [Custom Content](https://www.telerik.com/kendo-angular-ui/components/buttons/chip/content/#toc-custom-content) demo.

inikolova: Good catch! Thank you :)

PR #1370, opened 2020-03-17
refactor(buttons): extract k-widget styles

JoomFX:
part of https://github.com/telerik/kendo-themes/issues/1359

PR #1414, opened 2020-03-27
Refactor button

joneff:
* simplify disabled button styles (no visual change)
* use `k-button-outline` class for styling outline button (no visual change)
* unify appearance of split button between kendo-jquery and kendo-angular

Note: after this PR in IE 11 there will be a double border between buttons in button group and split button due to using `margin-inline-start` instead of more convoluted practices.

FYI @tsvetomir @Dimitar-Goshev @mparvanov @dmanova 

Needed for #1310 and #1370


tsvetomir:
The borders don't look too bad in IE 11, they're just a bit blurry:
image

joneff:
@tsvetomir the blurry borders are due to:

* the buttons are not int pixels wide, but float pixels wide
* the borders are transparent
* probably `background-clip` plays a role too

I am not sure we could improve on (1), but (2) and (3) are open improvement

PR #1415, opened 2020-03-28
Reduce filemanager tests

joneff:
* fix: reduce spacing in filemanager preview
* reduce filemanager tests

The improvement in reducing space will need some code tweaking in kendo-jquery


Juveniel:
@joneff 

If reducing the spacing of the preview pane is intended, I think that we should also update [the designs](https://github.com/telerik/kendo-ux/blob/master/themes/default/PNG/file%20manager.png) to reflect this change ?

joneff:
As per discussion with @ivelina-penkova, I am going to reduce the padding from 96 to 48px but in addition, I am going to set min-height to the preview container.

Issue #1306, opened 2020-02-24
Improve doc generation


joneff:
- [x] Switch template engine from handlebars to nunjucks. You can take inspiration from https://github.com/SassDoc/sassdoc-theme-default. Take special note how that theme can have group descriptions and the various helpers.
- add the following features:
  - [x] **Will do in v.next** ~~Group description~~
  - [x] Add row for examples, if any
  - [x] Deprecation
  - [x] resolved variables
  - [x] variable type
- [x] Arrange everything in ./docs folder in theme-tasks with ./docs/theme containing the themefiles
- [x] Change task name from `api` to `docs` and `api:check` to `docs:check`
- we can add more features

PR #1435, opened 2020-04-02
Switch docs template engine from handlebars to nunjucks

PreslavKozovski:
closes https://github.com/telerik/kendo-themes/issues/1306


packages/theme-tasks/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@progress/kendo-theme-tasks",
-  "version": "0.5.2",
+  "version": "0.5.3",
```
PreslavKozovski, line 3: @joneff 
When branched, the version was "0.5.0", due to the new packages, the version automatically went to "0.5.1". During the rebase, I had a conflict as it is "0.5.2" in the "develop" branch. I manually update the version to "0.5.3". Is this okay?

joneff: yes, it's OK

joneff: Yes, it's OK, but we'll have to rebase everything one more time

Issue #1432, opened 2020-04-02
Progressbar is broken when a custom width/height is set.


silviyaboteva:
**Describe the bug**
When customizing the progress bar thickness, only the empty part is resized.

**To reproduce**

1. Go to https://stackblitz.com/edit/angular-wfyzay?file=app%2Fapp.component.ts where it is demonstrated.

**Expected behavior**
Resizing the whole progress bar.

**Screenshots**
None

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**
* All

**Build system information (please remove the unneeded items)**
* Not Applicable

**Additional context**
None

PR #1436, opened 2020-04-02
fix(progressbar): progress is broken when a custom width/height is set

silviyaboteva:
fixes: https://github.com/telerik/kendo-themes/issues/1432

In the design, the status label in the Material theme should have offset of 5px: https://github.com/telerik/kendo-ux/blob/master/themes/material/PSD/progressbar-jquery.psd


silviyaboteva:
Hey @TeyaVes I pushed the broken tests firstly, and then made a commit with the actual fix. However, the Travis didn't produce any commit :( I will research why.

TeyaVes:
The labels in Material look a bit higher then they were before, you can add line-height: 1 (so that the line-height equals the font-size) and this should fix it.

silviyaboteva:
I looked at the designs and there is a 5px offset, so I left it that way. :)

PR #1446, opened 2020-04-04
Prepare inputs

joneff:
Needs #1445 to be merged and then rebased

* various fixes, mostly in material theme, to resemble design more closely
* split `k-textbox` and `k-textarea` styles
* extract input placeholder variables

Needed for #1307 

There are still some things to be done about textarea in material theme, but that's going to be solved as well.

Fixes telerik/blazor#1034 (as per https://github.com/telerik/kendo-themes/pull/1446#issuecomment-612850294)


TeyaVes:
Also fixes: https://github.com/telerik/blazor/issues/1034

PR #1475, opened 2020-04-14
refactor(notification): extract k-widget styles

JoomFX:


JoomFX:
In my opinion, the generated visual previews show the correct styling (as it should be).

PR #1483, opened 2020-04-16
test(forms): update the react radio group rendering in the test

silviyaboteva:


silviyaboteva:
Is it necessary to be in this branch as we currently waiting for @joneff to think of a more appropriate solution?

Issue #1501, opened 2020-04-22
Invalid Checkbox/RadioButton Label should be red


JoomFX:
**Describe the bug**
As per [the new design](https://xd.adobe.com/view/a816c5e0-0c51-4fba-69b5-67b2dc43da5c-de27/screen/660ed1d4-c334-4c45-8722-4b325cc546e9/Check-Box-and-Radio-Buttons), the label, when the Checkbox/RadioButton is in invalid state, should be red as well.

**To reproduce**
N/A

**Expected behavior**
To have red labels in invalid state.

**Screenshots**
![invalid-label](https://user-images.githubusercontent.com/7753940/79989277-56015c00-84b8-11ea-943a-bf27f91c2c64.png)

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**
* All

**Build system information (please remove the unneeded items)**
* Not Applicable

PR #1503, opened 2020-04-23
fix(checkbox): invalid state label color

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/1501


JoomFX:
With the current rendering it is not possible to style the invalid state label when the input is wrapped in a label:

```
<label class="k-checkbox-label">
    <input type="checkbox" class="k-checkbox k-state-invalid" required>
    Required
</label>
```

We might need to render an additional `k-state-invalid` class on the `k-checkbox-label` (and `k-radio-label`) element.

PR #1512, opened 2020-04-27
refactor(splitter): extract k-widget styles

JoomFX:


JoomFX:
The `$splitter-bg` and `$splitter-text` variables were already taken, which is the reason for the unconventional `$splitter-base-*` naming.

PR #1517, opened 2020-04-28
refactor(timeline): extract k-widget styles

JoomFX:


joneff:
The timeline, importunately, uses `k-timeline`, which is used in gantt, but... I guess we can work that out

Issue #1440, opened 2020-04-03
Wrong icons directions in RTL mode


inikolova:
**Describe the bug**
In RTL mode slider icons have wrong directions. The problem can be observed in this example:

https://stackblitz.com/edit/angular-jqxx8s?file=app/app.component.ts

PR #1547, opened 2020-05-12
fix(slider): wrong icons directions in RTL mode

silviyaboteva:
fixes: https://github.com/telerik/kendo-themes/issues/1440


tests/visual/slider.html
```diff
@@ -605,6 +605,54 @@
                 </div>
             </div>
         </section>
+
+        <span class="k-rtl">RTL Horizontal (Angular)</span>
+        <span></span>
+        <span></span>
+
+        <section class="k-rtl">
+            <kendo-slider class="k-slider-horizontal k-slider k-widget k-state-default k-slider-transitions" style="width: 276px;">
```
JoomFX, line 614: Could you please add `dir="rtl"` to the `<kendo-slider> element and remove the `k-rtl` class from the `<section>`?
That's how it works in Kendo Angular - the `dir` attribute is rendered on the component host element.

PR #1563, opened 2020-05-14
fix(label): angular label adjustments

JoomFX:
related to https://github.com/telerik/kendo-angular-inputs/pull/454


packages/default/scss/checkbox/_layout.scss
```diff
@@ -121,9 +121,19 @@
     .k-checkbox-label + .k-checkbox {
         margin-left: $checkbox-label-margin-x;
     }
+    .k-checkbox + kendo-label.k-checkbox-label,
+    .k-checkbox-label + kendo-label.k-checkbox {
+        margin-left: calc( #{$checkbox-label-margin-x} - 4px );
+    }
```
joneff, line 127: We don't know how those are placed and where. Remove the styles.

```suggestion
```

packages/default/scss/checkbox/_layout.scss
```diff
@@ -121,9 +121,19 @@
...
     .k-checkbox-label > .k-checkbox {
         margin-right: $checkbox-label-margin-x;
     }
+    kendo-label.k-checkbox-label > .k-checkbox {
+        margin-right: 0;
+    }
+    kendo-label.k-checkbox-label > .k-label {
+        margin-right: $checkbox-label-margin-x;
+    }
```
joneff, line 136: You're assuming the checkbox is after the label, which might not be the case. Remove the styles or rework it to handle both cases.

```suggestion
```

packages/default/scss/checkbox/_layout.scss
```diff
@@ -159,10 +169,22 @@
             margin-left: 0;
             margin-right: $checkbox-label-margin-x;
         }
+        .k-checkbox + kendo-label.k-checkbox-label,
+        .k-checkbox-label + kendo-label.k-checkbox {
+            margin-left: 0;
+            margin-right: calc( #{$checkbox-label-margin-x} - 4px );
+        }
```
joneff, line 176: Remove.
```suggestion
```

packages/default/scss/checkbox/_layout.scss
```diff
@@ -159,10 +169,22 @@
...
             margin-right: 0;
             margin-left: $checkbox-label-margin-x;
         }
+        kendo-label.k-checkbox-label > .k-checkbox {
+            margin-left: 0;
+        }
+        kendo-label.k-checkbox-label > .k-label {
+            margin-right: 0;
+            margin-left: $checkbox-label-margin-x;
+        }
```
joneff, line 187: That related to the second comment. Remove or rework to handle the cases of checkbox being first or last.
```suggestion
```

packages/default/scss/radio/_layout.scss
```diff
@@ -91,9 +95,19 @@
     .k-radio-label + .k-radio {
         margin-left: $radio-label-margin-x;
     }
+    .k-radio + kendo-label.k-radio-label,
```
joneff, line 98: see comments about checkbox

tests/visual/label-angular.html
```diff
@@ -0,0 +1,89 @@
...
+        <span>Checkbox</span>
+        <span>Radio</span>
+        <span dir="rtl">Checkbox (RTL)</span>
+        <span dir="rtl">Radio (RTL)</span>
+
+        <section>
+            <input type="checkbox" class="k-checkbox" checked>
+            <kendo-label class="k-checkbox-label">
+                <label class="k-label">Checked</label>
+            </kendo-label>
```
joneff, line 29: in this case, you have two `inline-bock` elements that will have a space between them. You are trying to compensate it by subtracting what you think might be the width of a space character. However, if those are placed inside a flex element, or otherwise there is no space between them, they will overlap.

We cannot handle this particular case, so we don't handle it.

tests/visual/label-angular.html
```diff
@@ -0,0 +1,89 @@
...
+        <section>
+            <input type="radio" class="k-radio" checked>
+            <kendo-label class="k-radio-label">
+                <label class="k-label">Checked</label>
+            </kendo-label>
+        </section>
+
+        <section dir="rtl">
+            <input type="checkbox" class="k-checkbox" checked>
+            <kendo-label class="k-checkbox-label" dir="rtl">
```
joneff, line 41: you don't need `dir="rtl"` on `kendo-label`. The parent has it.

tests/visual/label-angular.html
```diff
@@ -0,0 +1,89 @@
...
+            </kendo-label>
+        </section>
+
+        <span>Input inside Label</span>
+        <span>Input inside Label</span>
+        <span dir="rtl">Input inside Label (RTL)</span>
+        <span dir="rtl">Input inside Label (RTL)</span>
+
+        <section>
+            <kendo-label class="k-checkbox-label">
```
joneff, line 59: you don't have a test for
```
kendo-label
   input
   label
```

which is a valid scenario.

tests/visual/label-angular.html
JoomFX: I will keep the `dir="rtl"` on `kendo-label` so we can mimic the exact component rendering.

PR #1570, opened 2020-05-25
fix(checkbox): checkbox variables are messing default flag

IKKI2000:


joneff:
@IKKI2000 can you change the commit message to match the guidelines e.g. "fix(checkbox): checkbox variables are messing default flag" so your commit will be picked up by the conventional change log script?

Thanks for the commit. We really missed that one.

joneff:
I meant the commit message ;) Look at https://github.com/telerik/kendo-themes/commits/develop. See how most commits have either `chore`, `fix`, `feat` or some similar prefix in the message? We use to filter those to extract changelog information.

Of course, I can merge your commit without that. There is no technical problem with that. It's just the item won't make it in the change log

joneff:
superseded by #1576

PR #1584, opened 2020-06-02
fix(tooltip): add max-width to container

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/816


JoomFX:
We use `k-animation-container` in many places so I'm not 100% sure if this approach is ok.
It should be fine, but other opinions and reviews are needed.

PR #1586, opened 2020-06-02
chore: add funcitonality to resolve variables values

joneff:
* to be used for the documentation
* since sass-variables-parser is unmaintained, there is some trickery applied to bootstrap files

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -3,12 +3,16 @@ const path = require("path");
 const glob = require("glob");
 const cp = require("child_process");
 
+const fse = require("fs-extra");
 const gulp = require("gulp");
 const logger = require("gulplog");
 const eslint = require("gulp-eslint");
 const sasslint = require("gulp-sass-lint");
 const mime = require("mime");
 const sassdoc = require("sassdoc");
+
+const baka = require("@joneff/baka");
+const { parse } = require('sass-variable-parser');
 
 
 // Settings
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -114,3 +117,42 @@ gulp.task("docs:check", function() {
     });
 });
 // #endregion
+
+
+gulp.task("resolve-vars", function() {
+    let themes = glob.sync(paths.sass.themes);
+    const { parse } = require('sass-variable-parser');
```
PreslavKozovski, line 124: I believe we should hoist this var.

gulpfile.js
```diff
@@ -114,3 +117,42 @@ gulp.task("docs:check", function() {
...
+    const cwd = process.cwd();
+
+    themes.forEach( theme => {
+
```
PreslavKozovski, line 128: What JS guidelines do we use for this repo? In Kendo for jQuery, we do not add an empty line at the beginning and the end of a function, is this okay for this repo?

gulpfile.js
```diff
@@ -114,3 +117,42 @@ gulp.task("docs:check", function() {
...
+        let content = fs.readFileSync( variablesScss, 'utf-8' );
+
+        content = content.replace(/ url\("data.*?\)/g, 'none');
+        content = content.replace(/\/\/.*/gm, '');
+        content = content.replace(/\n/gm, '');
+        content = content.replace(/;/gm, ';\n');
+
+        const variables = parse( content, { camelCase: false } );
+
+        fse.ensureFileSync( variablesJson );
```
PreslavKozovski, line 152: I believe this does nothing, and we can remove it.

gulpfile.js
joneff: we'll have to think on that.

PR #1594, opened 2020-06-04
remove base styles for k-widget

joneff:
We can't remove all `k-widget` styles in one pass, but that's a good starting point.


inikolova:
I observed the following differences: 
[ ] According to the design, there should be border for the Drawer in the Bootstrap theme (drawer-rtl.png):
https://github.com/telerik/kendo-ux/blob/master/themes/bootstrap/PNG/drawer.png

PR #1614, opened 2020-06-11
chore: update lerna to v3

Juveniel:
Update lerna to v3 along with related dependencies.


joneff:
Converting to draft. Will merge next week **after** kendo-jquery 2020 R2 SP1 is out.

Issue #1118, opened 2019-12-11
Treelist footer row styles


danielkaradachki:
**Describe the bug**
No styles to separate the footer rows with the data rows

**To reproduce**
1. Go to https://demos.telerik.com/kendo-ui/treelist/aggregates
2. The root item and the footer have the same color and there is no border.
3. The same is true for the children, if the last row before the footer does not have k-alt class.

**Expected behavior**
Visually separate the data and the footers.

**Affected package (please provide)**
* theme-default
* theme-bootstrap

**Affected suites (please provide)**
* Kendo UI for jQuery
* Kendo UI for Angular

PR #1640, opened 2020-06-19
Treelist footer row styling

inikolova:
Closes https://github.com/telerik/kendo-themes/issues/1118


packages/material/scss/treelist/_layout.scss
```diff
@@ -1 +1,17 @@
 @import "~@progress/kendo-theme-default/scss/treelist/_layout.scss";
+
+@include exports("treelist/layout/material") {
+    .k-treelist {
+        tr.k-footer-template td {
+            border-width: 1px 0;
```
PreslavKozovski, line 6: This style is causing a double border - 

image

packages/material/scss/treelist/_layout.scss
```diff
@@ -1 +1,17 @@
...
+        }
+
+        tr.k-footer td {
```
PreslavKozovski, line 9: I cannot see a case where we use this selector. Could you please show me an example in the TreeList? 

If we are using it, as the CSS is the same, we can add the selector to the next selector.

packages/material/scss/treelist/_layout.scss
```diff
@@ -1 +1,17 @@
...
+            border-width: 0 0 1px 0;
+        }
+
+        tr.k-footer-template + tr.k-footer-template td {
+            border-width: 0 0 1px 0;
```
PreslavKozovski, line 14: This could be a variable

packages/default/scss/treelist/_layout.scss
```diff
@@ -35,6 +35,17 @@
...
         }
+
+        tr.k-footer td,
+        tr.k-footer-template td {
+            border-width: 1px 0;
+            border-style: solid;
+            font-weight: bold;
+        }
+
+        tr.k-footer + tr.k-footer td {
```
PreslavKozovski, line 46: ```tr.k-footer-template + tr.k-footer-template td ```

Should resolve the double border problem - 

image

packages/material/scss/treelist/_layout.scss
inikolova: This selector is used in the Kendo UI TreeList for Angular, however I will add it to the next one :)

packages/default/scss/treelist/_layout.scss
inikolova: Adding this rule will remove the top border from every footer row in jQuery TreeList:

Screenshot 2020-06-30 at 13 03 23

In jQuery TreeList all aggregates are initially rendered, however their visibility is toggled via **display: none;**

packages/material/scss/treelist/_layout.scss
inikolova: According to the PSD file, there are borders: 
https://github.com/telerik/kendo-ux/blob/master/themes/default/PSD/treelist.psd

silviyaboteva:
Only in the React tree list test, please add k-group-footer class to the inner group footer rows as like this: 
Screenshot 2020-07-22 at 11 41 20

Issue #798, opened 2019-06-25
Create theme based on CSS variables


joneff:
Our themes, as of now, are based on explicit values and that comes with it's short comings. CSS variables might be able to solve some of those, especially nesting selectors and customization.

Consider the following for button and toolbar button:

```scss
:root {
    // use k- prefix not to conflict; or maybe not
    --k-button-bg: black;
    --k-button-text: white;
    --k-toolbar-button-bg: gold;
    --k-toolbar-button-text: black;
}
.k-button {
    color: var(--k-button-text);
    background-color: var(--k-button-bg);
}

.k-toolbar {
    --k-button-bg: var(--k-toolbar-button-bg);
    --k-button-text: var(--k-toolbar-button-text);
}
```

That will enable us not to override the complete selector and in terms increase complexity. It may also be possible to solve #450 (Allow multiple themes for the same component on the same page).

That approach is also not without it's shortcoming. Currently, no post- or pre- processor can grasp the syntax and make polyfils for IE. Let's not forget that some of the logic is not immediately obvious too and debugging is a bit more convoluted.

MaklaCof:
In our company we are very interested in this feature. It would allow us to change Theme at run-time, based on logged in user.

Choosing custom theme is a must nowadays. Every serious web application supports it. And a lot of desktop applications support it to. At least light and dark.

I had another idea, which may be easier/quicker to implement:
I am building Angular application and I am using Kendo UI for Angular and SASS. It would be awesome if SASS variable could accept CSS variable (for example):
- current (Material): `$primary: #73B3A2`
- propose solution is: `$primary: var(--my-app-primary-color)`

I think this would work, but we need theme without processing colors like this:
```
//File: node_modules\@progress\kendo-theme-material\modules\@progress\kendo-theme-default\scss\mixins\_colors.scss
@function true-mix($color1, $color2, $weight: 50%) {
    @return mix(rgba($color1, 1), rgba($color2, 1), $weight);
}
```

What do you think @joneff ? Any timeline?

OlehFylypiv:
I've got the same error. :(
I'm really waiting for this feature.

Is it possible to implement, @joneff?

OlehFylypiv:
Colleagues, maybe you can help with the answer about the timeline?
@veselints
@vladimirivanoviliev
@Dimitar-Goshev
@elena-gancheva

joneff:
I've prepared a small PR (#1677) and while parts of the discussions should be there, here are a couple of key points:

[...]

PR #1677, opened 2020-07-02
fix(button): add initial styles for button

joneff:
Related to #798

* sass variables allow us to have `null` value and do not generate output when there is no need, which is not the case with css variables;
* on top of that, css variables, if not provided with fallback and have `null` value will fallback to `prop: initial`
  * focus state, per se, is `null`, so should we provide a fallback to normal state?
* should we have a sort of handling of css variables, so we use either css or sass variables?
  * `prop: k-var( --kendo-css-var, $kendo-sass-var )` and have a `$use-css-variables` switch...

Apart from that, it looks like it would be doable to have e theme that's based on css variables but can also benefit from sass toolchain.

### DIFF
--- a/packages/nouvelle/.sassdocrc
+++ b/packages/nouvelle/.sassdocrc
@@ -1,0 +1,6 @@
+{
+    "meta": {
+        "name": "nouvelle",
+        "friendlyName": "Nouvelle"
+    }
+}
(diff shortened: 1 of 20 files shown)


joneff:
Merging to have a patch version of the theme, so we have something deployed to test against.

PR #1700, opened 2020-07-17
feat: TextArea styles

silviyaboteva:
Related to: https://github.com/telerik/kendo-themes/issues/1672.

Remaining things:
 - resize mark - custom or default? Color like the border.
Update: we will keep browsers default resize mark and there should be no coloring as the resize mark doesn't support it.
 - material floating label position and font size
Update: the position will remain the same, only the font size will become smaller. A text area with floating label should have an additional padding-top to provide the needed space for the label.


dmanova:
It seems some additional changes are needed in the styles to make the following functionalities work properly:

- [ ] No vertical scrollbar appears when the text is more rows than the visible textarea height. Ant it should be styled scrollbar.

- [ ] Cols should be configurable but at the moment its value is always set to 20. And actually it is not considered as some other style with set up width has overwritten the width of the textarea. Even in the width of textarea with floating label is different than the textararea without label.

- [ ] It is not possible to resize the text area horizontally, only vertically. Could be related to the above issue.


The styles implemented are according to the visual previews in this PR, but they are quite different than the designs here - https://github.com/telerik/kendo-ux-private/issues/143. I'm mentioning some of the differences to make sure they are not changed by accident:

- [ ] Resize mark is colored like the border of the textarea by design.

- [ ] Material theme has underline (bottom border is wider) only when floating label is set up. By design, it has it in all cases, although the design here is not consistent with the inputs (they also have the underline only with floating labels).

- [ ] Material theme textarea has light gray background according to design.

- [ ] Material theme floating label is glued to the textarea border. Not very clear if it is by design.

silviyaboteva:
@dmanova 
Only I need to update you about the cols attribute, as I have spoken to the team in the chat about this. The Textarea has default width and min-height. The col/row attributes are not working as expected as they depend on the font size. So, in our components, the col attribute will not do anything.

silviyaboteva:
@joneff currently, we have a width set to the textarea that doesn't allow the handle to resize it horizontally. What should we do in this case to allow resizing in both directions?

inikolova:
I observed there is a missing border in the demo with floating label in RTL
Screenshot 2020-07-31 at 8 14 33

silviyaboteva:
Hi @inikolova,
Thanks for the review. I only have one concern. It's about the first point with the missing border. I'm not sure, but I see the same thing in both normal and RTL textarea - there is a border for the scrollbar only if you click on it and it is not rendered to the end, but to the place where the resize mark starts. Attaching screenshots. Please elaborate more on this.
ltr
rtl

inikolova:
@silviyaboteva, thanks for addressing my notes -> the textarea looks really nice :) good job :)

dmanova:
- [ ] Material SASS Theme - Floating label in RTL doesn't look properly aligned.
image


- [ ] Material SASS Theme - I can't click on the floating label to start writing. In the other themes that is the first row and you could click.

IE - no focus styles are applied. This seems to be valid for a few more components like textBox, maskedTextBox. But we have it for numericTextBox. So maybe this should be a separate issue.

silviyaboteva:
Hey @dmanova,

 - I fixed the RTL issue in this PR: https://github.com/telerik/kendo-themes/pull/1771
 - As for the second one, Presli is going to investigate as we think that this is not a problem with the styles. 
 - About the IE issue, yes, can you please open a new issue for the focus styles in IE browser?

Issue #1673, opened 2020-06-30
Restyle sticky and locked columns


joneff:
TBA

inikolova:
According to the latest design, there shouldn't be darkening overlay and should have darker borders:

https://github.com/telerik/kendo-ux/blob/master/themes/default/PNG/grid.png
https://github.com/telerik/kendo-ux/blob/master/themes/bootstrap/PNG/grid.png
https://github.com/telerik/kendo-ux/blob/master/themes/material/PNG/grid.png

PR #1727, opened 2020-07-28
fix: update grid locked/sticky columns styling

inikolova:
Fixes: #1673


inikolova:
@silviyaboteva , thanks for the help :)

PR #1737, opened 2020-07-29
feat: provide additional theme color classes

inikolova:


inikolova:
@PreslavKozovski, removing the _important_ rule is an approach we discussed offline with @joneff . However, I agree that adding a second set of colors without _!important_ is a good alternative. My suggestion is to use **k-color-primary**, **k-color-info** etc. classes. Please share your thoughts :)
@joneff @silviyaboteva

silviyaboteva:
I'm ok to have k-color-primary, etc. without the !important and use it in the SvgIcon/FontIcon components. And preserve the current k-text-primary, etc. with the !important.

joneff:
We should have a convention how and when we use important... It does make sense to have the important cases, however, we can't really two distinct versions of the syntax, say `color-primary` and `text-primary`. It gets quite tricky to know which is with important and which is not.

Perhaps, we could have a suffix or prefix or something that shows that the class will force the behaviour (unless a stronger `!important` selector)

packages/default/scss/common/_decoration.scss
```diff
@@ -1,12 +1,16 @@
 @include exports( "common/decoration/color" ) {
 
-    // sass-lint:disable-block no-important
-
-    .k-text-primary     { color: $primary !important; }
-    .k-text-info        { color: $info !important; }
-    .k-text-success     { color: $success !important; }
-    .k-text-warning     { color: $warning !important; }
-    .k-text-error       { color: $error !important; }
+    .k-text-inherit     { color: inherit; }
```
joneff, line 3: apart from inherit, the rest could be generated from the theme map ;)

inikolova:
@joneff , could you please elaborate a bit more on the idea for classes with suffix and prefix? Thank you in advance :)

inikolova:
In offline discussion with @joneff , we agreed to:

- Keep the `!important` rule, which is currently applied to **k-text-primary** etc., however we should consider these styles as legacy;
- Add new classes **k-color-primary** etc. without `!important`;

As of now, new utility classes should not have `!important`.

PR #1771, opened 2020-08-11
fix(textarea): remove additional padding-top from the jQuery textarea

silviyaboteva:
Based on the comments here: https://github.com/telerik/kendo-themes/pull/1700#issuecomment-670608688


packages/material/scss/textarea/_layout.scss
```diff
@@ -25,7 +25,7 @@
 
 
         .k-textarea .k-input {
-            padding-top: calc( #{$textarea-padding-y} + 14px );
+            padding-top: calc( #{$textarea-padding-y} + 20px );
```
joneff, line 28: is this always 20px, or is it 1 line? if the latter, consider using `$line-height-em` or a similar unit of measurement.

silviyaboteva: I've added `$input-line-height-em` and `$textarea-line-height-em` variables and fixed the calculation :)

PR #1809, opened 2020-09-02
Gantt toolbar

joneff:
@silviyaboteva I am not sure how responsive toolbar is handled in react, but if we need to, we can additionally extend the styles.

### DIFF
--- a/packages/default/scss/gantt/_layout.scss
+++ b/packages/default/scss/gantt/_layout.scss
@@ -159,24 +159,7 @@
 
 
     // Toolbar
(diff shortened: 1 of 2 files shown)


veselints (approved):

silviyaboteva:
While we implementing the Gantt component in the React suite, the toolbar gantt views issue were addressed. Check the commit in my open PR: https://github.com/telerik/kendo-themes/pull/1788/files#diff-ff26ca3237d938bbce7382fd1f0960cbR162

PR #1844, opened 2020-09-09
infra: add stable release action

### LINKED ISSUE #1770: npm tag for kendo-jquery
~~Provide a `kendo-jquery` tag for use in kendo-jquery official  / sp releases.~~

Releases compatible with kendo-jquery will be tagged with `stable` tag.

joneff:
Action that adds **stable** dist-tag to themes themes. Used to denote version compatible with kendo-jquery.

Fixes #1770

### DIFF
--- a/build/tag-stable.sh
+++ b/build/tag-stable.sh
@@ -2,9 +2,12 @@
 
 set -e
 
-for theme in default bootstrap classic material nouvelle
+source $(pwd)/build/constants.sh
+
+for theme in $themes
 do
-    pkg=@progress/kendo-theme-$theme;
-    version=$(npm view $pkg version);
+    pkg=@progress/kendo-theme-$theme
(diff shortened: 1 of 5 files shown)


build/tag-stable.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+for theme in default bootstrap classic material nouvelle
```
tsvetomir, line 5: The list of themes could be moved to a separate file and `source`'d when needed. It is used in at least 3 locations in `/build` and can fall out of sync.

joneff: Indeed. We just need to leave a comment in the file that **default** should always be first, so we can omit it when needed.

joneff: done

tsvetomir (approved):

Issue #787, opened 2019-06-14
Inconsistent borders with bootstrap themes


jlchereau:
The use of `<hr>` and bootstrap `.border-*` classes is not consistent with widget borders. This is especially relevant with dark themes produced by the SASS theme builder.

The following html:

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Title</title>
    <link href="highconytrast.css" rel="stylesheet">
</head>
<body class="p-4">
<h1 class="border-bottom">A title with border-bottom</h1>
<div class="k-widget border-top mt-4 mp-1">A .k-widget with border-top</div>
<hr>
<div class="k-widget k-window k-dialog">
    <div class="k-window-titlebar k-dialog-titlebar k-header">
        <span class="k-window-title k-dialog-title">A .k-window-title.k-dialog-title</span>
    </div>
    <div class="k-content k-window-content k-dialog-content">
        A .k-window-content.k-dialog-content
    </div>
</div>
</body>
</html>
```

And the following SCSS:

```scss
$base-theme:Bootstrap;
$skin-name:highcontrast;
$swatch-name:High Contrast;
$font-size: 1rem;
$enable-border-radius: false;
$enable-gradients: false;
$border-radius: 0;
$accent: #fff;
$secondary: #600840;
$info: #73e3ff;
$success: #40f540;
$warning: #ffcd42;
$error: #ff4d4d;
$body-bg: #000;
$body-color: #fff;
$component-bg: #000;
$component-color: #fff;
$card-cap-bg: #600840;
$card-cap-color: #fff;
$series-a: #a7008f;
$series-b: #ffb800;
$series-c: #3aafff;
$series-d: #99c900;
$series-e: #b20753;
$series-f: #ff4195;

@import "~/@progress/kendo-theme-bootstrap/scss/all";
@import "~/bootstrap/scss/bootstrap";
```

produce:

![image](https://user-images.githubusercontent.com/2556751/59513044-e5f67280-8eb1-11e9-8792-5256ff3f046c.png)

In fact `<hr>` and `.border-*` classes are assigned the default bootstrap color of $gray-300: #dee2e6. There is no kendo sass theming applied.

![image](https://user-images.githubusercontent.com/2556751/59513525-ea6f5b00-8eb2-11e9-9527-f134f21fe677.png)

> In addition `.k-*` border colors are hardly visible with dark themes.

## Code

[...]

PR #1845, opened 2020-09-09
Update card colors to match bootstrap 4.5.1

joneff:
Part of #1830. Targets #787


silviyaboteva (commented):
Why are there so many modified tests?

joneff:
The computed value of `$header-border` differs ever so there you have it :)

silviyaboteva (approved):

PR #1874, opened 2020-09-15
fix(icon): use inline-flex for icon layout

joneff:
prerequisite for #1873


inikolova:
- Icons in scheduler events are not vertically centered (Default and Material);
-  Color Picker -> k-selected-color span not centered (Material theme): 
Screenshot 2020-09-16 at 10 44 42
Is this intentional?

joneff:
Scheduler event icons: with the current structure of events and actions, icons will rarely look aligned.

colorpicker -- it is centered: 
image

So yes, both are intentional.

PreslavKozovski (approved):

Issue #1872, opened 2020-09-15
Extract component variables to their modiles


joneff:
Extract component variables to their modules i.e. for the component `autocomplete` the folder structure should be:
```
autocomplete/_variables.scss
autocomplete/_index.scss // <= this will be renamed to index.scss at some point
autocomplete/_layout.scss
autocomplete/_theme.scss
```

Inside `index.scss`

```scss
@import "../_variables.scss"; // <= this will be removed at some point
@import "../mixins/_index.scss"; // <= this will be removed at some point


// Dependencies
@import "../common/_index.scss"; // <= this will be removed at some point
@import "../input/_variables.scss"; // <= or if the component depends entirely, index.scss
@import "../popup/_index.scss"; // <= like so
@import "../icons/_index.scss"; // <= like so


// Component
@import "_variables.scss";
@import "_layout.scss";
@import "_theme.scss";
```

* [x] actions: #1935 (@joneff)
* [x] adaptive: #1935 (@joneff)
* [x] appbar: #1883 (@joneff)
* [x] autocomplete: #1873 (@joneff)
* [x] avatar: #1879 (@joneff)
* [x] badge: #1882 (@joneff)
* [x] breadcrumb: #1884 (@joneff)
* [x] button: #1936 (@joneff)
* [x] calendar: #1886 (@joneff)
* [x] card: #1935 (@joneff)
* [x] chat:  #1888 (@joneff)
* [x] checkbox: #1911 (@joneff)
* [x] chip: #1889 (@joneff)
* [x] colorpicker:  #1891 (@joneff)
* [x] combobox:  #1890 (@joneff)
* [x] dataviz: #1892 (@joneff)
* [x] datetime: #1886 (@joneff)
* [x] dialog: #1893 (@joneff)
* [x] drawer: #1894 (@joneff)
* [x] dropdowngrid: N/A
* [x] dropdownlist: #1919 (@joneff)
* [x] dropdowntree: #1895 (@joneff)
* [x] dropzone: #1897 (@joneff)
* [x] editor: #1898 (@joneff)
* [x] filemanager: #1899 (@joneff)
* [x] filter: #1900 (@joneff)
* [x] forms: #1936 (@joneff)
* [x] gantt: #1901 (@joneff)
* [x] grid: #1934 (@joneff)
* [x] icons: N/A
* [x] imageeditor: #1903 (@joneff)
* [x] input: #1936 (@joneff)
* [x] listbox: #1904 (@joneff)
[...]

PR #1882, opened 2020-09-17
Extract badge variables into separate files

joneff:
Part of #1872


PreslavKozovski (approved):

packages/bootstrap/scss/badge/_index.scss
```diff
@@ -8,5 +8,6 @@
 
 
 // Component
+@import "_variable.scss";
```
Juveniel, line 11: should be _variable**s**

Juveniel (changes requested):
The bootstrap package file is names "_variable.scss" and is included in the _index.scss as "_variable.scss". Please, change this to be consistent with other packages as "_variables".

packages/bootstrap/scss/badge/_index.scss
joneff: done

PR #1914, opened 2020-09-23
fix: use calc method to calculate radio line height

feitzi:
Use calc to calculate a css property to allow mix `px` and 'em' sizes! Like PR #1060


joneff (approved):

joneff:
Nice catch!

PR #1920, opened 2020-09-25
Update form tests

joneff:
* match updated rendering
* cleanup form tests
* cleanup test results


joneff:
@silviyaboteva  there is a small bug in the floating label: when using `textarea.k-textarea` the floating label is not positioned correctly in material theme. but that's a small thing.

silviyaboteva:
@joneff I see that previously, it was positioned as expected. Can we make a separate PR and fix this?

PR #1930, opened 2020-09-29
fix(checkbox): use calc to calculate checkbox dash width

feitzi:
same as #1914 and #1916 
Maybe it is a good idea to always check if new calculations are done by calc()?


joneff (approved):

joneff:
@feitzi thanks

Issue #1657, opened 2020-06-25
Validation message in Material theme in grid popup wrong position


TeyaVes:
**Describe the bug**
Validation message in Material-v2 theme in grid popup is not positioned correctly and the value cannot be filled:

![image](https://user-images.githubusercontent.com/52446546/82459458-a7edcf80-9ac0-11ea-9cb5-8ecbc08533a1.png)

In IE the position is wrong as well:

![image](https://user-images.githubusercontent.com/52446546/82459676-e4b9c680-9ac0-11ea-945c-92a05deb90f3.png)

**To reproduce**

https://demos.telerik.com/aspnet-mvc/grid/editing-popup

**Expected behavior**
The validation message should not cover the inputs

**Affected package (please remove the unneeded items)**
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React

**Affected browsers (please remove the unneeded items)**
* All

joneff:
Fixed by #1941

PR #1941, opened 2020-10-02
validation popup position in Material theme

PreslavKozovski:
targets: https://github.com/telerik/kendo-themes/issues/1657


joneff (approved):
Just to be clear, as soon as we transition all forms to our form guidelines, this will not be needed.

Issue #1854, opened 2020-09-11
Restyle sticky and locked rows


silviyaboteva:
Designs: 
 - https://github.com/telerik/kendo-ux/blob/master/themes/default/PNG/grid.png
 - https://github.com/telerik/kendo-ux/blob/master/themes/bootstrap/PNG/grid.png
 - https://github.com/telerik/kendo-ux/blob/master/themes/material/PNG/grid.png

PR #1960, opened 2020-10-08
Add sticky rows styling and restyle grid footer

silviyaboteva:
Closes https://github.com/telerik/kendo-themes/issues/1854

Remaining questions:
 - [x] Do the sticky rows need to apply the same background for all rows. For example in Default and Bootstrap themes, there are alt and non-alt rows, and if 2alt rows are stick one after another, the styling looks strange (at least for me)
Screenshot 2020-10-08 at 15 14 35

- [x] How to render a grid with grouping and a sticky row? 

/cc @ivelina-penkova @joneff @telerik/kendo-front-end


joneff:
To me:

1) Yes, we should render the alt class and all the relevant styles.
2) Any sticky row is the row it is -- be it grouping, alt, -- what ever. The idea of a sticky row is to show needed and relevant information to the user, per the discretion of the developer. Should a developer decide to stick a a grouping row, then that's it. True, we should advise against that scenario, but don't forbid it.

joneff (changes requested):
There is discrepancy in one of the screenshots
image

joneff (approved):

PR #1977, opened 2020-10-14
Detect jenkins for clean and install jobs

joneff:
Ugly hack. Requires environmental `$JENKINS` variable to be set to `true`, even though there is no true `true` in bash :D

Alternatively, I could use `$CI` variable for all such cases, in case we need the same handling for all cases.

/cc @ipeshev @tsvetomir ideas?

### DIFF
--- a/build/clean.sh
+++ b/build/clean.sh
@@ -1,8 +1,11 @@
 #!/bin/bash
 
-if [[ $TRAVIS == true ]]
+if [[ $TRAVIS == 'true' ]]
 then
     echo "Travis CI detected. Skipping clean."
+elif [[ $JENKINS == 'true' ]] || [[ -n $JENKINS_URL ]]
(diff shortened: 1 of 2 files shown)


tsvetomir:
Why not use the `JENKINS_URL` variable as in [this plug-in](https://github.com/krux/condition-jenkins/blob/master/index.js)?

We only need to test if it's present:
```sh
  if [[ -n $JENKINS_URL` ]]
```

joneff:
My only concern is the following paragraph:

> JENKINS_URL
> Full URL of Jenkins, such as http://example.com:port/jenkins/ (NOTE: only available if Jenkins URL is set in "System Configuration".)

So I am adding this as a secondary check, just in case we stumble in a situation where `JENKINS_URL` is not present.

Issue #1986, opened 2020-10-16
The floating label container should use the font-size of the input


PreslavKozovski:
**Describe the bug**

The floating label container should use the font-size of the input. At this stage, it is inheriting the font-size. 

**To reproduce**

https://dojo.telerik.com/IZuWoPOl

**Expected behavior**

https://dojo.telerik.com/iBOViLEg

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**
* All

**Build system information (please remove the unneeded items)**
* Not Applicable

PR #2000, opened 2020-10-21
Add font-size to floating-label-container

GeorgeGindev:
Targets: https://github.com/telerik/kendo-themes/issues/1986


PreslavKozovski (approved):

joneff (changes requested):
The fix will not solve the issue: what if there are styles targeting `.k-floating-label-container`? The outcome will still be the same.

Instead, look inside the floating label variables, more specifically `$floating-label-height`. It's currently based on `em` units, but you could very well base it on exact pixels, by passing the correct value when multiplying.

GeorgeGindev:
So if `.k-floating-label-container` is added `font-size`, is the correct behavior to apply the font-size to all inner elements like so https://dojo.telerik.com/iBOViLEg/8 , or should it have no effect on the inner elements with pixel values, like so https://dojo.telerik.com/iBOViLEg/9 ?

joneff:
@GeorgeGindev the floating label container is just that: a container. it's not like it can't have font size, it just doesn't, because it doesn't need to: apart from the floating label, the only other content item is the input and it has it's own set of variables.

Try to look at things this way: if I set floating label font size to be 20px and input font size to be 25, how do I expect them to apply inside a floating label container?

To me that means: the label is 20px and the input is 25.

Or to answer your question directly: https://dojo.telerik.com/iBOViLEg/9, to me, seems like the correct way to go.

joneff (approved):

Issue #1950, opened 2020-10-06
Add styles for expansion panel component


joneff:
Designs:
[Default theme](https://xd.adobe.com/view/829efe58-3e5b-408f-9dc6-039f9392eda7-d7cd/)
[Bootstrap theme](https://xd.adobe.com/view/4ba238a8-525c-4f54-a15e-d07b2a7fc8b1-1ef1/)
[Material theme](https://xd.adobe.com/view/e66685ff-b597-4337-955a-ca533b24cf9c-b335/)

@inikolova will implement Sass themes
@PreslavKozovski will port to less

inikolova:
In the following dojo can be found my proposal for rendering of the component: 

https://dojo.telerik.com/@inikolova/ASAjaZIW

Because of the component name is long enough, I suggest using **k-expander** and **k-expander-....** classes instead.

@joneff @PreslavKozovski @silviyaboteva please share your thoughts.

PreslavKozovski:
As I am not a fan of adding a lot of HTML where it is not needed. I would simplify the "header" rendering to:

```
      <div class="k-expander-header">
        <div class="k-expander-title">brazil</div>
        <span class="k-expander-spacer"></span>
        <div class="k-expander-sub-title">South America</div>
        <span class="k-expander-indicator k-icon k-i-arrow-chevron-down"></span>
      </div>

      .k-expander-spacer {
        flex: 1 1 auto;  
      }

```

In the above code, I removed the "k-expander-inner" and "k-expander-indicator" wrappers.

However, the rendering, at its current state, LGTM, so the above suggestion is just for consideration.

silviyaboteva:
I'm dropping my suggestions as well:
 - I agree with Presli's proposal for removing the inner element and use the spacer instead. We use this approach in the AppBar and Scheduler toolbar as well.
 - I was expecting to see a root element with a class by it's name, e.g. `<div class="k-expander">...</div>`. And inside, we can list the expander collection of items. 

Here is a demo with the proposed changes: https://dojo.telerik.com/oZOKUhEG/2.

joneff:
Just have in mind that flex does not yield the results you always expect:

```
// imagine the following consecutive expanders

gosho          sub v
pesho    longersub v

// in material guidelines it should be
gosho    sub       v
pesho    longersub v
```
All being said, I am in favour of flex.
I would probably

inikolova:
Thank you for the feedback. Please find my comment: 

[...]

PR #2003, opened 2020-10-22
Expansion panel styles

inikolova:
Part of: https://github.com/telerik/kendo-themes/issues/1950


silviyaboteva (commented):
Great work @inikolova! I have only 2 concerns:
 - the focus state in the Default and Bootstrap themes designs is 1px border, but you have used 2px and 3px inset box-shadow. What is the correct one?
 - why only the first expansion panel doesn't have top padding? It looks strange to me.

inikolova:
@silviyaboteva Silve, below are my comments to the points:

- I discussed with @nedevp in the chat and we used the same focus state like in Panelbar/Menu etc. I will ping him to update the designs;
- This is by design; the main reason is that the expansion panel can be used as a single panel.

silviyaboteva (approved):
💯

joneff (approved):
I have one minor comment regarding the background of the tests, but that can be targeted at a later stage

packages/default/scss/expansion-panel/_layout.scss
```diff
@@ -0,0 +1,59 @@
...
+            padding: $expander-content-padding-y $expander-content-padding-x;
+            overflow: auto;
+        }
+
+        // Multiple expansion panels immediately one after another
+        +.k-expander {
+            .k-expander-header {
+                padding-top: $expander-padding-y;
+            }
+        }
```
PreslavKozovski, line 56:
```suggestion
        + .k-expander .k-expander-header {
            padding-top: $expander-padding-y;
        }
```

packages/material/scss/expansion-panel/_theme.scss
```diff
@@ -0,0 +1,12 @@
...
+
+$expander-header-focused-bg: rgba( black, .12 ) !default;
+
+.k-expander {
+    .k-expander-header {
+        &.k-state-focused {
+            background-color: $expander-header-focused-bg;
+        }
+    }
+}
```
PreslavKozovski, line 12:
```suggestion
.k-expander .k-expander-header .k-state-focused {
    background-color: $expander-header-focused-bg;
}
```

packages/material/scss/expansion-panel/_theme.scss
```diff
@@ -0,0 +1,12 @@
+@import "~@progress/kendo-theme-default/scss/expansion-panel/_theme.scss";
```
PreslavKozovski, line 1: Wrap the Material theme in a module

packages/material/scss/expansion-panel/_theme.scss
```diff
@@ -0,0 +1,12 @@
...
+
+
+$expander-header-focused-bg: rgba( black, .12 ) !default;
```
PreslavKozovski, line 4: Move this to the `_variables.scss` file

tests/visual/expansion-panel.html
```diff
@@ -0,0 +1,93 @@
+<!doctype html>
+<html lang="en" class="k-typography k-no-animations">
+<head>
+    <title>Panelbar</title>
```
PreslavKozovski, line 4:
```suggestion
    <title>Expansion Panel</title>
```

tests/visual/expansion-panel.html
```diff
@@ -0,0 +1,93 @@
...
+    <meta charset="utf-8" />
+    <link rel="stylesheet" data-role="kendo-theme" href="../../packages/default/dist/all.css" />
+    <link rel="stylesheet" href="assets/styles.css" />
+    <script src="assets/scripts.js"></script>
+
+    <style>
+        #test-area {
+            /*grid-template-columns: repeat(3, 1fr);*/
+            align-items: start;
```
PreslavKozovski, line 13:
```suggestion
```

PreslavKozovski (changes requested):

packages/material/scss/expansion-panel/_theme.scss
inikolova: Header focused background should be applied only in the Material theme, therefore I added the variable here. Do you think we should add this variable in the Default and Bootstrap themes too?

packages/default/scss/expansion-panel/_layout.scss
inikolova: We usually use nesting for the selectors. What's the reason for this change

packages/material/scss/expansion-panel/_theme.scss
inikolova: Same question as above :)

packages/default/scss/expansion-panel/_layout.scss
PreslavKozovski: Indeed, we are using nested selectors for easier implementation. However, nesting a selector slowers the build and it is harder to read/find in code when only a single declaration is used.

We can add this as a question to the agenda of our tomorrow's meeting, and discuss this topic with other FEs. After that, we can add our decision to the following wiki - https://github.com/telerik/kendo-themes/wiki/Code-Styleguide

packages/material/scss/expansion-panel/_theme.scss
PreslavKozovski: I am the person who started the initiative to make all of our theme `variables` similar, however, in some situations, I believe this makes no sense.

I do believe that this is one of them. For example, if we make all the `_variables` files identical, and we import the "Default" layout/theme to the other themes - we no longer have a different theme but a swatch.

With the above said, I believe that it is okay to have a different number of variables for the different themes when it makes sense.

We can also add this as a topic for our meeting.

silviyaboteva: Can we have the same variable in the Default and Bootstrap themes, but with `null` value?

inikolova: Yes, we could, but do we need them? We will discuss this case in our meeting

PreslavKozovski (approved):

joneff (approved):

Issue #1948, opened 2020-10-06
Add styles for floating action button


joneff:
Designs:
[Default](https://xd.adobe.com/view/71c896dd-b1e4-435e-9bd2-89f55a2d68b0-249b/)
[Material](https://xd.adobe.com/view/481aea43-7f90-4774-be3f-c761f38fac48-1d49/)
[Bootstrap](https://xd.adobe.com/view/8fc898ac-6581-4b27-81c6-f1840999ecd9-da9b/)

@silviyaboteva will implement sass themes
@Juveniel will port to less

silviyaboteva:
I prepared a demo with the rendering of the component: https://dojo.telerik.com/OPevoLoY/7

- In the demo, I listed all the classes that can be applied.
- I used the rendering of the Icon and Badge components as it is noted in the designs

@telerik/kendo-front-end please share your thoughts/suggestions for improvement on this.

PreslavKozovski:
The rendering looks good. I would suggest renaming `k-fab-button` to `k-speed-dial-button`

inikolova:
I agree with Presli -> rename the class to **k-speed-dial-button**. Rendering looks good to me, too :)

silviyaboteva:
After offline discussion with @elena-gancheva and @Juveniel, and based on the feedback of Presli and Iliyanka, I updated the rendering. Here you can find the latest version: https://dojo.telerik.com/eREjORiV

After the quick brainstorming session, there are a few questions that need to be discussed:
-  We need to add `role` and `aria` attributes to the component
- We need a root wrapping element that will hold both - the button and the speed dia. **Q:** What will be the class name? k-fab` or`k-floatingactionbutton` or `k-floating-action-button` or `k-floating-action-btn` or `k-speed-dial`?
- We need to decide what will be the main class name of the button itself. Is the `k-floatingbutton` or `k-floating-button` or `k-floating-btn` or something else?
- Based on https://github.com/telerik/kendo-themes/issues/1983#issuecomment-709926692, we are thinking of adding a wrapper element over the icons. But what will be the class name? `k-btn-icon` and respectively `k-btn-text`?
- Based on the designs, there are different positions of the labels of the speed dial items, how can we handle this? From the themes and introducing new classes for the positioning or via the component to swap the elements (icon and label)?
- How to handle animations - via CSS or JS animations? What if the users want custom animation duration?

[...]

PR #2007, opened 2020-10-23
Fab styles

silviyaboteva:
Part of: https://github.com/telerik/kendo-themes/issues/1948

### DIFF
--- a/packages/default/scss/fab/_layout.scss
+++ b/packages/default/scss/fab/_layout.scss
@@ -1,0 +1,145 @@
+@include exports( "fab/layout" ) {
+
+    // Floating Action Button (FAB)
+    .k-fab {
+        padding: 0;
+        border-width: 0;
+        border-style: solid;
+        border-color: transparent;
+        box-sizing: border-box;
+        color: inherit;
+        background-color: transparent;
+        font-size: $fab-font-size;
+        line-height: $fab-line-height;
+        font-family: $fab-font-family;
+        text-align: center;
+        white-space: nowrap;
+        display: inline-flex;
+        align-items: center;
+        justify-content: center;
+        vertical-align: middle;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        cursor: pointer;
+        outline: none;
+    }
+
+    // FAB Icon
+    .k-fab-icon {
+        width: $fab-icon-width;
+        height: $fab-icon-height;
+
+        &:not(:only-child) {
+            margin-right: $fab-icon-spacing;
+            margin-left: -$fab-icon-spacing;
+        }
+    }
+
+
+    // Sizes
+    .k-fab-sm {
+        padding: $fab-padding-y-sm $fab-padding-x-sm;
+    }
+    .k-fab-md {
+        padding: $fab-padding-y $fab-padding-x;
+    }
+    .k-fab-lg {
+        padding: $fab-padding-y-lg $fab-padding-x-lg;
+    }
+
+
+    // Shapes
+    .k-fab-circle {
+        border-radius: 50%;
+    }
+    .k-fab-pill {
+        border-radius: 5rem;
+    }
+    .k-fab-rounded {
+        @include border-radius( $fab-border-radius );
+    }
+
+
+    // Items
+    .k-fab-items {
+        margin: 0;
+        padding: $fab-items-padding-y $fab-items-padding-x;
+        display: flex;
+    }
+
+    .k-fab-items-bottom {
+        flex-direction: column;
+    }
+    .k-fab-items-bottom .k-fab-item:last-child {
+        margin-bottom: 0;
+    }
+
+    .k-fab-items-top {
+        flex-direction: column-reverse;
+    }
+    .k-fab-items-top .k-fab-item:first-child {
+        margin-bottom: 0;
+    }
+
+    .k-fab-item {
+        list-style-type: none;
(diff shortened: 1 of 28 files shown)


Juveniel (approved):

packages/default/scss/fab/_layout.scss
```diff
@@ -0,0 +1,145 @@
...
+        @include border-radius( $fab-border-radius );
+    }
+    .k-fab-rectangle { }
+    .k-fab-square { }
+
+
+    // Items
+    .k-fab-items {
+        margin: 0;
+        padding: 12px 0;
```
elena-gancheva, line 113: fab-item(s) spacings could be extracted to variables as well?

elena-gancheva (approved):

packages/default/scss/fab/_theme.scss
```diff
@@ -0,0 +1,54 @@
...
+            background-color: adjust-hue( $color, 3 );
+            background-image: linear-gradient(rgba( adjust-hue( adjust-hue( $color, 3 ), 14), 0 ), rgba( adjust-hue( adjust-hue( $color, 3 ), 14), .2 ));
+        }
+    }
+
+    // Focus Theme colors
+    @each $name, $color in $fab-theme-colors {
+        .k-state-focus.k-fab-#{$name},
+        .k-fab-#{$name}:focus {
+            box-shadow: 0 3px 4px 0 rgba($color, .4);
```
inikolova, line 32: box-shadow can be extracted as a variable

packages/default/scss/fab/_theme.scss
```diff
@@ -0,0 +1,54 @@
...
+    // Active Theme colors
+    @each $name, $color in $fab-theme-colors {
+        .k-state-active.k-fab-#{$name},
+        .k-state-selected.k-fab-#{$name},
+        .k-fab-#{$name}:active {
+            border-color: $color;
+            color: contrast-wcag( $color );
+            background-color: $color;
+            background-image: linear-gradient(rgba( adjust-hue( $color, 14), .2 ), rgba( adjust-hue( $color, 14), 0 ));
+            box-shadow: inset 0 2px 2px 0 rgba(0, 0, 0, .06);
```
inikolova, line 45: box-shadow can be extracted as a variable

packages/default/scss/fab/_layout.scss
```diff
@@ -0,0 +1,145 @@
...
+
+
+    // FAB Icon
+    .k-fab-icon {
+        width: $fab-icon-width;
+        height: $fab-icon-height;
+
+        &:not(:only-child) {
+            margin-right: $icon-spacing;
+            margin-left: -$icon-spacing;
```
inikolova, line 79: Why we need this negative left margin? According to the design, there is only padding for the fab

packages/default/scss/fab/_layout.scss
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+
+    // FAB Icon
+    .k-fab-icon {
+        width: $fab-icon-width;
+        height: $fab-icon-height;
+
+        &:not(:only-child) {
+            margin-right: $icon-spacing;
```
inikolova, line 78: Shouldn't this margin be $icon-spacing * 2?

inikolova:
Not sure if this is intended, but I observed that:
    - Hover/Focus/Active states in the Material theme differs from the design;
    - Focus/Active shadows in the Bootstrap theme differs from the design;

packages/default/scss/fab/_theme.scss
silviyaboteva: I couldn't extract these as variables (I know the buttons has their values extracted) but here we have a map over the themecolors. What will be the `$color` in this case if it's in _variables.scss?

packages/default/scss/fab/_layout.scss
silviyaboteva: Its because if I just set margin-right: 8px then there will be no equal spacing righthand and lefthand of the button. Take a look at this screenshot for reference: 
Screenshot 2020-10-27 at 11 37 08
I took the approach implemented in buttons.

However, you are right about the Bootstrap theme. I'll fix it.

packages/default/scss/fab/_layout.scss
silviyaboteva: Look at the previous comment :)

packages/default/scss/fab/_layout.scss
```diff
@@ -77,13 +77,31 @@
...
+            }
+        }
     }
 
     .k-fab-item-text {
-        padding: 4px 4px;
+        padding: map-get( $spacing, 1 ) map-get( $spacing, 1 );
         border-width: 1px;
         border-radius: 2px;
         font-size: 10px;
```
inikolova, line 107: We could use $font-size-xs instead of hardcoded value

inikolova, line 107: We could use $font-size-xs instead of hardcoded value

inikolova (approved):

packages/default/scss/fab/_layout.scss
silviyaboteva: I extracted all variables for both text and icon in order for clients to customize them as they want.

joneff (approved):
Don't forget to include all needed files, like `utils/position` ;)

PR #2017, opened 2020-10-30
fix(spread): remove lines in disabled merged cells telerik/kendo-ui-core#6107

veselints:


packages/default/scss/spreadsheet/_layout.scss
```diff
@@ -460,6 +460,14 @@
         background-clip: padding-box;
         white-space: pre;
         overflow: hidden;
+
+        &.k-state-disabled {
+            opacity: 1;
+        }
+
+        &.k-state-disabled > div {
+            opacity: .6;
```
Juveniel, line 469: Use theme-specific disabled styling as follows:

```suggestion
         @include disabled( $disabled-styling );
```

Juveniel (changes requested):

packages/default/scss/spreadsheet/_layout.scss
veselints: In this specific scenario, it is the default (the universal, the one which comes with the mixin) disabled styling that we want to explicitly OVERRIDE. For a Spreadsheet cell, which is disabled, we need to keep `opacity: 1;` on the cell element, so that we could fix the referenced bug. That is why I am moving the `opacity: .6;` to the inner text element. If we apply the suggested mixin, that would mean the default disabled styling will be present on both the spreadsheet cell AND the inner text container element. 

Anyways, the suggested fix results in a bug in another scenario. That is why I am closing the PR.

PR #2018, opened 2020-10-30
Update icons/_layout.scss

kakone:
Web Font Icons class names for charts don't corresponding to the [documentation](https://docs.telerik.com/kendo-ui/styles-and-layout/icons-web) or [less file](https://github.com/telerik/kendo-ui-core/blob/master/styles/web/common/font-icons.less)


joneff:
The documentation is lagging behind. We'll take care of the less themes as well.

joneff:
Superseded by telerik/kendo-ui-core#6115.

@kakone if you have found any other discrepencies, you can share them in the related issue.

PR #2032, opened 2020-11-03
fix: add editor readonly focus styles

inikolova:


joneff (approved):

joneff:
If we see the need to do so in other suites, we can modify the selectors to match any suite.

Issue #2016, opened 2020-10-29
Automate creation of files for new components


PreslavKozovski:
We need a script that creates all the needed files for a new component.

PR #2038, opened 2020-11-04
script for creating initial files for new components

PreslavKozovski:
closes: https://github.com/telerik/kendo-themes/issues/2016

After this PR is merged, I will create a Wiki - "Steps of creating a new component"

### DIFF
--- a/scripts/new-component/create-component-files.js
+++ b/scripts/new-component/create-component-files.js
@@ -1,0 +1,32 @@
+// Example usage: npm run create-component-files --name=accordion
+const fs = require('fs');
+const path = require('path');
+const nunjucks = require('nunjucks');
+const { getEnvArg } = require('../utils');
+
+const componentName = getEnvArg('name');
+const templatesDir = path.resolve(__dirname, 'templates');
+const themes = [ 'default', 'bootstrap', 'material', 'classic' ];
+const files = fs.readdirSync(templatesDir);
+
+nunjucks.configure(templatesDir, { autoescape: false });
+
+if (!componentName) {
+    throw 'Pass the component name as a parameter!';
+}
+
+themes.forEach((theme) => {
+    const dist = `./packages/${theme}/scss/`;
+    const data = {
+        componentName: componentName,
+        theme: theme
+    };
+
+    fs.mkdirSync(path.resolve(dist, componentName), { recursive: true });
+
+    files.forEach((file) => {
+        const output = nunjucks.render(file, data);
+
+        fs.writeFileSync(path.join(dist, componentName, path.basename(file, path.extname(file))), output);
+    });
+});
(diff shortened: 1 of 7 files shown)


scripts/new-component/create-component-files.js
```diff
@@ -0,0 +1,42 @@
+const fs = require('fs');
+const path = require('path');
+const yargs = require('yargs');
+const nunjucks = require('nunjucks');
+
+const componentName = yargs.argv._[0];
+const themes = [ 'default', 'bootstrap', 'material', 'classic' ];
+const files = fs.readdirSync(path.resolve(path.dirname(process.argv[1]) + '/templates'));
+
+nunjucks.configure(path.resolve(__dirname, './', 'templates'), { autoescape: false });
```
joneff, line 10: `path.resolve` adds the end slashes, so `./` is meaningless

scripts/new-component/create-component-files.js
```diff
@@ -0,0 +1,42 @@
...
+
+if (!componentName) {
+    throw 'Pass the component name as a parameter!';
+}
+
+themes.forEach((theme) => {
+    const dist = './packages/' + theme + '/scss/';
```
joneff, line 17: string interpolation

scripts/new-component/create-component-files.js
```diff
@@ -0,0 +1,42 @@
...
+    const data = {
+        componentName: componentName,
+        theme: theme
+    };
+
+    ensureDirSync(path.join(dist + componentName));
```
joneff, line 23: `fs.mkdirSync(... , { recursive: true });` is sufficient.  If it fails, it fails.

joneff (changes requested):
Use `path.join` and `path.resolve` consistently. Also, prefer comma, instead of `+`.

Finally, do we want `exports( ... )` or `exports(...)`

scripts/new-component/create-component-files.js
```diff
@@ -0,0 +1,42 @@
+const fs = require('fs');
+const path = require('path');
+const yargs = require('yargs');
+const nunjucks = require('nunjucks');
+
+const componentName = yargs.argv._[0];
```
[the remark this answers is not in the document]
PreslavKozovski: The `getArg` approach is not working when invoking a `package.json` script.

What do you think of the following approach:

```
npm run create-component-files --name=accordion
```

```
const { getEnvArg } = require('../utils');

const componentName = getEnvArg('name');
```

```
function getEnvArg(key) {
    let name = `npm_config_${key}`;
    let arg = process.env[name];

    return arg;
}
```

PreslavKozovski:
At this point, I see no value in exporting this functionality. I believe that, for now, we are going to use it directly by running the relevant node script.

joneff (approved):

PR #2055, opened 2020-11-10
Fab improvements

Juveniel:
* duplicate selectors for kendo-angular button service
* remove explicit positioning from fab popup
* flip item label an icon in rtl


joneff (dismissed):
~~New components should only have canonical class names for interactive states i.e. `hover`, `focus`. Remove classes with `-ed` suffix.~~

I stand corrected. Those classes are needed for kendo-angular, as there is a shared service that provides the class names. That services uses those old class names

silviyaboteva (approved):

joneff (approved):
I would only suggest no using inline styles in the tests, regarding the position, but that's minor thing.

Also, try not to edit tests and styles in the same commit.

PR #2056, opened 2020-11-10
fix(editor): viewHTML editor-dialog styling fixed with flex

### LINKED ISSUE #1620: Editor 'view HTML' dialog styling is broken
**Describe the bug**
When _'view HTML'_ tool is used in Safari, the text container has small height. When viewed in Chrome and FireFox the text container is too large and the buttons are not visible

Angular Safari:
MicrosoftTeams-image (1)

jQuery Safari:
Screenshot 2020-06-12 at 10 11 00

jQuery Chrome:
image

jQuery FireFox:
image

**To reproduce**
Only in Safari browser.
https://stackblitz.com/edit/angular-muc1sj?file=app/app.component.ts

In Chrome and FireFox:
[Editor All Tools Demo](https://demos.telerik.com/kendo-ui/editor/all-tools)

**Affected package**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites**
* Kendo UI for jQuery
* Kendo UI for Angular

**Affected browsers**
* Safari, Chrome, Firefox

GeorgeGindev:
Fixes #1620 

Needs testing in Safari Browser


PreslavKozovski:
I believe we should add a test for this.

joneff:
I was going to say the same thing. In addition, I hardly believe the fix is that simple. Rather, we should look at a holistic solution for this particular scenario. Still, we should start with the tests first

joneff (approved):

PR #2057, opened 2020-11-10
Input: styled input with icon on left/right

PreslavKozovski:
targets: #1978


joneff (changes requested):
That doesn't really target the real issue. At the very least, we should use something more precise than `1.9em`.

joneff (approved):

Issue #2061, opened 2020-11-12
Right resizing arrow points in wrong direction


Xizario:
`k-resize-e` looks like `k-resize-w`. 
Chrome on Mac

![Screen Recording 2020-11-12 at 9 53 23](https://user-images.githubusercontent.com/9458606/98911676-a6c8b880-24cd-11eb-9432-d79419f2521a.gif)

PR #2062, opened 2020-11-12
fix(resizing): add correct east resize cursor

silviyaboteva:
Fixes https://github.com/telerik/kendo-themes/issues/2061


joneff (approved):

joneff:
Strictly speaking the cursor in the window should be `ew-resize`, `ns-resize`, `nesw-resize` and `nwse-resize`, but I think we can leave the current behaviour as it is.

PR #2063, opened 2020-11-12
test(floating-label): add visual tests

GeorgeGindev:
targets: #1986


joneff (approved):

PreslavKozovski (changes requested):
We do not need to test all the states for this issue - https://github.com/telerik/kendo-themes/issues/1986

A single HTML test file is adding 3 PNGs.

Having said that, I would suggest adding one more row to this HTML file https://github.com/telerik/kendo-themes/blob/develop/tests/visual/floating-label.html

And removing the current `floating-label-container-font-size.html `

PreslavKozovski (approved):

joneff (approved):

PR #2064, opened 2020-11-12
fix(fab): add disabled styling for the items

silviyaboteva:


elena-gancheva (approved):

joneff (approved):
As a general note, if a component has not yet been released, we shouldn't use "fix".

Also, if you expect that someone might want to customize the disabled state, it should be part of the themes, no layout.

That being said, everything looks good to me.

joneff:
@silviyaboteva we should tests fir the sisabled as well.

packages/default/scss/fab/_layout.scss
```diff
@@ -122,6 +122,17 @@
         outline: none;
     }
 
+    // Disabled state
+    .k-fab-item.k-state-disabled,
```
joneff, line 126: Not all elements have `:disbled` pseudo class. As of Nov 2020, it applies onyl to native form elements like `button` and `input`. See https://dojo.telerik.com/@joneff/emeSETEf

As a general rule of thumb, if the element allows for it, prefer placing pseudo `:disabled` selector first, then the attribute `[disabled]` selector and then the rest of the classnames.

PR #2080, opened 2020-11-17
feat(utils): add grid-auto-flow unset class

silviyaboteva:


joneff:
Any specific reason why we might need that?

silviyaboteva:
@joneff yes, we will use the auto flow classes for the TileLayout component instead of inline styles to allow the users to change the auto flowing mechanism - https://github.com/telerik/kendo-react-private/blob/tile-layout/packages/layout/src/tilelayout/TileLayout.tsx#L16

joneff (approved):

Issue #2037, opened 2020-11-04
Rework grid toolbar styling to use flexbox


inikolova:
Currently, the Grid toolbar does not look ok on smaller screen sizes - there is no space around the input / buttons.
![Screenshot 2020-11-04 at 13 51 49](https://user-images.githubusercontent.com/1826086/98108598-e3c8f580-1ea4-11eb-8837-abd60779729a.png)

Rework toolbar styling to use flexbox and add margin to the toolbar children (similar to the standalone toolbar component). 

Expected outcome:
![Screenshot 2020-11-04 at 15 23 53](https://user-images.githubusercontent.com/1826086/98116973-c0f10e00-1eb1-11eb-89dc-9bd788291b7a.png)

joneff:
Note on kendo-jquery: a certain small fix needs to be applied in kendo-jquery!

Dimitar-Goshev:
Kendo UI for jQuery Grid's toolbar has no space around its buttons anymore:
![image](https://user-images.githubusercontent.com/6614183/101803859-a695fa00-3b19-11eb-99ab-a0c9ff4a8285.png)

Reproducible on SASS staging - Grid Batch Editing

joneff:
Yes, some styles have to be transffered

Dimitar-Goshev:
@joneff I looked at it, but as there was no linked PR at that time, the comment was a little unclear to me.

PR #2081, opened 2020-11-18
Grid toolbar reworked with flex

GeorgeGindev:
Targets #2037


packages/default/scss/grid/_layout.scss
```diff
@@ -364,9 +364,9 @@
             }
 
             .k-grid-toolbar {
-
-                .k-button + .k-button {
-                    margin: 0 $grid-toolbar-spacing 0 0;
+                padding: $grid-toolbar-padding-y $grid-toolbar-padding-x 0 0;
+                > * {
```
joneff, line 368: If we are using flex, we can use `gap`

packages/default/scss/grid/_layout.scss
```diff
@@ -376,7 +376,8 @@
             }
 
             .k-grid-search {
-                float: left;
+                margin-left: $grid-toolbar-spacing;
```
joneff, line 379: no need here for those.

see if you can manually place a spacer

packages/default/scss/grid/_layout.scss
```diff
@@ -413,7 +414,7 @@
 
         .k-grid-search {
             width: $grid-search-width;
-            float: right;
+            margin-left: auto;
```
joneff, line 417: probably this will also not be needed

tests/visual/grid-toolbar.html
```diff
@@ -60,9 +60,9 @@
                 <div class="k-header k-grid-toolbar">
                     <button class="k-button">Button</button>
                     <span class="k-button-group">
-                        <button class="k-button-icon k-button k-group-start"><span class="k-icon k-i-bold"></span></button>
+                        <button class="k-button-icon k-button k-group-end"><span class="k-icon k-i-bold"></span></button>
```
joneff, line 63: Why? The first button is the start button. If for some reason the buttons do not look correctly, it's not issue of the class names, rather, the styles. You can look at the button-group styles for a fix.

joneff (changes requested):

joneff:
Quick question: we already have the styles for toolbar. What if we add the class for toolbar to the grid toolbar:

```html<!-- current -->
<div class="k-header k-grid-toolbar">
    ...
</div>


<!-- suggestion -->
<div class="k-toolbar k-grid-toolbar">
    ...
</div>
```

Looking at the styles for toolbar, that will make the items wrap on small resolutions with the appropriate spacing. Being a styled toolbar, we can also use spacers. Finally, that will also clean up some variables, as we won't need them [when we have the classnames for styling toolbar size and color wise].

@VicTachev @tapopov @Xizario @Raisolution @inikolova thoughts?

joneff:
@silviyaboteva @inikolova we'll be moving the styles to dev!

joneff (approved):

PreslavKozovski (approved):

inikolova (changes requested):
Add **flex-shrink: 0** to **.k-grid-toolbar** otherwise toolbar squashes when the Grid is scrollable and small height is set

inikolova (approved):

ivanchev:
@joneff can myself and @jivanova get a mention when such changes are required in the future, since right now the Blazor Grid is broken, and we had no idea that this is going to happen.

joneff:
An oversight on my part. I thought I mentioned all managers in the previous comment.

Issue #1971, opened 2020-10-14
Add styles for new column menu in Grid


VicTachev:
Design issue - https://github.com/telerik/kendo-ux-private/issues/187

@PreslavKozovski will create sass themes and port them to less

PR #2085, opened 2020-11-19
column-menu: update spacings based on latest design

PreslavKozovski:
part of: https://github.com/telerik/kendo-themes/issues/1971


silviyaboteva (approved):

inikolova:
From my testing I observed the following differences with the design:

- [ ]  .k-column-chooser-title, .k-column-list-item -> padding should be the same as .k-columnmenu-item - 4px  (now 2px);

- [ ] Filter -> And/Оr DDL -> margin should be 8px (currently 14px);

- [ ] k-filter-menu .k-actions, .k-filter-menu .k-action-buttons, .k-filter-menu .k-columnmenu-actions -> padding-top should be 4px (now is 14px);

- [ ] .k-filter-menu .k-widget, .k-filter-menu .k-radio-list, .k-filter-menu .k-textbox -> margin should be 8px (currently 7px);

I will add more visual tests so we could handle these cases.

PreslavKozovski:
@inikolova About the first point, I believe that the correct padding is 2px:

image

The space between two items is 4px, so 2px padding for both items is doing the magic.

inikolova:
Adding padding to .k-grid-columnmenu-popup.k-popup makes visible border on the last item in the Angular column menu: 
Screenshot 2020-12-08 at 8 41 07

inikolova (approved):

inikolova:
The column menu looks great :)

joneff (approved):

PR #2090, opened 2020-11-26
feat(datetime): hover style for single k-link hover

GeorgeGindev:
Datetime picker was missing styles for hover on individual k-link for calendar and clock, now they are added.

https://github.com/telerik/kendo/issues/11642#issuecomment-712129067


JoomFX (approved):

packages/default/scss/datetime/_theme.scss
```diff
@@ -37,7 +37,18 @@
                         $datetime-select-hovered-border,
                         $datetime-select-hovered-gradient
                     );
+
+                    .k-link:hover {
```
JoomFX, line 41: We do not need this nesting.
The `.k-link:hover` styles should be moved one level up, so they are not nested in:

```
            // Hovered state
            &:hover,
            &.k-state-hover
```

JoomFX (changes requested):

packages/default/scss/datetime/_theme.scss
```diff
@@ -61,6 +63,15 @@
                 $datetime-select-border,
                 $datetime-select-gradient
             );
+
+            .k-link:hover {
```
JoomFX, line 67: There's no need to scope the `.k-link:hover` styles in `.k-select`, move them one level up.
As an example, you can check the [layout file](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/datetime/_layout.scss#L151).

JoomFX (changes requested):

packages/default/scss/datetime/_theme.scss
GeorgeGindev: I have moved it and also added the styles to the `.k-dateinput` scope...

However the hover styles aren't applied to the `.k-datepicker` and `.k-timepicker` because they don't contain `.k-link` elements inside the `.k-select`. This discrepancy in the HTML markup can be fixed by simply adding `k-link` class where it is missing to the `k-select`  wrapper class list.

Another way to solve this hover issue is with this temporary fix, to change the font color of the icon:
```
.k-link:hover > .k-icon,
.k-select:hover > .k-icon {
     color: primary
 }
```

JoomFX (approved):
Looks good.

This change affects the hover state styling of both the DateInput and DateTimePicker.
The design team should be informed so they can update both designs:

[DateInput design](https://xd.adobe.com/view/f3c82c38-be8d-4781-8aea-e5f4249edf2e-a69b/)
[DateTimePicker design](https://xd.adobe.com/view/9119f2b2-7508-4480-a414-24f5e3518d57-c44a/)

PreslavKozovski (changes requested):
### Is this PR related to an issue?

If yes => https://github.com/telerik/kendo-themes/wiki/Front-End-Workflow#the-procedure

If not => add the relevant labels, projects, milestones to the PR.

<hr />

### I see a `fix`, I should also see a `test`.

tests/visual/dateinputs.html
```diff
@@ -35,6 +35,12 @@
...
         <span class="k-widget k-dateinput"><span class="k-dateinput-wrap k-state-hover"><input type="text" class="k-input" value="Hovered" /><span class="k-select"><span class="k-link k-link-increase"><span class="k-icon k-i-arrow-n"></span></span><span class="k-link k-link-decrease"><span class="k-icon k-i-arrow-s"></span></span></span></span></span>
 
+        <!-- jQuery - hovered button  -->
+        <span class="k-widget k-datepicker"><span class="k-picker-wrap k-state-hover"><input type="text" class="k-input" value="Hovered button" /><span class="k-select"><span class="k-icon k-i-calendar"></span></span></span></span>
+        <span class="k-widget k-timepicker"><span class="k-picker-wrap k-state-hover"><input type="text" class="k-input" value="Hovered button" /><span class="k-select"><span class="k-icon k-i-clock"></span></span></span></span>
```
PreslavKozovski, line 40: The buttons of these pickers are not "hovered".

PreslavKozovski (changes requested):
We should also check why there is a change in the `ripple` Bootstrap test.

tests/visual/dateinputs.html
GeorgeGindev: Because there is no k-link there, k-select already has hovered effect

PreslavKozovski (approved):

joneff (approved):

Issue #1111, opened 2019-12-06
Ripple effect is not positioned as expected over the checkbox and radio button elements.


simonssspirit:
**Describe the bug**
Ripple effect is not positioned as expected over the checkbox and radio button elements.

**To reproduce**

This can be observed by clicking over the radio buttons or the checkboxes:

https://www.telerik.com/kendo-react-ui/components/ripple/ - KendoReact
https://www.telerik.com/kendo-angular-ui/components/ripple/ - Kendo UI for Angular

**Expected behavior**

**Screenshots**

**Affected package (please provide)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please provide)**
* Kendo UI for Angular
* Kendo UI for React

**Affected browsers (please provide)**
- Firefox: [for example, 67, lts]
- Chrome: [for example, 77]
- Edge version: [for example, 15]
- Safari: [for example, 12]
- IE version: [for example, 11]

**Build system information (please provide)**
- OS: [for example, Windows, Linux, Mac]
- Node version: [for example, 10.15.3]
- NPM version: [for example, 6.4.1]

**Additional context**

joneff:
@eeasss @tsvetomir @Xizario @tapopov @Raisolution 

Here is the issue: previously, we've shown the the ripple inside `k-radio-label` / `k-checkbox-label` in a fixed location.

Now, with natively styles checkbox / radio, they can be anywhere in the label, including outside of it. Since the activation of the ripple relies on manually supplying selectors we could do one of three things:

1) apply the ripple only when checkbox / radio is within label
2) apply the ripple regardless where checkbox radio is
3) in addition to 1) apply it when the checkbox is inside k-ripple-container element

To me, 1) and 3) seem like the best choices: either place the label in label or explicit ripple container.

**Possible issues**

With 2) there are definitely issues: we can't know what the checkbox parent element is. it could be in a cell, in a clipped div ... in short, it's unreliable.

That will require some modification of kendo-ripple npm package. The quick workaround is to hide the ripple for now in labels.

Thoughts?

Raisolution:
Vanka,

sorry if I am missing something here. Didn't we always demoed the Ripple as a container with checkbox elements that have their labels as direct siblings. I noticed that this approach is currently used in all suites.
E.g.
```
<div kendoRippleContainer>
    <input type="checkbox" id="c1" class="k-checkbox" />
    <label class="k-checkbox-label" for="c1">Checkbox 1</label>
</div>
[...]

PR #2091, opened 2020-11-26
Checkbox radio ripple

inikolova:
Part of https://github.com/telerik/kendo-themes/issues/1111


joneff (approved):
We may decide to introduce `$checkbox-ripple-bg` variable at some point, but even without it, the implementation is splendid.

PR #2097, opened 2020-11-30
fix(fab): line height issue in material and bootstra…

silviyaboteva:
…p themes

The commits are reworded from a chore to fix as the styles are already live.
closes: https://github.com/telerik/kendo-themes/issues/2100


joneff (approved):

joneff:
@silviyaboteva `floatingactionbutton` is quite a mouthful. You can use `fab` for commits. Developers will still know which component the fix is referring to ;)

Juveniel (approved):

inikolova (approved):

joneff (changes requested):
the line height doesn't look good to me as value. we should perhaps find another approach

joneff (approved):

PR #2101, opened 2020-12-01
chore(fab): use solid colors for disabled states

Juveniel:
The floating button and its items should stay above other elements. Thus, when the fab/fab items are disabled, we should use solid colors.

related to: https://github.com/telerik/kendo-themes/issues/1948


packages/default/scss/fab/_theme.scss
```diff
@@ -85,6 +99,18 @@
...
     }
 
+    // Disabled state
+    .k-fab-item.k-state-disabled,
+    .k-fab-item:disabled {
+        opacity: 1;
+
+        .k-fab-item-text,
+        .k-fab-item-icon {
+            color: tint( $fab-item-text, 6 );
```
silviyaboteva, line 109: The text color in secondary, warning and light themeColor variants is sharper than before. Can we make it lighter? Like the color of the disabled item.

joneff (approved):

packages/default/scss/fab/_theme.scss
joneff: we can tackle this at a later point

PR #2107, opened 2020-12-03
fix(datetimepicker): make space for 5 timepicker columns

JoomFX:
When using the DateTimePicker, there is not enough space to accommodate 5 columns in the Timepicker.
In the screenshot below, the last column (AM/PM) gets hidden as there is no space for it:

timepicker


packages/bootstrap/scss/datetime/_variables.scss
```diff
@@ -10,12 +10,15 @@ $timepicker-header-height: 2em;
...
+$time-list-width: 4em !default;
 $time-list-height: 200px !default;
 
+$time-list-item-padding-x: $padding-x !default;
+$time-list-item-padding-y: $padding-y !default;
+
 $dateinput-text: null !default;
 
-$datetime-width: ((4 * $time-list-width) + 1em) !default;
+$datetime-width: 21em !default;
```
joneff, line 21: leave the formula. use calc just in case

packages/bootstrap/scss/datetime/_variables.scss
```diff
@@ -10,12 +10,15 @@ $timepicker-header-height: 2em;
 
 $time-list-padding: $padding-y * 5 !default;
 $time-list-focus-size: 2px !default;
-$time-list-width: 5em !default;
+$time-list-width: 4em !default;
```
joneff, line 13: this is bootstrap, so perhaps it's best to use `rem`

joneff (changes requested):
This is a feature: `feat(datimepicker): add milliseconds column in popup picker` or something similar.

joneff (approved):

Issue #410, opened 2017-09-01
Sorting icon is not visible when header text overflows


inikolova:
Sorting icon is not visible when header text is long but the column width is small. Can be observed in the Master-Detail Grids demo -> http://www.telerik.com/kendo-angular-ui/components/grid/hierarchy/

To reproduce the issue sort the CategoryID column

vsviridov:
> A workaround for this issue is to move the sort indicator icon before the header text, for example with the following rule:

Does not work correctly with multi-sort...
![image](https://user-images.githubusercontent.com/221684/50869094-7f4fb480-1367-11e9-8018-b7f57ec345d5.png)

vsviridov:
Not sure why the un-assignment happened, that was not the intention of the comment...

TeyaVes:
A possible workaround would be:

```
 .k-grid-header .k-header>.k-link {
        padding-right: 1.5em;
}

.k-grid-header .k-header>.k-link>.k-icon {
        position: absolute;
        top: 50%;
        right: 0.5em;
        transform: translateY(-50%);
        margin-left: 0;
}

.k-grid-header .k-sort-order {
        position: absolute;
        right: 0.25em;
}
```
Live example here: [https://stackblitz.com/edit/angular-x8dpcl](https://stackblitz.com/edit/angular-x8dpcl)
In this way, the sorting icon and the number will be aligned to the right of the cell:


![2019-03-25_1107](https://user-images.githubusercontent.com/45764088/54907478-41861500-4eee-11e9-9e4e-19d52a6635d9.png)

@joneff @lkarakoleva , what do you think?

DaleyKD:
@TeyaVes : That's the best one yet.  However, it does look a bit odd with the filter icon, and if filtering is enabled, you can't see it.

I can't believe this hasn't been resolved yet.  I'm using 3.4.0

jadamrogers:
I ran across this post specifically because I was having this issue and wanted to move the sort indicator and number to the left side of the header text. After much work, I found a very good solution that only requires CSS.

`
/* sort icon color */
.k-grid-header .k-header > .k-link > .k-icon:before {
    color: #f2b661;
}
/* sort order number color */
.k-grid-header .k-header > .k-link > span.k-sort-order {
    color: #f2b661;
}

/* move sort icon and number to left and shift text right to make room for them when a column is sorted */
.k-grid-header th.k-header.k-sorted > .k-link {
    position: relative;
    padding-left: 18px;
}
[...]

PR #2117, opened 2020-12-07
Grid sorting indicator

inikolova:
Discussion: https://github.com/telerik/k2/issues/280

Part of: https://github.com/telerik/kendo-themes/issues/410


PreslavKozovski (approved):
The changes LGTM. However, you could order the properties as described here - https://github.com/telerik/kendo-themes/wiki/Code-Styleguide#declaration-order

joneff (approved):
I see only two minor things:

* order of props
* alignment of k-cell-inner: do we need space between the items?

those are minor, so we can go forth with the merge

silviyaboteva (approved):
In React Grid, we have not planned to change the rendering yet, but I tested and the new styles don't break the current layout.

PreslavKozovski (changes requested):
These styles are not going to work in an RTL scenario.

PreslavKozovski (approved):

joneff (approved):

PR #2123, opened 2020-12-08
fix(calendar): double border in angular classic calendar

JoomFX:
Since there are nested `.k-calendar` elements, there is a double border:

cal-border


joneff:
In the image tests it's visible that the bottom border of kendo-jquery-calendar is cut off. Please test in firefox and determine if this is browser behaviour, or legitimate issue.

joneff (approved):

PR #2124, opened 2020-12-08
chore(fab): apply pointer cursor on item

Juveniel:
related to: #1948

Cursor is currently applied to the icon only. However, as we use template items, the cursor should be applied over the entire item instead.


silviyaboteva (approved):
Mitak, can you rename the `chore` commit to `fix` as the styles are already live.

joneff (approved):

Issue #2098, opened 2020-12-01
Revise Slider rendering


jivanova:
### Reason

**React**
``` html
 <!--  Flex positioning through tons of inline styles. -->
<div class="k-slider k-state-default k-slider-horizontal">
  <div class="k-slider-wrap k-slider-buttons" style="display: flex; position: relative; flex-direction: row;">
    <a class="k-button k-button-decrease" title="Decrease" style="position: relative;">
       <span class="k-icon k-i-arrow-w"></span>
    </a>
    <div class="k-slider-track-wrap" style="flex-grow: 1; position: relative; touch-action: none; margin-left: 0.5rem; margin-right: 0.5rem;">
        <ul class="k-reset k-slider-items" style="margin: 0px;">
          <li role="presentation" class="k-tick k-tick-large" style="z-index: 1; position: absolute; left: 0%; width: 1px;">
            <span data-slider-label="true" class="k-label" style="position: absolute; width: 40px; left: -20px;">1</span>
          </li>
           <!-- more ticks -->
        </ul>
        <div class="k-slider-track" style="left: 0px; width: 100%;">
           <div class="k-slider-selection" style="width: 66.6667%;"></div>
           <a class="k-draghandle" title="Drag" style="left: calc(66.6667% - 8px); z-index: 1;"></a>
        </div>
   </div>
   <a class="k-button k-button-increase" title="Increase" style="position: relative;">
     <span class="k-icon k-i-arrow-e"></span>
   </a>
  </div>
</div>
```
**Angular**
``` html
 <!--  Manually calculated widths for every inner part of slider. -->
<kendo-slider dir="ltr" class="k-slider-horizontal k-slider k-slider-transitions k-widget k-state-default" style="width: 226px;">
    <!--  Classes for label position are set here. ex: `k-slider-topleft`  -->
    <div class="k-slider-wrap k-slider-buttons" style="position: relative;">
           <a class="k-button k-button-decrease">
               <span class="k-icon k-i-arrow-w">
                </span>
           </a>
           <a class="k-button k-button-increase">
                <span class="k-icon k-i-arrow-e">
                </span>
          </a>
          <ul class="k-reset k-slider-items">
                <li class="k-tick k-first" style="width: 5px;">
                </li><li class="k-tick" style="width: 10px;">
                </li>
                     <!-- other ticks -->
          </ul>
          <div class="k-slider-track" style="width: 150px;">
                <div class="k-slider-selection" style="width: 0px;">
                </div>
[...]

PR #2127, opened 2020-12-09
Slider css vars

joneff:
Introduce slider custom properties
* slider-start used to determine the beginning of the selection and start thumb of range slider
* slider-end used to determine the end of the selection and end thumb for slider / range slider

Note: we could use intricate writing modes to use a single piece of styles and then just change the writing mode, but that's not as widely supported as we would like.


jivanova (commented):
Related to https://github.com/telerik/kendo-themes/issues/2098

- `inset-inline-start` has no support on mobile browser, and has official support in Chrome since v87 which was recently released. We need to update the theme implementations to cover mobile browsers as well.
- Ticks positions does not seem to be handled - we have **before, after, both** values - https://www.telerik.com/kendo-angular-ui/components/inputs/slider/ticks/#toc-width

joneff:
Changed `inset-inline-start` to `left` and `right` respectively.

The other bit puzzles me: if you add the class to `k-slider-track-wrap` it seems to be working OK with the existing styles.

jivanova (approved):

PR #2129, opened 2020-12-09
fix: add generic spacer styles

inikolova:
I am adding this class / style so we could avoid repetition of styles in different components.

### DIFF
--- a/packages/classic/scss/utils/_spacer.scss
+++ b/packages/classic/scss/utils/_spacer.scss
@@ -1,0 +1,1 @@
+@import "~@progress/kendo-theme-default/scss/utils/_spacer.scss";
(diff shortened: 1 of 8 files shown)


PreslavKozovski (changes requested):
Do not forget to add this util class to other themes :)

joneff (approved):

joneff:
Note: in the future we'll also add sized spacers

packages/classic/scss/utils/_spacer.scss
```diff
@@ -0,0 +1 @@
+@import "~@progress/kendo-theme-default/scss/utils/_spacing.scss";
```
silviyaboteva, line 1: Change to `_spacer.scss` instead.

silviyaboteva (changes requested):

PreslavKozovski (approved):

silviyaboteva (approved):

PR #2089, opened 2020-11-25
Bottom Navigation Styles

Juveniel:
targets: https://github.com/telerik/kendo-themes/issues/1949

PR #2131, opened 2020-12-10
Bottom nav improvements

Juveniel:
related to: #2089 

* Added shading for secondary and light flat variants to improve visibility of selected & focused state.
* Adjusted focus styles by making the flat variant lighter and also added spacing for the default theme only (synced with @tyanko )


joneff (approved):

silviyaboteva (approved):
🚀

Issue #1626, opened 2020-06-15
Toolbar appearance in PDFViewer is not correct on smaller screen sizes


veselints:
### Bug report
Toolbar appearance in PDFViewer is not correct on smaller screen sizes. That applies to all LESS and SASS thems.

Related to:
https://github.com/telerik/kendo/issues/11103

### Reproduction of the problem
Reduce the window width and run the following demo:
https://demos.telerik.com/kendo-ui/pdfviewer/index

The result would be:
![PDFViewerToolbarSass](https://user-images.githubusercontent.com/3431388/84630499-b038ff00-aef4-11ea-95e8-9a5567461c44.png)

### Expected/desired behavior
Toolbar tools should not overlap the overflow button.

### Environment
* **Kendo UI version:** 2020.2.513
* **Browser:** [all]

joneff:
As noted in the related PR, the issue cannot be fixed with the themes, but rather in the js code.

PR #2132, opened 2020-12-10
fix(pdfviewer): remove gap from toolbar children

GeorgeGindev:
targets: #1626 

Looks like the gap property isn't included in the calculations for the outer width of toolbar children. 
It works correctly with margin in place of gap.


joneff:
@GeorgeGindev the whole point is to use gap instead of margin. margins are unreliable: they tend to collapse, they tend not to be direction independent.

Unlike gap.

As discussed, the issue can be solved by modifying the toolbar code to include value for gap in calculation.

closing.

PR #2146, opened 2020-12-15
fix: vertically center checkbox in grid cell

inikolova:


packages/bootstrap/scss/grid/_variables.scss
```diff
@@ -24,6 +24,8 @@ $grid-cell-padding-y: $grid-padding-y !default;
 $grid-cell-horizontal-border-width: 0px !default;
 $grid-cell-vertical-border-width: 1px !default;
 
+$grid-cell-checkbox-margin-y: calc( #{$cell-padding-y} - (#{$button-calc-size} - #{$line-height-em}) / 2 ) / 1.25  !default; // $grid-command-cell-padding-y / 1.25
```
PreslavKozovski, line 27: This is not a valid value. Here, we are assuming that the `postcss-calc` will remove the `calc`, however, if any of these variables change, and the `calc` is not removed, we will have an invalid output of the CSS.

PreslavKozovski (changes requested):
The below comment is applicable to most of the `$grid-cell-checkbox-margin-y` variables.

PreslavKozovski (approved):

joneff:
Due to upcoming changes in the buttons, this fix will not be relevant as it depends on variables related to the current state of the button. Converting to draft for the time being

joneff:
To align checkboxes correctly, we need a wrapper around them (`k-checkbox-wrap`). Then and only then we can align them correctly, be it top or center align.

PR #2147, opened 2020-12-15
fix(dropdowngrid): remove unicode space character

kspeyanski:
Fixes a regression introduced in the following commit: https://github.com/telerik/kendo-themes/commit/e01bee22d4f7762a2618a1516c034679127c6ddc

Somehow, 8 [zero-width-space](https://www.fileformat.info/info/unicode/char/200b/index.htm) characters appeared. This led to incorrect distribution of the `kendo-theme-default`, with the final `all.css` file being not being parsed correctly in the browser:
image

cc, @InaGlushkova 

edit: Please check your linters, as such symbols are incorrect and can not be properly visualised during a PR process.


silviyaboteva (approved):

joneff:
@kspeyanski it's odd for something like this to appear. zero width space is not a trivial character that can be easily produced. I'll see what we can do about linters

joneff (approved):

PR #2156, opened 2020-12-16
Input separator fixes

inikolova:


PreslavKozovski (approved):

silviyaboteva (approved):

joneff (approved):
The title of the tests can be changed to `textarea -- angular` or something similar and we could place the textareas on the same line.

those are not blockers, so we are merging this one

PR #2163, opened 2020-12-17
fix(scrollview): wrap the content inside scrollview pages

GeorgeGindev:


joneff (changes requested):
it's hard to imagine what exactly you want to achieve. add a test that shows

GeorgeGindev:
@joneff It looks like the problem is demo-related only. This branch can be closed now.

PR #2166, opened 2020-12-17
Add summary 'horns' using clip-path

### LINKED ISSUE #919: Summary task which has only one Milestone task does not render its milestone visualization in the Timeline
**Describe the bug**
A Summary task which has only one Milestone task does not render its milestone visualization in the Timeline.

**To reproduce**
Run the following Dojo: https://dojo.telerik.com/UVuVAcoG/3
The summary (parent) task is not visualized in the Timeline.

**Expected behavior**
The summary (parent) task should be visualized in the Timeline.

**Affected package (please provide)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please provide)**
* Kendo UI for jQuery
* Kendo UI for Vue

**Affected browsers (please provide)**
ALL

joneff:
note: will not work in IE browsers because clip-path does not work in IE.

fixes #919.

@Dimitar-Goshev @mparvanov FYI this will make the summary a tad higher in non supporting browsers.


veselints (changes requested):
The horns change results in two issues when rendering delayed summary tasks:
- For such tasks the horn is rendered at the end of the planned portion of the task, while it should be rendered at the end of the actual execution (the delayed part) of the task;
- The delayed part of the task is rendered with the height of the horn, while it should have the same height as the planned part of the task.

Here is the current result in Default SASS:
PlannedSassDefault

Here is the expected result as it is currently rendered in Default LESS;
PlannedLess

joneff:
The fix is becoming non-trivial. I am putting this one on hold.

joneff:
This PR is outdated. Closing.

Issue #2105, opened 2020-12-02
When the widths of the Splitter panes exceed the width of the component, panes visibly overflow


dtopalov:
**Is your feature request related to a problem? Please describe.**
When the Splitter component has an explicitly set width, and the sum of the widths of the panes exceeds the preset Splitter width, the panes overflow. This could happen both when one or more panes have "size" set such that the panes width exceeds the Splitter width, and when one or more panes have "max" set such that when resizing the pane it goes beyond the boundaries of the container:

1) Middle pane overflows:
https://stackblitz.com/edit/angular-14fhxv-xmmnyc?file=app/app.component.ts

2) Resize the Middle pane to the right
https://stackblitz.com/edit/angular-14fhxv-gticyo?file=app/app.component.ts

**Describe the solution you would like to see implemented**
Addressing the overflow with a scrollbar seems to be a more intuitive behavior:

https://stackblitz.com/edit/angular-14fhxv-vpftxf?file=app/app.component.ts

**Describe the alternatives you have considered**
All other suggestions to make the Splitter behavior in such scenarios more user-friendly are welcome.

joneff:
We could provide utility classes -- `.k-overflow-*` that developers themselves attach **when needed**. Would that work?

Issue #2065, opened 2020-11-12
Add a class that sets touch-action and user-select to 'none' when dragging


dtopalov:
**Is your feature request related to a problem? Please describe.**
When KendoDraggable is used for dragging-based functionalities like reordering, resizing, drag selection, etc., the pointer-cancel event is thrown on mobile devices, effectively stopping the drag action, and triggering the **handleRelease** event.

To prevent this, we need to dynamically add the **touch-action: none** style to a suitable element while dragging.

There are similar concerns, related to the **user-select: none** style that prevents text selection while dragging.

**Describe the solution you would like to see implemented**
A dedicated new class(es) applying the described styles will allow us to add/remove the said class(es) instead of applying and removing specific styles directly in the DOM.

**Describe the alternatives you have considered**
The currently available alternative is to add/remove the custom styling manually for the related elements, but having a dedicated class that can be reused across all suites seems a cleaner and more scalable solution.

**Additional context**
Currently there is a class, scoped to the Grid header that is handling the scenario for draggable columns:
https://github.com/telerik/kendo-themes/blob/4f71b79b60646e7898f5f04aa9c643ecfa3f0a95/packages/default/scss/grid/_layout.scss#L627

joneff:
This will make a fine addition to our utils framework

joneff:
@dtopalov, looking at the less themes, the same class is called `k-prevent-selection`. It's not a bad name, but since we are naming our utility classes after the property they target, we are calling this one `k-user-select-none`.

PR #2175, opened 2020-12-22
Extend utilities

joneff:
Add utility classes for:
* overflow
* horizontal and vertical padding and margin
* clear / float
* pointer events / touch action / user select

A tiny refactoring in slider

Fixes: #2065. Fixes: #2105.


PreslavKozovski (approved):
Before merging, we should change the util classes in the Grid rendering, add prepare a similar change for the Less themes.

PR #2184, opened 2020-12-23
feat(chart): new series colors

tsvetomir:
* Replaces chart series colors according to updated design from https://github.com/telerik/kendo-ux-private/issues/178.
* Adds darker and lighter series color variants for a total of 30 colors.
* Adds a variable for controlling area series inactive opacity.
* Replace a rule used only in tests (see #521) with a run-time patch.

**Question**
This is a breaking change. Should we mark it as such?


joneff (approved):

joneff:
@nedevp I've introduced the following changes:
* default colors update will be for v5 of the themes
* bootstrap colors teal, blue, pruple, red, yellow, green point to their respective variables
* material colors also point to same named variables, instead of the colors you provided
* I still use the same basic principle for darken / lighten for bootstrap and material, but for v5 I will update them to use the respective framework way

PR #2189, opened 2021-01-01
infra: add script for generating utils docs

PreslavKozovski:
After checking and testing different approaches, I believe that generating [JSON data with DSS](https://github.com/DSSORG/DSS) from the comments is the way to go. 

Indeed, the classes are not included in the JSON, and we have to add them manually via the `@name` parser. However, this helps when documenting the cycle-generated classes.

I also tested complete solutions, for example, [KSS](https://github.com/kss-node/kss-node). As they serve a single purpose for a style guide - they are too restrictive and not feasible in our use case.

You could check the output here - https://github.com/telerik/kendo-themes/tree/util-docs-dss/packages/default/scss/utils#kendo-ui-themes-utility-classes

Any suggestions are welcome.

cc: @telerik/kendo-front-end


joneff:
@PreslavKozovski @silviyaboteva take a closer look at `material/input/theme`: prior this change it has `@include exports("input/theme")` which is the same usage in `default/input/theme`. The string inside `exports` is unique and we use it to "cache" the file so we don't duplicate output. In terms we cache empty styles. It's quite normal to have visual changes in material theme as soon as we use the actual styles and not empty styles :)

joneff (approved):

PR #2198, opened 2021-01-06
Clear button focus

inikolova:


PreslavKozovski (approved):
The changes in the code are okay. We should investigate what is causing so many visual test changes.

silviyaboteva (approved):
I agree with Presli about the visual tests.

joneff:
@PreslavKozovski @silviyaboteva take a closer look at `material/input/theme`: prior this change it has `@include exports("input/theme")` which is the same usage in `default/input/theme`. The string inside `exports` is unique and we use it to "cache" the file so we don't duplicate output. In terms we cache empty styles. It's quite normal to have visual changes in material theme as soon as we use the actual styles and not empty styles :)

joneff (approved):

inikolova:
I will revert the previous width of the icons

PR #2209, opened 2021-01-07
fix(inputs): angular checkbox and radio labels on the same line

silviyaboteva:
The issue can be seen on Angular Forms Guideline page: 
 - https://www.telerik.com/kendo-angular-ui/components/forms/forms-guideline/#toc-form-controls
 - https://www.telerik.com/kendo-angular-ui/components/forms/forms-guideline/#toc-checkboxes--radiobuttons

The reason that Angular checkboxes and radios have their labels on a new line is the difference in the rendering compared to the other suites. Here, we have a host element that is inline and there is a `<label>`, that is a block element.

### DIFF
--- a/packages/default/scss/checkbox/_layout.scss
+++ b/packages/default/scss/checkbox/_layout.scss
@@ -144,6 +144,9 @@
     kendo-label.k-checkbox-label > .k-label:first-child {
         margin-right: $checkbox-label-margin-x;
     }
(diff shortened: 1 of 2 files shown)


elena-gancheva (approved):
Thanks! ❤️

JoomFX (approved):

inikolova (approved):

joneff (approved):
at some point we are going to remove those, say version 5

PR #2211, opened 2021-01-07
Stepper uses progressbar element but does not include styling

webmandman:


joneff (changes requested):
Can you squash all commits in a single one and name it something like `fix(stepper): stepper uses progressbar element but does not include styling` ?

The `fix(stepper):` part is needed for our semantic release to make this commit visible in our changelog

Btw, nice catch!

webmandman:
I messed up the squash so I just created another PR #2244

PR #2220, opened 2021-01-11
chore: remove redundant width from adaptive timezone editor

Juveniel:
Reported from Sitefinity client (tool validation) that units are not specified for the width property.

Design ref: https://github.com/telerik/kendo-ux-private/issues/13

Note: width is not required due to [flex](https://github.com/telerik/kendo-themes/blob/86d6b6dcee890dafa0c3dc0b8cf515d0b69a1cff/packages/default/scss/adaptive/_layout.scss#L464).

### DIFF
--- a/packages/default/scss/adaptive/_layout.scss
+++ b/packages/default/scss/adaptive/_layout.scss
@@ -457,7 +457,6 @@
             }
 
             .k-mobiletimezoneeditor {
-                width: 100;
                 display: flex;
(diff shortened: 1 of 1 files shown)


Juveniel:
PR in Less-themes: https://github.com/telerik/kendo/pull/12010

joneff (approved):
It's not redundant, but `width: 100` doesn't mean anything -- it needs value.

PR #2232, opened 2021-01-12
fix(dropdowntree): horizontal scroll not always visible

stnikolova:
When the `width` is limited, the horizontal scrollbar should be fixed at the bottom

Current behaviour:
horizontal_scroll_not_sticky

With fix:
horizontal_scroll_sticky


joneff (changes requested):
The dropdowntree overflows outside of the container in the tests:
* either the test must be improved
* or the fix causes another bug

stnikolova:
No need of changes in the theme. Issue is fixed by https://github.com/telerik/kendo-angular-dropdowns/pull/772

PR #2236, opened 2021-01-13
chore(editor): export as select with icon telerik/kendo#11997

veselints:


joneff (approved):

joneff:
This should be a "fix", but is OK

PR #2239, opened 2021-01-15
fix(window): resize handles overflow container

Juveniel:
targets: https://github.com/telerik/kendo-ui-core/issues/3700
related to: https://github.com/telerik/kendo-ui-core/issues/4940

Resize handles should be positioned inside the Window wrapper to prevent overflow.

The issue can also be reproduced with the React Window([Stackblitz](https://react-xnw7dy.stackblitz.io/)).


silviyaboteva (approved):

joneff (approved):

joneff:
@Juveniel there is the small possibility that a resizable element is placed inside the window. Should that happen, we'll need to make the selector a tad more specific.

PR #2258, opened 2021-01-27
Update draghint styles

inikolova:
Currently the drag-clue font-size / line-height differ from these of the grouping indicator. In my opinion, they should be equal


silviyaboteva (approved):
I'm fine with the changes, however, if there is a design for the drag hint, it should be updated.

joneff (approved):
The design will be updated at a later state

PreslavKozovski (approved):

PR #2261, opened 2021-01-29
Utility classnames for border related properties

JoomFX:
**Border width**

We can now specify border to elements by using `k-border-*` and `k-border-SIDE-*` classes e.g. `k-border-top` or `k-border-x-0`.

Since both `k-border` and `k-border-1` made sense for default behaviour, there is a **default** property in the map apart from our usually numeric props.

Added class names for: all, top, right, bottom, left, x (on both horizontal sides), y (on both vertical sides).

**Border style**

This one is pretty standard `k-border-STYLE`. Added class names for `solid`, `dashed`, `dotted`, `double`, `none`. There are no directional classes e.g. `k-border-solid-top`.

**Border color**

Straight from `$theme-colors` -- `.k-border-COLOR`.

**Border radius**

We can now specify border-radius to elements by using `k-rounded-*` and `k-rounded-SIDE-*`. Default border radius is determined the same way default border width is.

Added class names for: all, top, right, bottom, left. Pill / circle are not yet added, but should follow.

As a final note: we are not worried that someone will specify both border-color primary and border-radius primary. That's up to developers to decide.


packages/default/scss/utils/_border.scss
```diff
@@ -0,0 +1,26 @@
+@include exports( "utils/border" ) {
+
+    // Border utility classes
+    $utils-border: (
+        "borders-all",
```
PreslavKozovski, line 5: I do not like the idea of `k-borders-all`. I believe `k-border` is okay. With this, we can remove the `if/else` logic.

@joneff @inikolova thoughts?

JoomFX: I also do not like it :)
But we already discussed it with @joneff and agreed on the class name to be `k-borders-all`, since we already have `k-border` which can be seen [here](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/dataviz/_layout.scss#L183).

inikolova (approved):

packages/default/scss/utils/_border.scss
joneff: the border class in that particular snippet is no longer in use. i guess we can safely remove it

joneff (approved):

joneff:
On a side note: with each utility file containing both the code and the variables it may be somewhat difficult to customize certain values. We are going to improve that.

inikolova (approved):

silviyaboteva (approved):

PreslavKozovski (approved):
By having more and more utility styles, I believe it is a better idea to move them to a new repo and create a new package. For example - `kendo-theme-utils`. In the themes, I think, we should have only the mandatory utilities. For example, `k-hidden`.

joneff:
For v5 of the themes it makes sense to move the utils into a separate package.

PR #2262, opened 2021-01-29
feat(utils): add font-size utils

JoomFX:


PreslavKozovski (changes requested):
Add these classes to all themes.

JoomFX:
The other themes import the default theme. Here they are:

[Bootstrap](https://github.com/telerik/kendo-themes/blob/develop/packages/bootstrap/scss/utils/_text.scss)
[Material](https://github.com/telerik/kendo-themes/blob/develop/packages/material/scss/utils/_text.scss)
[Classic](https://github.com/telerik/kendo-themes/blob/develop/packages/classic/scss/utils/_text.scss)

PreslavKozovski (approved):
my bad :)

joneff (changes requested):
I think we should go with `k-fs-*`. The same way we do with margin / padding -- https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/utils/_spacing.scss

inikolova (approved):

joneff (approved):

PR #2265, opened 2021-01-29
Traling comma breaks build with WebCompiler VS addon

joneff:
Reported by customer

TID: 1504251


PreslavKozovski (approved):

jivanova:
The compilation works as expected now :)

Issue #2253, opened 2021-01-22
Wrong border color around invalid files


inikolova:
**Describe the bug**
Wrong border color around invalid files
![Screenshot 2021-01-22 at 17 56 17](https://user-images.githubusercontent.com/1826086/105513557-21d2fa00-5cdb-11eb-8f14-25b29b974b56.png)

**To reproduce**

1. Go to https://www.telerik.com/kendo-angular-ui/components/uploads/upload/forms/#toc-template-driven-forms;
1. Upload different from image format;
2. Click outside the Upload to trigger the change detection.

**Expected behavior**
Only the component should get red border when validation fails, but not the list items

PR #2295, opened 2021-02-11
Upload invalid border

inikolova:
Fixes: https://github.com/telerik/kendo-themes/issues/2253


joneff (approved):

silviyaboteva (approved):
We can add tests for other upload files too, not only Angular one.

Issue #2234, opened 2021-01-12
Investigate Color Distance in the Default/Boostrap/Material Theme


eeasss:
It seems that the default theme uses more colors that what our color palette would suggest.

@joneff is to sync with the designers to identify the deviations and reduce the usages to the colors available in the palette.

Done

* [x] Use correct styling for input buttons (#2296)
* [x] Reduce color distance in chart (#2303)
* [x] Remove redundant upload variables (#2285)
* [x] Reduce color distance in progressbar and slider (#2306)
* [x] Use `$subtle-text` variable for styling cards styling (#2330)
* [x] Use subtle-text variable in adaptive scheduler styling (#2329) 
* [x] Use `$subtle-text` for calendar styling (#2328)
* [x] Use `$subtle-text` variable for upload styling (#2332)
* [x] Use `$subtle-text` variable for input styling (#2320)
* [x] Use `$subtle-text` variable for tabstrip styling (#2334)
* [x] Use subtle text variable for colorpicker styling (#2338)
* [x] Use subtle text variable for menu styling (#2338)
* [x] Use subtle text variable for split button styling (#2338)
* [x] Remove opacity of window titlebar actions in material theme (#2338)
* [x] Use correct colors for multiselect in material theme (#2338)
* [x] Improve chip icon and remove icon colors in material theme (#2338)
* [x] use `$subtle-text` variable for rating styling (#2340)
* [x] use `$subtle-text` variable for form styling (#2340)
* [x] use correct color determination for `$app-bg` (#2340)
* [x] use correct color determination for `$popup-bg` (#2340)
* [x] update breadcrumb styles to match material and bootstrap design (#2340)
* [x] update switch styles to match material and bootstrap designs  (#2341)
* [x] Remove duplicated drag drop styles (#2342)
* [x] reduce colors used by image editor (#2354)
* [x] use subtle-text and disabled-text variables for stepper styling (#2353)
* [x] bootstrap-theme: replace usages of darken with shade (#2352)
* [x] classic-theme: replace usages of darken with shade (#2352)
* [x] default-theme: replace usages of darken with shade (#2352)
* [x] material-theme: replace usages of darken with shade  (#2352)

eeasss:
This item will get more specific once we setup the foundations of the Kendo Front-End docs and with base variables per theme available.

PR #2296, opened 2021-02-11
Use correct styling for input buttons

joneff:
part of #2234

What's changed:

* states in material theme are no longer tied to opacity. Instead use `$input-placeholder-text` for normal state and `$input-text` for hover state.
* reduced width of numeric buttons
* correct border position of colorpicker
* default theme has normalized clicked state for numeric / dateinput.
* cleaned up redundant selectors and styles

As a result, there is more theme reuse.


JoomFX (approved):

inikolova:
Most of the changes look great, however I would like to make sure the following ones are intended:

- Default
	 -- combobox.png -> Missing shadow -> Open Combobox
	 -- dateinputs-angular.png -> Invalid + Focus shadow is changed
	 -- dateinputs.png -> Active background is changed
	 -- inputs.png -> Numeric -> hover, active states -> different background

-Material
	--dateinputs.png -> Invalid icons should not get the error color
	-- grid-editing-custom-editor.png -> Invalid icons should not get the error color
	-- toolbar-components-rtl.png -> DropDownList in Toolbar - Open state - wrong icon direction

- Bootstrap
	-- combobox.png, toolbar-components-rtl.png -> ComboBox Open state does not get focused shadow
	-- dateinputs.png -> Active state - select background

@joneff , please check and confirm :) Thank you in advance :)

joneff:
As a general rule, the shadow should be set by `focus` and not `active` (open) state. With that in mind, there is my response:

**Default**
* combobox shadow: open combobox should not have shadow. Not sure if any other component has it, so I removed it for consistency.
* dateinputs: timepicker, datepicker and datetimepicker internally hold a full dateinput component. In terms that component casts a shadow too. I completely disabled both "inner" border and shadow of nested dateinput. That's better visible in the bootstrap theme.
* dateinput, numeric and datetimepicker select background: originally, we changed the background of `.k-select` element. now we change only the background of `.k-link` background.
* active state of `.k-link` in `.k-select`: I switched the state to match pressed button. Unfortunately, it's very close to normal state. We'll discuss this with the designers

** Material **

* dateinputs: the icons are not getting the color. it used to, but now it doesn't.
* grid custom editor: same as above
* combobox / dropdown for jquery: neither angular, nor blazor, nor react, nor vue have the icon switching behaviour. I changed it to be consistent

** Bootsrap **
* combobox shadow -- see note about default theme
* dateinput, numeric and datetimepicker select background: see note about default theme.

inikolova:
@joneff I agree with most of the comments, but the following ones need further discussion:

Default
- combobox shadow: open combobox should not have shadow. Not sure if any other component has it, so I removed it for consistency - Actually opened combobox is focused - see the cursor https://www.telerik.com/kendo-angular-ui/components/dropdowns/combobox/. Also, AutoComplete, DropDownList, MultiSelect all get the same focused shadow when opened. So in my opinion, we should keep the combobox shadow in open state;
- active state of .k-link in .k-select: I switched the state to match pressed button. Unfortunately, it's very close to normal state. We'll discuss this with the designers - no visual indication for pressed increase/decrease button in NumericTextBox. Further more, it looks strange because there is hover effect.

Let me know what do you think for the above statemenets

joneff:
Only in kendo-jquery there is the concept of "opened" popups. Namely, there is `k-state-border-down` and `k-state-border-up` only in that suit. In practice, when kendo-jquery-comboxbo is opened, it's also focused. Put simply: the change will not affect any other suite, or jquery for that matter.

On the active state thing -- I will change the styles to use primary button variables for the time being.

Lastly, I will postpone this particular PR until after we release 2021 R1 SP1.

PR #2328, opened 2021-02-22
Use `$subtle-text` for calendar styling

joneff:
Part of #2234


inikolova:
@joneff, are these changes intended:

- calendar-range-angular.png -> **Today** text is now lowercase (previously is uppercase);
- calendar-infinite.png -> same as above;
- Calendar -> week numbers color and disabled text color is very close to the active dates (Material theme especially);

PreslavKozovski (approved):

joneff:
>     * calendar-range-angular.png -> **Today** text is now lowercase (previously is uppercase). In all Angular demos it is uppercase -> https://www.telerik.com/kendo-angular-ui/components/dateinputs/calendar/
>     * calendar-infinite.png -> same as above;

I changed the rendering, not the styling. If we want to always have uppercase in Material theme, perhaps we should add styling about that

>     * Calendar -> week numbers color and disabled text color is very close to the active dates (Material theme especially);

Indeed. However, the contrast ratio is better than it was before, especially in Material, where the colors were the same.

inikolova (approved):

PR #2334, opened 2021-02-22
Use `$subtle-text` for tabstrip styling

joneff:
Part of #2234

On a related note, the actual material tabstrip **does not** have border, but I guess is too big of a change to introduce now. We'll leave it for v5 of the themes.


inikolova:
Materia theme: 

- Missing hover state for tabs;
- Active text color is changed;
- Disabled color is very close to the base one.

joneff:
> * Missing hover state for tabs;

It's not missing. It's just very faint. I will see what we can do about that.

> * Active text color is changed;

As per design.

> * Disabled color is very close to the base one.
yep. i will see what I can do about that

PreslavKozovski:
I believe we should consult with the designers and see if we can remove the `k-spreadsheet-tabstrip` specific styles.

inikolova (approved):

PreslavKozovski (approved):

PR #2361, opened 2021-03-09
fix(dragdrop): add indicator container spacings for the KendoReact Grid

silviyaboteva:


PreslavKozovski (approved):
On a side note, an issue cannot be both "bug" and "regression" at the same time :)

joneff (approved):

PR #2373, opened 2021-03-15
chore(editor): rename table- to element-resize-handle telerik/kendo#12101

veselints:


joneff (approved):

joneff:
@veselints this ought to be a `fix`, instead of `chore` because we actually change something in the themes. But that's just nit picking.

Issue #2376, opened 2021-03-16
Enable interaction states on static markup


eeasss:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you would like to see implemented**

In Unite UX templates expose different components and their states. In order to achieve this in the infinite canvas we render components HTML and apply additional k-state-[stateName] classes in addition. For example: ‘k-button k-state-hover’ to show button in hover state.
 
However, it looks like some components do not have exposed such classes but use only pseudo classes (:hover, :active, etc) . Such are:
 
- [x] Checkbox - hover , focus
- [x] RadioButton - hover, focus
- [x] Calendar - header - data and today - hover, focus
- [x] Calendar - side navigation - hover
- [x] Calendar - cell td element - hover
- ~~Calendar - navigation - focus, hover and active state~~
- [x] DateInput - wrap element - hover state
- [x] DatePicker wrap element - hover state
- [x] DateRangePicker wrap element - hover state
- [x] MultiSelect - hover on close button
- [x] AutoComplete- hover on close button
- [x] ComboBox- hover on close button
- ~~Grid row hover~~
 
This list will go bigger when we start adding other components.
 
Kendo themes should be updated so that every time pseudo class is used, the corresponding k-state-[stateName] should be added. 

**Describe the alternatives you have considered**
There is no viable alternative.

**Additional context**
//cc @rkirilov @joneff @PreslavKozovski

PR #2379, opened 2021-03-17
feat: add interaction states for checkboxes

inikolova:
https://github.com/telerik/kendo-themes/issues/2376

### DIFF
--- a/packages/default/scss/checkbox/_layout.scss
+++ b/packages/default/scss/checkbox/_layout.scss
@@ -43,13 +43,15 @@
     }
 
     // Hover state
-    .k-checkbox:hover {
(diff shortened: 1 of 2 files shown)


packages/default/scss/checkbox/_layout.scss
```diff
@@ -43,21 +43,24 @@
     }
 
     // Hover state
-    .k-checkbox:hover {
+    .k-checkbox:hover,
+    .k-checkbox.k-state-hover {
```
joneff, line 47: The checkbox can always have a cursor poniter :) After all, we see it only when we hover it.

joneff (approved):

PR #2396, opened 2021-03-23
fix(grid): apply correct styles for filer menu that is part of the po…

silviyaboteva:
…pup content

This change is needed as the KendoReact GridColumnMenuCheckboxFilter renders the filter menu as a content inside the popup, not as a popup itself. See the demo here: https://www.telerik.com/kendo-react-ui/components/grid/filtering/#toc-column-menu-filter.

The following change applies the items' paddings and 'Select all' option bottom border.

### DIFF
--- a/packages/default/scss/grid/_layout.scss
+++ b/packages/default/scss/grid/_layout.scss
@@ -1265,7 +1265,12 @@
         .k-filter-menu-container {
             width: $grid-filter-menu-width;
         }
+    }
 
+    .k-filter-menu.k-popup,
+    .k-grid-filter-popup.k-popup,
+    .k-popup .k-filter-menu,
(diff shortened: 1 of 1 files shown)


PreslavKozovski (approved):
Having 4-5 different HTML renderings for the same components is hard to maintain. We need to do something about these Grid Filter/ColumnMenu popups.

silviyaboteva:
The thing is that in the React Grid, these components are separate modules that can be combined with others in a column menu. For example with sorting, column pinning etc. Check the demo here: https://www.telerik.com/kendo-react-ui/components/grid/columns/locked/#toc-locked-columns-with-column-menu.

joneff (approved):

Issue #2204, opened 2021-01-06
PanelBar Styles Revision


jivanova:
- Design issue: https://github.com/telerik/kendo-ux-private/issues/264
- Blazor issue: https://github.com/telerik/blazor/issues/1841
- Aha link: https://progresssoftware.aha.io/features/BLZR-116

- [ ] Revise the panelbar styles and rendering
      - Differences in angular and react rendering, along with inline styles added to the elements
- [ ] Apply changes from design revision, if any

**_Blazor release: 12th of May (Blazor Dev will start working on the PanelBar at the beginning of April)_**

PR #2397, opened 2021-03-23
Panelbar styles revision

JoomFX:
closes https://github.com/telerik/kendo-themes/issues/2204

The updated rendering can be seen in the visual tests, but here's a quick overview:

```
<ul class="k-panelbar">
    <li class="k-panelbar-header k-item k-level-0">
        <span class="k-link">
            <span class="k-panelbar-item-icon k-icon k-i-cog"></span>
            <span class="k-panelbar-item-text">First (Level 0)</span>
            <span class="k-panelbar-toggle k-panelbar-collapse k-icon k-i-arrow-chevron-up"></span>
        </span>
        <div class="k-panelbar-content k-content">
            <h4>Title in template</h4>
            <p>Multi line paragraph in template. Some more content to break onto multiple lines</p>
        </div>
    </li>
    <li class="k-panelbar-header k-item k-state-expanded k-level-0">
        <span class="k-link">
            <span class="k-panelbar-item-text">Second (Level 0)</span>
            <span class="k-panelbar-toggle k-panelbar-collapse k-icon k-i-arrow-chevron-up"></span>
        </span>
        <ul class="k-panelbar-group k-panel k-group">
            <li class="k-panelbar-item k-item k-level-1">
                <span class="k-link">
                    <span class="k-panelbar-item-text">Child (Level 1)</span>
                </span>
            </li>
            <li class="k-panelbar-item k-item k-state-expanded k-level-1">
                <span class="k-link">
                    <span class="k-panelbar-item-text">Child (Level 1)</span>
                    <span class="k-panelbar-toggle k-panelbar-collapse k-icon k-i-arrow-chevron-up"></span>
                </span>
                <ul class="k-panelbar-group k-panel k-group">
                    <li class="k-panelbar-item k-item k-level-2">
                        <span class="k-link">
                            <span class="k-panelbar-item-text">Child (Level 2)</span>
                        </span>
                    </li>
                </ul>
            </li>
        </ul>
    </li>
</ul>
```

Here's what's different in the rendering and needs to be updated in all suits:

**Adding more specific (scoped) classes**
- On the root level `k-item` element we add `k-panelbar-header` as well
- We add `k-panelbar-item-icon` class to the `k-icon` and `k-image` elements
- Currently the item title text is not wrapped in an HTML element. We wrap it in a `span` with the `k-panelbar-item-text` class
- We add the `k-panelbar-toggle` class to the `k-panelbar-collapse` and `k-panelbar-expand` elements
- We add `k-panelbar-content` class to the `k-content` element
- We add `k-panelbar-group` class to the `k-group` element
- We add `k-panelbar-item` class to all `k-item` elements (except the root ones, which we added the `k-panelbar-header` class to)

**Adding level/hierarchy class**
- On each `k-item` element we add `k-level-#` class. The root items are `k-level-0`, then `k-level-1` and so on. 

**Replacing the arrow icon**
- We replace `k-i-arrow-s` with `k-i-arrow-chevron-down` and `k-i-arrow-n` with `k-i-arrow-chevron-up`

**Removing the obsolete/legacy classes:**
- We remove all `k-widget` and `k-header` classes from the rendering


packages/default/scss/panelbar/_layout.scss
```diff
@@ -75,25 +81,73 @@
...
+            &.k-level-3 {
+                .k-link {
+                    padding-left: $panelbar-item-padding-x * 3;
+                }
+            }
+            &.k-level-4 {
+                .k-link {
+                    padding-left: $panelbar-item-padding-x * 4;
+                }
+            }
```
PreslavKozovski, line 100: How about using a variable that can be changed and a cycle, so developers that need more levels can easily change this and build the theme?

packages/material/scss/panelbar/_variables.scss
```diff
@@ -6,41 +6,41 @@ $panelbar-font-size: $font-size !default;
 $panelbar-line-height: (20 / 14) !default;
 $panelbar-border-width: 1px !default;
 
-$panelbar-header-padding-x: 24px !default;
-$panelbar-header-padding-y: 12px !default;
+$panelbar-header-padding-x: map-get( $spacing, 6 ) !default;;
```
PreslavKozovski, line 9:
```suggestion
$panelbar-header-padding-x: map-get( $spacing, 6 ) !default;
```

tests/visual/panelbar-angular.html
```diff
@@ -29,40 +29,40 @@
 
         <section>
             <kendo-panelbar class="k-panelbar">
-                <kendo-panelbar-item class="k-panelbar-header k-item">
+                <kendo-panelbar-item class="k-panelbar-header k-item k-level-0">
```
PreslavKozovski, line 32: I see no `k-level-0` & `k-level-1` related styles. Why do we need these classes in the rendering?

Also, shouldn't it be `k-panelbar-level-*` ?

PreslavKozovski (changes requested):

tests/visual/panelbar-angular.html
JoomFX: As discussed, we will keep the `k-level-0` and `k-level-1` classes in the rendering.

PreslavKozovski (approved):
The changes look okay.

What do you think of using these class-changes for making the HTML renderings in all flavors the same? This will help to remove the 4 (almost) identical tests. @JoomFX @joneff @inikolova @silviyaboteva @Juveniel

silviyaboteva:
I agree to have unified rendering across the flavors. But this has to be logged for each team and plan accordingly.
However, I need to ask, do these style updates break the current rendering?

JoomFX:
@PreslavKozovski Yep, this is one of the purposes of this PR -  to discuss, unify and update the rendering in all flavors. And this is the main reason to assign the @telerik/kendo-front-end group for review - because it concerns all of us.
The current rendering in the different flavors is pretty much the same. I mean, the only differences are:

- jQuery use `ul` and `li` while Angular and React use `div`
- There are a couple of additional `div` wrappers in Angular and React (just a `div`, without a class)
- React have two additional wrappers (`<div class="k-animation-container k-animation-container-relative">` and `<div class="k-child-animation-container">`)

The changes in the rendering are basically replacing the arrow icon with a new one and adding new classes, as described in my comment above.

@silviyaboteva Nope, the style changes do not break the current rendering.
And yes, once this PR is merged, everybody should log an issue in their team to update the component rendering as explained above.

inikolova (approved):

PR #2398, opened 2021-03-24
Grid sticky columns and rows

inikolova:


silviyaboteva (commented):
Do we need to wrap the sticky classes in kendo-grid tag (just for Angular) and duplicate the styles?
Or these are explicitly for Angular rendering?

inikolova:
I added new styles for Angular grid only, because in Angular we use a bit different rendering:
- There is no **k-master-row** class on each of the rows;
-  **k-grid-row-sticky** class is set to `<tr>` elements (in contrast to React where this class is set to `<td>` elements).

silviyaboteva (approved):

packages/bootstrap/scss/grid/_theme.scss
```diff
@@ -228,6 +228,80 @@
...
+        .k-state-selected.k-grid-row-sticky:hover td,
+        .k-state-selected.k-grid-row-sticky.k-state-hover td,
+        .k-state-selected.k-alt.k-grid-row-sticky:hover td,
+        .k-state-selected.k-alt.k-grid-row-sticky.k-state-hover td,
+        tr.k-state-selected.k-alt:hover .k-grid-content-sticky,
+        tr.k-state-selected.k-alt.k-state-hover .k-grid-content-sticky,
+        .k-grid-row-sticky:hover td.k-state-selected,
+        .k-grid-row-sticky.k-state-hover td.k-state-selected,
+        tr:hover .k-grid-content-sticky.k-state-selected,
+        tr.k-state-hover .k-grid-content-sticky.k-state-selected {
```
joneff, line 300: I think we can all agree that we **must** improve the rendering... The selectors shouldn't be that complicated...

I know why we do what we do, but this is getting a bit...

```scss
// that should all be root selectors
.k-grid-cell-sticky { ... }
.k-grid-alt-row .k-grid-cell-sticky { ... }

.k-grid-row:hover .k-grid-cell-sticky,
.k-grid-row.k-state-hover .k-grid-cell-sticky { ... }

.k-grid-row.k-state-selected .k-grid-cell-sticky { ... }

.k-grid-row.k-state-selected:hover .k-grid-cell-sticky,
.k-grid-row.k-state-selected.k-state-hover .k-grid-cell-sticky { ... }
```

@inikolova thoughts?

inikolova: I totally agree that selectors are too complicated (especially Selected + Hover), however they are grouped by states and follow the order of styles in the other suites. If we separate them as suggested, we should repeat the same rules multiple times. According to me, I prefer the huge selectors to duplication of styles

PreslavKozovski (approved):

PR #2409, opened 2021-03-30
Bottom nav link

inikolova:


PreslavKozovski (commented):
Could you please elaborate on what this PR is resolving :)?

Based on my tests, these styles do not apply any visual changes to the new HTML. Also, I am not а fan of using negative margins, and I believe we should use them only if there is no other solution (similar to the `!important` flag). However, as I cannot understand the problem that we need to resolve, I cannot provide suggestions on the implementation.

inikolova:
Sorry for not explaining the case :) The idea of this class is the same as the **k-drawer-link** one -> https://github.com/telerik/kendo-themes/blob/01e10c0f95c1c8120541c3d00300ffd1084712d2/packages/default/scss/drawer/_layout.scss#L137
This is a CSS class which clients set by themselves to the link elements in the ng-template when routing is used - > the idea is to expand the element to fit the whole Bottom nav item. The Drawer routing example illustrates this scenario: 
https://www.telerik.com/kendo-angular-ui/components/layout/drawer/routing/

joneff (approved):

silviyaboteva (approved):

PreslavKozovski (approved):

PR #2414, opened 2021-04-02
trigger image test in FF 87

PreslavKozovski:
https://www.mozilla.org/en-US/firefox/87.0/releasenotes/


joneff (approved):
There are four major changes:

* inputs have slightly different height
* textarea has visible resize handler
* support for `overflow-clip-box` has been removed 😞 
* tables in calendars are strangely sized

PR #2422, opened 2021-04-06
Rating readonly

silviyaboteva:
The idea of this change is that the read-only mode equals non-interactive mode (even on hover). And here are examples of a correct way of visualizing the read-only mode in the Rating component.

 - https://material-ui.com/components/rating/ 
 - https://bootstrap-vue.org/docs/components/form-rating Readonly (non-interactive) rating


packages/default/scss/rating/_layout.scss
```diff
@@ -13,6 +13,10 @@
         display: inline-flex;
         -webkit-touch-callout: none;
         -webkit-tap-highlight-color: $rgba-transparent;
+
+        &.k-state-readonly {
```
joneff, line 17: this should be `k-readonly`, but we can rename it another time

joneff (approved):

PR #2423, opened 2021-04-06
feat(drawer): hierarchical items

PreslavKozovski:
part of: https://github.com/telerik/kendo/issues/12111
design: https://github.com/telerik/kendo-ux-private/issues/292


joneff (approved):
The solution is OK. Perhaps at some point we might want to consider `.k-level-*` approach we had in panelbar

Issue #2227, opened 2021-01-12
Add styles for Task Board component


nenchef:
## Meta

<details>

* Aha item -- [KUI-505](https://progresssoftware.aha.io/features/KUI-505)
* Cross-team issue -- telerik/k2#311
* Design issue -- telerik/kendo-ux-private#281

</details>

**Design:**
* [Default theme](https://xd.adobe.com/view/5eaa6e2d-f4d8-4f77-a1df-b12c73907c4b-0bf9/)
* [Bootstrap theme](https://xd.adobe.com/view/39773fef-9238-4231-b029-8340c6c9e87d-6c24/)
* [Material theme](https://xd.adobe.com/view/5cd67fef-a825-4064-89fb-c97b7a4271eb-0985/)

**Assignee(s):**

@Juveniel 

## Markup

Component Layout ([test sample](https://dojo.telerik.com/oCUKIzUL/2)):

```html
<div class="k-taskboard">
    <!-- Toolbar -->
    <div class="k-taskboard-header">
        <div class="k-taskboard-toolbar k-toolbar">
            <button class="k-button k-button-icontext">
                <span class="k-icon k-i-plus"></span>
                Add Column
            </button>
            <span class="k-spacer"></span>
            <span class="k-taskboard-searchbox k-searchbox">
                <span class="k-input-prefix">
                    <span class="k-icon k-i-search"></span>
                </span>
                <input class="k-input" type="text" placeholder="Search...">
            </span>
        </div>
    </div>
    <!-- Content -->
    <div class="k-taskboard-content">
        <div class="k-taskboard-columns-container">
            <div class="k-taskboard-column">
                <div class="k-taskboard-column-header">
                    <div class="k-taskboard-column-header-text k-text-ellipsis">Column Header</div>
                    <span class="k-spacer"></span>
                    <div class="k-taskboard-column-header-actions">
                        <button class="k-button k-button-icon k-flat">
                            <i class="k-icon k-i-plus"></i>
                        </button>
                    </div>
                </div>
                <div class="k-taskboard-column-cards-container">
                    <div class="k-taskboard-column-cards">
                        <div class="k-taskboard-card k-card">
                            <div class="k-card-header k-hbox">
                                <a class="k-card-title k-link" href="#">Card Title</a>
                                <span class="k-spacer"></span>
                                <div class="k-card-header-actions">
                                    <button class="k-button k-flat">
                                        <span class="k-icon k-i-more-vertical"></span>
                                    </button>
[...]

PR #2425, opened 2021-04-07
TaskBoard fixes

Juveniel:
related to: #2227

This PR addresses two issues:

* setting overflow wrapping for card title to prevent the text from overflowing its line box 
* utilizing correct class for flat button in tests


tests/visual/taskboard-column.html
```diff
@@ -23,13 +23,13 @@
                                 <div class="k-taskboard-column-header-text k-text-ellipsis">Column Header</div>
                                 <span class="k-spacer"></span>
                                 <div class="k-taskboard-column-header-actions">
-                                    <button class="k-button k-button-icon k-button-flat">
```
joneff, line 26: both are fine. In fact, in the scss source files we use `.k-button.k-flat` extends `.k-button-flat`

joneff (approved):

Issue #2245, opened 2021-01-19
[TileLayout] The header of an overlapped item protrudes through the item on top


dtopalov:
**Describe the bug**
When two tiles are placed such that one overlaps the other, depending on the items' order, the title of the one below can be seen through the header section of the one on top.

The border between the heading and content section is also doubled, still the content of the item below is not visible (which is correct):

![image](https://user-images.githubusercontent.com/7533599/105037494-b8e34c00-5a66-11eb-8993-9cba22d0c90c.png)

**To reproduce**
https://stackblitz.com/edit/react-4vms2w-nuua6l?file=app/main.jsx

**Expected behavior**
When an item is entirely below another item, no parts of it should be visible through the item on top:

![image](https://user-images.githubusercontent.com/7533599/105038057-86861e80-5a67-11eb-934c-d52b3f01832a.png)

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for Angular
* Kendo UI for React

**Additional context**
The problem seems to be caused by a z-index of 1 applied to the **k-card-header** class. The outcome is correct when this z-index is removed.

PR #2429, opened 2021-04-12
fix(tilelayout): hide the header of an overlapped item

silviyaboteva:
Resolves https://github.com/telerik/kendo-themes/issues/2245


PreslavKozovski (changes requested):
You can add a test for this scenario. However, on a side note, the Angular implementation of the TileLayout is wrong. We should not allow the usage of specific grid-area possitions. We use the "order" property of this.

Having said that, if the JS implementations were correct, this scenario would not be possible to achieve.

JoomFX (approved):
Apart from Presli's comment, it looks good to me.

@PreslavKozovski The StackBlitz example shows the React TileLayout, not Angular.
Is this the implementation you are talking about?

silviyaboteva:
@PreslavKozovski we also have order property applied by default to each of the tiles. However, the case is hit when 2 tiles have the same `row` applied. In this case, both Tile 4 and Tile 1 are on the third row, and the first spans 3 columns.

If the `row` prop is not applied, this case cannot be met, as the Tile will move to the next available row.

joneff (approved):

joneff:
Just to clarify: while a css grid "cell" may indeed have order, start, end and span, setting all those will result inevitably in overlapping! That's the idea of the css grid -- that you can do that!

So each suite is doing their part to avoid said overlapping:
* in kendo-jquery, there is no specific row / col index. instead there is order and size (span)
* in kendo-react, we are using an external library 

I am not sure about kendo-angular, but I tested with explicit overlapping boxes and it's possible to overlap them initially.

Issue #2421, opened 2021-04-05
[Grid] Missing border between expand/collapse column and the first Grid column


dtopalov:
**Describe the bug**
In master-detail scenarios there is no border between the expand/collapse cell and the first Grid column cells. This looks particularly bad when the first Grid column is a checkbox column or contains icons in its template.

I suppose the idea behind this is to hide the border between the detail cell and the expand/collapse column in the detail row, which seems reasonable. However, the border should be present in the master row itself.

**To reproduce**

![image](https://user-images.githubusercontent.com/7533599/113559945-e160f800-960a-11eb-8453-ba778e972de9.png)

![image](https://user-images.githubusercontent.com/7533599/113560008-fc336c80-960a-11eb-94e7-e205746bd9a5.png)

https://stackblitz.com/edit/angular-ewx7hf?file=app/app.component.ts

**Expected behavior**
There should be a border between the expand/collapse column and the first Grid column in the Grid master rows, and no border in the detail rows, e.g.:

https://stackblitz.com/edit/angular-ewx7hf-htey38?file=app/app.component.ts

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**
* All
* Not Applicable
* Firefox version: [for example, 72, lts]
* Chrome version: [for example, 80]
* Edge version: [for example, 80]
* Safari version: [for example, 13]
* IE version: [for example, 11]

**Build system information (please remove the unneeded items)**
* Not Applicable
* OS: [for example, Windows, Linux, Mac]
* Node version: [for example, 10.15.3]
* NPM version: [for example, 6.4.1]

**Additional context**
Add any other context about the bug.

PR #2430, opened 2021-04-12
Expand collapse column border

inikolova:
Fixes https://github.com/telerik/kendo-themes/issues/2421


PreslavKozovski (commented):
This is not as per design - https://xd.adobe.com/view/ec9e5dd9-34b1-4c03-b779-ae109c8e554b-b9cd/

Also, the icons in the Angular grid are wrong.

inikolova:
This change has been discussed on the last FE+design sync (07.04) and (as far as I know) there is agreement to implement it.

silviyaboteva (approved):

joneff:
@PreslavKozovski indeed the change was discussed and it was agreed upon that the border must be added. And you are right about the icons.

PR #2435, opened 2021-04-15
fix: remove active styles

inikolova:
This fix is related to the case when the **select** event is prevented. 
https://dojo.telerik.com/ifEXaHig


silviyaboteva (commented):
Based on the designs both active and selected states should have a primary color: https://xd.adobe.com/view/8e1c8b07-28f9-43a6-a96d-21fc158cbc2d-db4f/

inikolova:
I checked with @joneff and @tyanko and we agreed there should not be primary color on active state

silviyaboteva (approved):
Ok. Then the designs should be updated

Juveniel (approved):

joneff (approved):

PR #2436, opened 2021-04-16
fix: add focus state for draghandle

inikolova:


joneff (approved):

joneff (commented):
Just to clarify: that's not a bug, because we are not changing existing code. Instead we are adding new one. That's why I changed the label from `bug` to `enhancement`

PR #2449, opened 2021-04-26
chore(scheduler): add spacing to year view calendar in safari and ie

Juveniel:
Fallback for [gap support in flex layout](https://caniuse.com/flexbox-gap).


Dimitar-Goshev:
The issue is fixed in IE 11. I cannot test it on Safari on my side.

joneff (approved):

joneff:
> The issue is fixed in IE 11. I cannot test it on Safari on my side.

Currently, both don't support flex gap. I am guessing with Safari 15 things will be better and we can remove those styles

PR #2466, opened 2021-05-07
Adds suite visual tests

tsvetomir:
* Adds proposed new style of visual tests for each product.
* Needs #2465 to be able to screenshot sub-folders.

/cc @ivanchev


ivanchev:
In order to delineate the new tests from the old ones better, as well as provide a clean structure that can be mirrored in the product projects (the ones that will use the actual components to generate the html's) I propose we move the new tests to a new folder, separate from the current `visual`. I also believe that the output folders can be separate between the two test suites, so we can do one of the following:
- move the current contents of the `visual` folder into a dedicated sub-folder with an appropriate name. Then create a second sub-folder for the new tests
- create a new folder as a sibling to `visual` - `tests/visual-by-team` (with a better name), and have the separation at that level.

tsvetomir:
I agree with the idea of splitting tests into folders. This would also allow us to have a proper README at each level.

It would be a bit easier from the scripting side to have a common root for all visual tests, e.g. "src":
```
tests/visual
├── assets
├── output
└── src
    ├── misc                                   <-- current test go here
    │   ├── action-buttons.html
    │   ├── action-sheet-position.html
    │   └── avatar-shape.html
    └── reference                              <-- reference rendering and per-suite snapshots go here
        ├── buttons
        │   ├── action-buttons.html            <-- additional test pages
        │   ├── angular                        <-- per-suite snapshots
        │   │   ├── action-buttons.html
        │   │   └── index.html
        │   └── index.html                     <-- main test page
        └── grid
            ├── angular
            │   └── index.html
            ├── blazor
            │   └── index.html
            ├── index.html
            └── react
                └── index.html
```

dtopuzov (approved):

joneff:
can the source HTML files be adjacent to each other and have the name of the component? It makes things much easier to look at.

* grid (singular)
  * grid.html (or grid-spec.html)
  * grid-angular.html
  * grid-react.html
  * grid-toolbar.html (or grid-toolbar-spec.html)
  * grid-toolbar-vue.html
  * ...

tsvetomir:
That's a good idea 👍🏽 I suggest we don't add "spec" suffix as it messes with the sort order.

* grid
  * grid.html
  * grid-angular.html
  * grid-react.html
  * grid-vue.html
  * grid-toolbar.html
  * grid-toolbar-angular.html
  * grid-toolbar-react.html
  * grid-toolbar-vue.html

ivanchev:
This will make keeping track of the files extremely difficult. IMO the folder separation between technologies is the best way to organize them.

tsvetomir:
Folders are a bit worse for tracking as they require you to navigate to see what's in them. In a single folder you can treat each scenario as a group and look for the suite name in it. You can compare modification dates and commit message for each file right away.

Larger components can be broken down by scenario to keep folder size reasonable.

I'd even go one step further and put the snapshot right next to the source file.

tsvetomir:
New folder structure is done, see https://github.com/telerik/kendo-themes/tree/new-tests-spike/tests/visual

tsvetomir:
Quick summary on performance:
* Improvements in the build process reduced test time from ~ 6 minutes to ~ 4 minutes.
* The additional tests account for ~ 20 seconds.
* We should review performance again when "CI | Visual" hits the 5 minute mark.

joneff (approved):

Issue #2388, opened 2021-03-19
PanelBar expand/collapse icons are next to the text instead of being at the end of the item. (IE 11)


simonssspirit:
**Describe the bug**

PanelBar expand/collapse icons are next to the text instead of being at the end of the item.

**To reproduce**

This can be seen in any demo, in this for example:

https://www.telerik.com/kendo-react-ui-develop/components/layout/panelbar/ (IE 11)

**Expected behavior**

The expand/collapse icon is on the right

**Screenshots**

![image](https://user-images.githubusercontent.com/11087921/111764556-fe2bca80-88ab-11eb-8fd5-0ebe63f1cbd0.png)

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**

* IE version: [for example, 11]

**Build system information (please remove the unneeded items)**
* Not Applicable

PR #2472, opened 2021-05-10
fix(panelbar): the arrow icons are not aligned correctly in IE

silviyaboteva:
fixes https://github.com/telerik/kendo-themes/issues/2388

The fix is related to the CSS properties [margin-inline-start](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-start) and [margin-inline-end](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-end) that are not supported by IE.

### DIFF
--- a/packages/default/scss/panelbar/_layout.scss
+++ b/packages/default/scss/panelbar/_layout.scss
@@ -185,3 +185,44 @@
     }
 
 }
+
+
+
+
+@include exports( "panelbar/layout/compatibility" ) {
+
+    .k-ie {
+
+        // Toggle icon
+        .k-panelbar-expand,
+        .k-panelbar-collapse,
+        .k-panelbar-toggle {
+            margin-left: auto;
+        }
+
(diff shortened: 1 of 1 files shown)


silviyaboteva:
hm, I'm not sure why these screenshots are produced. They seem to add some missing borders and vertically align some arrows in the calendars.

JoomFX (approved):

joneff (changes requested):
Use `.k-ie` to target IE.

We are close to removing IE support, so it will be much easier if our IE specific cases are targeted.

Juveniel:
@silviyaboteva just a suggestion - for simplicity and better separation we can create a module for the legacy styles (e.g. compatibility) [like here](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/taskboard/_layout.scss#L221-L223) for example.

joneff (approved):

PR #2492, opened 2021-05-19
fix: pack themes in CI

tsvetomir:
Missed that in #2465


tsvetomir:
Not actually needed as the themes are used in-place. Closing.

Issue #2500, opened 2021-05-25
Add focus styles for the Grid Pager


JoomFX:
**Is your feature request related to a problem? Please describe.**
The Grid Pager is a focusable element and should have focus styles (the same as the focused cell styles shown in the screenshot below).

![pager](https://user-images.githubusercontent.com/7753940/119507838-db200a00-bd77-11eb-8569-192265c9d8c9.png)

PR #2512, opened 2021-05-27
fix(grid): add pager focus styles

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/2500

### DIFF
--- a/packages/bootstrap/docs/customization-charts.md
+++ b/packages/bootstrap/docs/customization-charts.md
@@ -28,8 +28,8 @@ The following table lists the available variables for customization.
         <tr>
     <td>$series-a</td>
     <td></td>
-    <td>$teal</td>
-    <td></td>
+    <td><span class="color-preview" style="background-color: #0275d8"></span>#0275d8</td>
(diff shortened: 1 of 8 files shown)


silviyaboteva (approved):

Juveniel:
@JoomFX the Pager wrapper is not supposed to have focus styles according to [the component design](https://github.com/telerik/kendo-ux/blob/master/themes/default/PNG/pager.png). 

Also, the Pager wrapper itself is not a focusable element and does not have tabindex that allows this (e.g. Angular Grid). Taking this into consideration, I think we should not add focus styles to the wrapper of the Pager.

//cc @silviyaboteva

JoomFX:
@Juveniel @dtopalov is currently working on feature that requires the whole Pager element to be focusable.
He has discussed this with @joneff already and they have agreed that we need to add focus styles to the Pager element.
If you need more details, please discuss it further with @dtopalov .

Juveniel (approved):

joneff (approved):

joneff:
@JoomFX if the item that is closed is enhancement, the PR should definitely not be bug.

Issue #2461, opened 2021-05-06
Wrong appearance of the date input when editing a Grid row


JoomFX:
**Describe the bug**
The date input has additional `padding` and `border-bottom` when editing a Grid row.

**To reproduce**

1. Go to [this Stackblitz example](https://stackblitz.com/edit/angular-er3ccm?file=app/app.component.ts)
1. Click on the "Edit" button of a given row
1. Notice the styling of the date input in the "Date" column

**Expected behavior**
The date input should have correct styling.

**Screenshots**
![grid-editing](https://user-images.githubusercontent.com/7753940/117295254-49119980-ae7c-11eb-94bc-93c32b2907e2.png)

**Affected package (please remove the unneeded items)**
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**
* All

**Build system information (please remove the unneeded items)**
* Not Applicable

PR #2513, opened 2021-05-28
fix(grid): do not apply styles for nested dateinput

silviyaboteva:
fixes: https://github.com/telerik/kendo-themes/issues/2461


JoomFX (approved):
Great!
Just the `input` elements in the visual tests - they do not need a closing/forward slash :)

Juveniel (approved):

joneff (changes requested):
* There is no longer a test for daterange picker
* grid editing test was designed to show overlapping of error tooltip and calendar popup in a grid
* the datetime in the same test is missing the time icon

joneff (changes requested):
See material grid editing custom editor -- there is unexpected padding in datepicker, timepicker and datetime picker when in rtl

silviyaboteva:
The fix with the border-radius in separate one. So ther are 2 things in this PR:
1. The double bottom border that renders in the Material. It was due to the fact that Angular and React datepickers uses the dateinput and it comes with its rendering. I changed the test of the grid to catch this scenario.
1. After changing the grid test, there was another issue catched. In RTL mode, the pickers do not switch their border radius from left to right.

joneff (approved):

PR #2522, opened 2021-06-01
Reduce usage of @extend

joneff:
Part of #2551

~~Note: this will probably require some tweaking in the components~~ Components updated

Update: reduced only to calendar header buttons!


silviyaboteva (commented):
@joneff can you list the things that need to be changed in the components rendering?
Also, I've noticed additional borders in the tests, are they expected?

packages/material/scss/spreadsheet/_layout.scss
```diff
@@ -28,17 +18,6 @@
...
-        .k-button {
-            @extend .k-button-flat;
-            margin: 4px;
-            border-radius: 50%;
-
-            &::before {
-                display: block;
-            }
-        }
-
```
Juveniel, line 41: The following styles can also be removed, as currently they add additional complexity and are not needed - https://github.com/telerik/kendo-themes/blob/12c7b0534a0a22460408c0037dee162c2185094e/packages/default/scss/spreadsheet/_layout.scss#L239-L243

packages/material/scss/spreadsheet/_layout.scss
```diff
@@ -51,40 +30,3 @@
     }
 
 }
-
-
-
-
-@include exports( "spreadsheet/window/layout/material" ) {
-
-    .k-spreadsheet-window {
```
Juveniel, line 60: The `.k-spreadsheet-window` styles could use a further positioning tweak (could be releated to rendering also):

image

packages/default/scss/forms/_layout.scss
```diff
@@ -379,24 +379,6 @@
...
-    }
-    .k-edit-buttons .k-button + .k-button {
-        margin-left: ( $spacer-x / 2 );
-    }
-    .k-edit-form-container .k-edit-buttons {
-        margin: $window-inner-padding-y (-$window-inner-padding-x) (-$window-inner-padding-y);
-        padding: $window-inner-padding-y $window-inner-padding-x;
-    }
-
-
```
Juveniel, line 399: The `.k-edit-buttons` is currently used in Kendo jQuery Grid, TreeList, Gant. Scheduler and Editor components and this change also requires changing the markup in jQuery. I currently see that the Grid and Editor tests in the themes are updated, however, please make sure that the other component tests are also updated.

I also believe we should also log and link a separate issue for updating the rendering in jQuery.

PR #2524, opened 2021-06-01
docs(a11y): update Grid pager spec

dtopalov:
When we decided to add KB Navigation to the Kendo UI for Angular Grid pager, we performed a thorough research and discussions with @tsvetomir and @Raisolution that lead us to believe the following behavior is most suitable and appropriate:

- The navigable Grid (following the best practices for KB navigation of composite components) consists of separate single tab-stop sections - the pager (or pagers if both top and bottom pagers exist), and the Grid table (alongside its headers).
- Each section is a single tabstop meaning that the focus can be moved from one section to another with a single Tab press.
- The pager as a whole gets focused when tabbed into. The various scenarios include:
  - Tabbing from a focusable element before the Grid focuses the top pager wrapper (if such exists) or a Grid cell if there is no top pager
  - Tabbing from a focused Grid cell moves the focus to the bottom pager wrapper (if such exists) or outside of the Grid
  - Tabbing from the top pager wrapper focuses Grid cell
  - Shift-tabbing from the bottom pager focuses Grid cell
  - Pressing Enter when the Pager wrapper is focused moves the focus within the pager and focuses the first non-disabled focusable element
  - from there on the end user can tab through all focusable non-disabled pager elements sequentially
  - Pressing enter triggers the default action, associated with the currently focused element
  - Tabbing out of the last element and shift-tabbing from the first element brings the focus back to the whole pager
  - Pressing Escape while the focus is within the pager focuses the wrapper
  
The outlined behavior allows for consistent KB navigation UX regardless of the specific internal elements currently rendered in the Grid pager (subject to configuration, Grid width, etc.) It also allows the developer to provide consistent KB navigation when using the pager template with built-in or custom pager inner elements.

Further details are available in the following PR:

https://github.com/telerik/kendo-angular-grid/pull/1095#issue-656729744

// for visibility: @joneff @Raisolution

### DIFF
--- a/docs/accessibility/Grid.md
+++ b/docs/accessibility/Grid.md
@@ -78,15 +78,14 @@ Opening the popup editor focuses the first editor.
 | `Enter`| Triggers a submit action for the editor, including validation.|
 
 **Pager**
-To access the pager press `Arrow Down` when focus is on the last row in the grid.
+To access the pager press `Tab` when the focus is on a Grid cell.
 
 | Shortcut | Behavior |
 |----------|----------|
-| `Left/Right Arrow` | Moves focus to page buttons|
-| `Up Arrow` | Moves focus to first cell of last row in grid|
-| `Enter`| Selects the page|
-
-**Note:** After a discussion it was decided that selecting the go to first/last page of the pager will move the focus to the first/last number button in the pager (instead of focusing the first actionable button).
+| `Enter` | When the pager wrapper is focused, moves the focus to first focusable pager element|
+| `Tab` | When a pager element is focused, moves focus through all focusable pager elements sequentially. Tabbing out of the last element or shift-tabbing out of the first element moves the focus back to the pager wrapper.|
+| `Enter`| When a pager element is focused, triggers the default action, associated with the currently focused element.|
+| `Escape`| When a pager element is focused, brings the focus back to the pager wrapper. If a popup within the pager is currently opened, `Escape` will close the popup instead. Pressing `Escape` again brings the focus back to the pager wrapper.|
 
 **Header**
 


ivanchev:
I have several concerns with this approach:
1. when the pager is used as a standalone component, focusing it won't allow the user to do anything, until `Enter` is pressed. This is inconvenient, and also unclear as Enter is not an expected key for a "focus first focusable element" action.
2. it's inconsistent with the Grid itself. With the grid you focus directly to a cell, and can start using the arrow keys for navigation. For the pager you focus the wrapper, where no actions are available.
3. After entering the pager elements and tabbing to the last item, if focus is returned to the wrapper with the Tab key this violates the expectation that the next focusable element on the page will be focused.

A better way to achieve the separate single-tab-stop sections of the Grid would be:
1. Tab from a grid cell focuses the first focusable button in the Pager (the same as with the arrow down from last cell in the current approach.
2. once focus is in the pager, use the arrow keys for navigation.

This way tab and shift+tab will work as expected every time, always focusing the next/previous focusable element in regards to the pager.

dtopalov:
> I have several concerns with this approach:
> 
> 1. when the pager is used as a standalone component, focusing it won't allow the user to do anything, until `Enter` is pressed. This is inconvenient, and also unclear as Enter is not an expected key for a "focus first focusable element" action.
> 2. it's inconsistent with the Grid itself. With the grid you focus directly to a cell, and can start using the arrow keys for navigation. For the pager you focus the wrapper, where no actions are available.
> 3. After entering the pager elements and tabbing to the last item, if focus is returned to the wrapper with the Tab key this violates the expectation that the next focusable element on the page will be focused.
> 
> A better way to achieve the separate single-tab-stop sections of the Grid would be:
> 
> 1. Tab from a grid cell focuses the first focusable button in the Pager (the same as with the arrow down from last cell in the current approach.
> 2. once focus is in the pager, use the arrow keys for navigation.
> 
> This way tab and shift+tab will work as expected every time, always focusing the next/previous focusable element in regards to the pager.

The issue with using arrow keys for navigation through the pager elements that we encountered is that the pager may contain input or numeric textbox elements which need the arrow keys and combinations of arrow key and a modifier key like Shift for their regular purpose - moving the cursor through the input content or selecting the text. It will be really hard to distinguish when the arrow key needs to serve its default purpose within such elements or the focus needs to be moved to the next element.

We chose focusing the Pager itself as an additional step so that assistive technologies can announce its role, and additional useful information through aria-label e.g. "Pager, current page X of Y".

The main reason for choosing the Tab key for navigating through the pager was to keep the current flexibility related to what exactly the inner pager elements are, and rely on the default browser behavior for moving the focus from one element to the next regardless of the nature of all possible inner elements (which will also be beyond our control when a template is used).

I agree that pressing Enter to enter in the pager when its wrapper is focused may not be the obvious choice, so this can be changed with another button or button combo.

ivanchev:
Have you considered the option to have a wrapper to the input elements, and use Enter key to focus them - basically the approach now proposed for the Pager, but applied to the inputs inside of it. This would preserve the arrow navigation and single tab-stop pager, with the option to Enter into an input.

Stamo-Gochev:
Adding to the concerns in https://github.com/telerik/kendo-themes/pull/2524#issuecomment-852075805 I suggest discussing the following so that we are all on the same page:
1. Is it OK to allow multiple `Tab` actions inside the grid and thus allow a composite component to behave as it is divided into several sections/components (toolbar, header, content, pager, etc.) instead of thinking about it as a composite target?
2. Is it OK to allow focusing wrapper elements of a component that do not provide any underlying functionality instead of focusing the first actionable element inside the wrapper?

Here are some ideas on the above questions:
1. Suppose that a section of a page contains a Grid and the following elements:
- external input 1
- toolbar (**Note:** The toolbar implementation varies across suites and the individual tools might or might not be made focusable by a configuration option)
- header
- filter row
- content (it can contain a detail template with another grid inside)
- pager
- external input 2

Navigating from "external input 1" to "external input 2" might take 4-5 `Tab`s instead of 2 (one `Tab` to focus the Grid and one `Tab` to focus input2). 

Are we sure this good UX? 

Will all of the above Grid sections be focusable - and if some of them will not - is this intuitive to an end-user? 

For example, the end-user might visually distinguish between the different rows - toolbar row, the header columns row, and the filter row, but if some of them are not focusable, they might be confused why.

Adding to the above, if individual Pager elements are navigable with `Tab` and any of them is focused, it will take several more keystrokes to get out of it. I see [pressing `Escape`](https://github.com/telerik/kendo-themes/pull/2524#issuecomment-852087324) and moving the focus to the pager wrapper more of an escape hole and I have some doubts described in the following section 2)

2. Regarding @dtopalov's:
> We chose focusing the Pager itself as an additional step so that assistive technologies can announce its role, and additional useful information through aria-label e.g. "Pager, current page X of Y".

Can this be handled in a different way? What about adding the label to the first focusable element inside the pager instead, so that it is still announced by accessibility readers?

tsvetomir:
To summarize, the counterpoint is that the pager is [a kind of toolbar](https://www.w3.org/TR/wai-aria-practices/#toolbar).  This is a reasonable role for it, but it doesn't quite fit:
> Avoid including controls whose operation requires the pair of arrow keys used for toolbar navigation. If unavoidable, include only one such control and make it the last element in the toolbar. For example, in a horizontal toolbar, a textbox could be included as the last element.

While a proper toolbar can restrict the possible child elements, the pager cannot. We have to fall back to the more generic guidance for [§ 6.6 Keyboard Navigation Inside Components](https://www.w3.org/TR/wai-aria-practices/#kbd_general_within):
> As described in section § 6.1 Fundamental Keyboard Navigation Conventions, the tab sequence should include only one focusable element of a composite UI component. Once a composite contains focus, keys other than Tab and Shift + Tab enable the user to move focus among its focusable elements. **Authors are free to choose which keys move focus inside of a composite**, but they are strongly advised to use the same key bindings as similar components in common GUI operating systems as demonstrated in § 3. Design Patterns and Widgets. 

There's no good role that matches the pager and the arrow keys are reserved for child elements, so we've opted for the [mode used by the Grid](https://www.w3.org/TR/wai-aria-practices/#gridNav_inside). It fulfills the basic requirements of having a single tab stop and does not interfere with anything while the focus is inside it.

Some customers will still prefer to have each pager element as a separate tab stop. This is a reasonable customization option to have and might make sense to be the default for standalone pagers.

Arrow keys can still be used while the focus is on the pager to navigate to the next and previous page without moving the focus. This should cover a good deal of use cases without having to move the focus inside the pager. We can even extend it to include Home, End, PageUp and PageDown.

Whether the pager should be part of the grid is a good question. The argument for treating it as an external element is that it doesn't belong to any cell and has no relation to the headers. Same is true for the toolbar too. This implies that it should be a separate composite control and also a single tab stop.

Stamo-Gochev:
> To summarize, the counterpoint is that the pager is [a kind of toolbar](https://www.w3.org/TR/wai-aria-practices/#toolbar).

I didn't mean to look at the pager as a kind of a toolbar as they are conceptual differences between them and the WAI-ARIA guidelines have explicit descriptions for how a toolbar should work (but they do not have such things for the pager). 

My point was that I do not see why moving with `Tab` is more intuitive from a UX perspective compared to arrow keys and I see it more as an implementation decision rather than choosing the better UX. 

In addition, I think it is important to agree on what we think on the questions in https://github.com/telerik/kendo-themes/pull/2524#issuecomment-852114830 (and mainly on):
> 2. Is it OK to allow focusing wrapper elements of a component that do not provide any underlying functionality instead of focusing the first actionable element inside the wrapper?

tsvetomir:
> My point was that I do not see why moving with Tab is preferred from a UX perspective compared to arrow keys in the list-like scenario and I see it more as an implementation decision rather than choosing the better UX.

The `Tab` key is not active by default. The idea is to have a composite parent that you can `Enter` in and navigate it, much like a row in inline edit mode. Pressing `Escape` gets you back in the regular page tab order.

> Is it OK to allow focusing wrapper elements of a component that do not provide any underlying functionality instead of focusing the first actionable element inside the wrapper?

In addition to the `aria-label`, there could be top-level shortcuts, for example `Page Up` / `Arrow Left`, `Page Down` / `Arrow Right`, `Home` and `End` that perform common actions without the need to focus the respective inner element.

Raisolution:
> Tabbing out of the last element and shift-tabbing from the first element brings the focus back to the whole pager

I think the more natural behavior when tabbing out of the last element of the Pager would be to just follow the tab order sequence and actually `tab out` of it i.e. focus the next focusable element on the page instead of bringing the focus back to the whole Pager element.

dtopalov:
Having in mind the latest comments and several offline discussions, the functionality in the KUI for Angular Grid PR was updated as follows:

- When the Pager is focused, Left and Right Arrows, Page Up/Down and Home/End load the respective Grid page (prev/next or first/last).
- When the focus is on an internal pager element, focus is trapped, and tabbing/shift-tabbing rotates the focus between all pager elements; Escape brings the focus back to the wrapper.

Juveniel:
Closing in favor of https://github.com/telerik/web-components-ux/issues/424

PR #2527, opened 2021-06-03
feat: allow setting browser scrollbar width via CSS variable

silviyaboteva:
In KendoReact we have a discussion to apply scrollbar width paddings in the grid header and footer elements via CSS variable.

For reference: https://github.com/telerik/kendo-react-private/pull/1386


joneff (commented):
It's not that I am against custom properties (css variables), it's just that they are not as supported.

@silviyaboteva Have you tested the outcome in IE 11 on Windows?

Lastly, if we are relying on custom properties, the browser bar is high enough that we can safely use `padding-inline-start` and other logical properties instead of explicit rtl handling.

Xizario:
If we target Chrome on Mac, I think we don't need to calculate it at all
```css
    .k-grid-header.k-grid-draggable-header {
      padding: 0px !important;
      overflow-y: scroll;
    }
```
Looks okay on my end, and works on browser zooming without need of any JS.

silviyaboteva:
@joneff yes, you are right about the current support of the CSS variables ([not supported in IE](https://caniuse.com/?search=%20CSS%20Variables%20(Custom%20Properties)). Supported in all others popular browsers), but we have made sure that it will fall back to the second argument of the var() function, e.g. `$default-scrollbar-width` that was used previously.

For example, in MacOS with `Show scrollbars option: When Scrolling`, there would be no padding set at all as the scrollbar has an absolute positioning, as @Xizario said.
When the option is changed to `Show scrollbars option: Always`, the scrollbar has 15px width.
On Windows machines, the width is 17px. The same as `$default-scrollbar-width`. So, in IE on Windows, it should apply 17px padding as desired.

I like the idea of the `padding-inline-start` instead of targeting RTL. I'll change that.

JoomFX (approved):

PR #2532, opened 2021-06-07
fix(grid): do not apply pager focus styles on non-navigable grid

JoomFX:
related to https://github.com/telerik/kendo-themes/issues/2500

The [Grid keyboard navigation](https://www.telerik.com/kendo-angular-ui/components/grid/keyboard-navigation/) is disabled by default.
If this is the case, no `focus` styles should be applied.

### DIFF
--- a/packages/default/scss/grid/_theme.scss
+++ b/packages/default/scss/grid/_theme.scss
@@ -94,8 +94,7 @@
         .k-grouping-row > td:focus,
         .k-detail-row > td:focus,
         .k-group-footer > td:focus,
-        .k-grid-pager.k-state-focused,
-        .k-grid-pager:focus {
+        .k-grid-pager.k-state-focused {
             box-shadow: $grid-focused-shadow;
         }
 


silviyaboteva:
@JoomFX, can't you prevent the focus (or adding a `k-state-focus` class) from the component implementation instead?

JoomFX:
@silviyaboteva yes, this might be an option if the other suites do not need it.
I would have to remove the `.k-grid-pager:focus` line and leave just the `.k-grid-pager.k-state-focused`. And then, depending on the `navigable` input property to assign the `.k-state-focused` class.

So I will need an input from @joneff and @Juveniel here as well.

P.S. I also do not like to use `:not()` but there is nothing else I could hook on.

JoomFX:
@silviyaboteva actually, you are right :)
I will remove the `.k-grid-pager:focus` rule and will handle the rest (adding the `.k-state-focused`) from the component itself.

silviyaboteva (approved):

joneff (approved):

PR #2543, opened 2021-06-11
Styles for emulating justify content stretch

joneff:
not supported by flexbox, so relying on a bit of mischief


jivanova (approved):
Thank you!

PR #2567, opened 2021-06-20
Bootstrap var fixes

joneff:
* use `base`,  `hover` and `selected` variables for treeview
* improve treeview item spacing
* link header variables to base variables
* streamline card variables
* other tweaks

Note: needed to handle themebuilder related issues with bootstrap


JoomFX (approved):

silviyaboteva (commented):
I have only one concern regarding the fix of the find and relace dialog in the Editor. The produced output doesn't match the designs. So, we can either sync with the designers to update the design or revert them.

silviyaboteva (approved):

Issue #2560, opened 2021-06-18
ColorGradient RTL style adjustment


PaskalevStoyan:
**Describe the bug**
Wrong style is applied to the outer slider in the ColorGradient component when in RTL mode.

**To reproduce**

1. Go to [this link](https://www.telerik.com/kendo-angular-ui-develop/components/inputs/globalization/#toc-right-to-left-support).
1. See "Right-To-Left Support" example

**Expected behavior**
ColorGradient should have the correct styles in RTL mode.

**Screenshots**
![image](https://user-images.githubusercontent.com/49073752/122568081-100b3e00-d052-11eb-9eca-1e936b655dbd.png)

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material
* theme-tasks

**Affected suites (please remove the unneeded items)**
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**
* All

**Build system information (please remove the unneeded items)**
* Not Applicable

PR #2568, opened 2021-06-21
fix(colorpicker): wrong rtl style in color-gradient 

PaskalevStoyan:
fixes https://github.com/telerik/kendo-themes/issues/2560


JoomFX (approved):

packages/default/scss/colorpicker/_layout.scss
```diff
@@ -392,12 +392,10 @@
...
             text-align: center;
         }
-    }
 
-    .k-rtl,
-    [dir="rtl"] {
-        .k-flatcolorpicker {
 
+        .k-rtl &,
+        &[dir="rtl"] {
```
silviyaboteva, line 398: The `dir` attribute can also be set at the root of the page, not directly to the element with class `k-flatcolorpicker`. So, I suggest adding a third selector here:
```
.k-rtl &,
[dir="rtl"]&,
[dir="rtl"] & { ... }
```

For reference, check badge styles: https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/badge/_layout.scss#L157

silviyaboteva (changes requested):
It is good practice to check other Kendo flavors for the same issue. In this case, only Angular reproduces this issue. In KendoReact the RTL styles are applied correctly: https://stackblitz.com/edit/react-wcv2md-sbuybp?file=app%2Fmain.jsx

You can add a tag FW:Angular :)

joneff (approved):
As a note, the visual distortion comes from the fact that previously, there was a `k-rtl` class on the container and there isnt (because angular applies it via attribute).

PR #2577, opened 2021-06-29
fix(taskboard): pane spacing is not consistent with action buttons

Juveniel:
extracted as a follow-up task from https://github.com/telerik/kendo-themes/pull/2552

Streamlines padding for the TaskBoard pane to be consistent with the new actions styles.


joneff:
@Juveniel for `calc` reasons we are trying not to use `0` for variables that may end up there... like padding. When appropriate, either use `null` or `0px`.

joneff (approved):

PR #2593, opened 2021-07-08
Fix scrollview nav dots

PaskalevStoyan:


packages/default/scss/scrollview/_layout.scss
```diff
@@ -47,29 +47,6 @@
         }
     }
 
-    kendo-scrollview.k-scrollview-wrap.k-scrollview-multidot .k-scrollview-pageable > .k-button,
-    kendo-scrollview.k-scrollview.k-scrollview-multidot .k-scrollview-pageable > .k-button,
-    .k-widget.k-scrollview.k-scrollview-multidot .k-scrollview-nav > .k-link {
```
silviyaboteva, line 52: Check if removing these styles will break the jQuery widget.

tests/visual/src/misc/colorpicker-rtl.html
```diff
@@ -60,7 +60,7 @@
                     </div>
                 </div>
             </section>
-            <section class="k-rtl">
```
silviyaboteva, line 63: Is this the fix for the ColorPicker that we sync before? If yes, then please cherry-pick it to the `fix-color-gradient-rtl` branch.

packages/default/scss/colorpicker/_layout.scss
```diff
@@ -392,12 +392,11 @@
             margin-top: ($colorpicker-margin * 2);
             text-align: center;
         }
-    }
 
-    .k-rtl,
```
silviyaboteva, line 397: This commit also should be in `fix-color-gradient-rtl` branch.

packages/default/scss/scrollview/_layout.scss
PaskalevStoyan: I couldn't find the same class pattern on the jQuery website when inspecting their scrollView demos.

silviyaboteva (approved):
Great work on this! 🚀

joneff (approved):

PR #2594, opened 2021-07-08
Introduce grid edit cell variables

joneff:
* introduce grid edit cell variables
* remove customization from grid inputs

The most visible area is Material theme. However, this will be solved when we introduce fill mode for inputs or roughly 2022 R1


silviyaboteva (commented):
Vanka are these spacing changes expected: 
example: tests/visual/output/bootstrap/reference/grid/grid-editing-inline-angular.png

silviyaboteva (approved):

Issue #2498, opened 2021-05-25
PivotGrid Styles Revision


silviyaboteva:
## Meta

<details>

* Aha item (React) -- [KUIR-224](https://progresssoftware.aha.io/features/KUIR-224)
* Cross-team issue -- https://github.com/telerik/k2/issues/288
* Design issue -- https://github.com/telerik/kendo-ux-private/issues/232

</details>

**Design:**
* [Default theme](https://xd.adobe.com/view/569db2c6-fcaa-414a-8d7d-b71cc210cd48-03ce/)
* [Bootstrap theme](https://xd.adobe.com/view/4cf9323f-1e6e-4a8c-bca9-1deb7d205b02-19aa/)
* [Material theme](https://xd.adobe.com/view/97ba673d-5d8f-4642-bfcd-33f3e667914c-c2e9/)

**Assignee(s):**

@silviyaboteva and @kspeyanski 

## Markup
in tests/visual/src/misc/pivotgrid-*.html

## Remarks: 
 - The Pivotgrid and the Configurator are separate components/widgets, the position of the Configurator in relation to the table. For example, if we want the Configuratior on the Right (end) position, we can wrap the two components in a div with the following classes: `k-d-flex k-flex-row k-pos-relative`. The rest of the positions can be set by replacing `k-flex-row` with `k-flex-row-reverse | k-flex-col | k-flex-col-reverse`.
 - The position will be determined by the wrapper element class, mentioned above.
 - Specific classes should be used to distinguish push/overlay modes - `k-pivotgrid-configurator-push` and `k-pivotgrid-configurator-overlay`.
 - Specific classes should be used to distinguish horizontal/vertical orientation of the content - `k-pivotgrid-configurator-vertical` and `k-pivotgrid-configurator-horizontal`.   
 - The height of the `k-pivotgrid-column-headers` table should be dynamically calculated and assigned via `grid-template-rows` CSS property to the element with class `k-pivotgrid` .
 - The same applies for the row headers table. It's width should be dynamically calculated and assigned via `grid-template-columns` CSS property to the element with class `k-pivotgrid`. 
 > Note: Check the approach in `tests/visual/src/misc/pivotgrid.html` visual test.
 - Kendo jQuery specific: As they still support IE11 browser, the grid display of the Pivotgrid is not an option for them. So, we handle flex display behind the `k-ie11` class. In this case, the 4 child elements must be wrapped in pairs like this:
```html
<div class="k-pivotgrid">
  <div>
[...]

PR #2635, opened 2021-07-20
Pivotgrid revision

silviyaboteva:
Closes: https://github.com/telerik/kendo-themes/issues/2498

There are two things that are discussed with the design team and they will be updated in the designs later:
- Bootstrap theme: The checkboxes in the Treeview component have a white background in an unchecked state.
- The darker border for the total column will be removed
- The darker border separating the different fields in the row/column headers will also be removed.

Screenshot for the 
Screenshot 2021-07-19 at 14 16 04

cc/ @ag-petrov @kspeyanski @nstoychev


packages/default/scss/pivotgrid/_layout.scss
```diff
@@ -1,4 +1,453 @@
...
+        }
+
+        .k-treeview {
+            padding: 0 map-get($spacing, 2);
+            overflow: auto;
+        }
+    }
+
+    .k-pivotgrid-configurator-actions {
+        margin: (-$pivotgrid-configurator-actions-padding-y) (-$pivotgrid-configurator-actions-padding-x);
```
Juveniel, line 295: we can remove this negative margin and instead utilize the `k-actions k-hstack k-justify-content-end` classes for the container. See https://github.com/telerik/kendo-themes/commit/9ff8244e9d17b6b473990b00d680548d67248f82 for reference.  

This will also make the spacing in the container consistent, as the currently, the actions wrapper have padding and receive additional padding from the parent container.

packages/material/scss/pivotgrid/_theme.scss
```diff
@@ -1,4 +1,19 @@
-@include exports("pivotgrid/theme") {
+@import "~@progress/kendo-theme-default/scss/pivotgrid/_theme.scss";
+
+@include exports("pivotgrid/theme/material") {
+    // Hover and Focus state have the same bg color
```
Juveniel, line 4: Add spacing rows here for consistency

Juveniel (changes requested):

Juveniel:
@silviyaboteva I have also noticed that 

1) After running PostCSS Sorting in the _layout file, there are some changes in the properties order (not all of which seem legit). You could check on your end and apply the sorting where possible,

2) I have also noticed that in the configurator container there is no padding, thus when focusing inputs/buttons the top outline will not be visible. We should probably discuss this with the design team and adjust the spacing:

image

packages/default/scss/pivotgrid/_layout.scss
```diff
@@ -1,4 +1,453 @@
...
+        }
+    }
+
+
+    // Modes
+    .k-pivotgrid-configurator-push {
+        .k-pivotgrid-configurator-vertical {
+            min-width: $pivotgrid-configurator-vertical-width;
+        }
+    }
```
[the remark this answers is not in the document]
silviyaboteva: This, in fact, is a leftover. I'll remove it.

packages/default/scss/pivotgrid/_layout.scss
silviyaboteva: Thank you for noticing this. I changed the action buttons in the column menus, but forget to do the same for configurator itself.

silviyaboteva:
Hey @Juveniel 
Thank you for the extensive review. I addressed the comments in the code. About the general comment:
1. I updated the properties order in all files as well.
2. I synced with @ivelina-penkova and we agreed to add 2px (for Default theme) and 0.25rem (for Bootstrap theme) to the `$pivotgrid-configurator-content-padding-y` variable in order to prevent cutting the focus shadow of the inputs/buttons. Other themes don't apply top shadow.

packages/bootstrap/scss/pivotgrid/_layout.scss
```diff
@@ -1 +1,9 @@
 @import "~@progress/kendo-theme-default/scss/pivotgrid/_layout.scss";
+
+@include exports ("pivotgrid/layout/bootstrap/legacy") {
+
+    .k-pivot-toolbar {
+        padding: $toolbar-padding-y;
```
joneff, line 6: if we want to override a variable, we create a local one. In this case, two local ones: `$pivotgrid-toolbar-padding-x` and `$pivotgrid-toolbar-padding-y` or we don't customize the padding at all

packages/bootstrap/scss/pivotgrid/_theme.scss
```diff
@@ -1,4 +1,6 @@
-@include exports("pivotgrid/theme") {
+@import "~@progress/kendo-theme-default/scss/pivotgrid/_theme.scss";
+
+@include exports("pivotgrid/theme/legacy") {
```
joneff, line 3: why are we keeping the legacy styling?

packages/default/scss/pivotgrid/_theme.scss
```diff
@@ -1,5 +1,170 @@
 @include exports("pivotgrid/theme") {
 
+    .k-pivotgrid,
+    .k-pivotgrid-values {
```
joneff, line 4: no need for the second selector.

packages/default/scss/pivotgrid/_theme.scss
```diff
@@ -1,5 +1,170 @@
...
+            $bg: $pivotgrid-headers-bg
+        );
+    }
+
+    .k-pivotgrid-column-headers,
+    .k-pivotgrid-row-headers {
+        @include fill (
+            $border: $pivotgrid-headers-border
+        );
+    }
```
joneff, line 26: consider combining the two selectors. you are setting empty cell border in the next line, so it will override the previous setting

packages/default/scss/treeview/_layout.scss
```diff
@@ -98,6 +98,10 @@
         .k-i-collapse {
             margin-left: -$treeview-indent;
             cursor: pointer;
+
+            + .k-checkbox-wrapper {
```
joneff, line 102: consider extracting this as a separate fix

joneff (changes requested):
Overall I like the rendering. I am still thinking on the styles

packages/bootstrap/scss/pivotgrid/_layout.scss
silviyaboteva: Vanka, this line of code is inside the legacy styles of the Pivotgrid. I didn't touch the styles there, only place them under `/legacy` module. In the current rendering, there is no toolbar at all.

packages/bootstrap/scss/pivotgrid/_theme.scss
silviyaboteva: Due to the fact that jQuery still uses it and if we merge the PR before they change their rendering with the new one, it will broke their widget. Once they apply the new rendering, we can safely remove the legacy modules.

silviyaboteva: Also, they will have both renderings for a while. So, we need the styles for now.

packages/bootstrap/scss/pivotgrid/_layout.scss
silviyaboteva: I noticed that in `https://github.com/telerik/kendo-themes/pull/2606`, you've removed these pivot toolbar styles. It might be my mistake when I rebased. I'll erase them.

Juveniel (approved):

joneff (approved):

Issue #2551, opened 2021-06-17
Remove @extend usage 


joneff:
**TLDR**

Using `@extend` directive allows us to not write styles twice. However it causes issues, like dist bloat and troubles with "scoped" themes.

**Longer explanation**

`@extend` will append the current selector to the extended one in all instances. So while `@extend .k-button` may read like "extend the instances of pure `k-button`", it actually means "extend all instances of `k-button` regardless if it's standalone, parent or child selector"... And that results in some pretty wild sectors in the dist.

On top of that, when scoping our theme (importing in a class name) things begin to behave strange and it is possible that the build will fail. In all fairness `@extend` is not the only one to blame here, as we've engineered the themes that way.

**To sum up**

Removing `@extend` usage will allow us precise control over the output, reduce the unintended side effects and (with some additional tweaking) will allow customers to scope themes.

**Tasks**

* [x] #2495
* [x] #2552
* [x] #2561
* [x] #2522
* [x] telerik/kendo-angular-dateinputs#428
* [x] #2602
* [x] telerik/blazor#2659
* [x] telerik/blazor#2766
* [x] telerik/kendo-react-private#1445
* [x] telerik/kendo-angular-grid#1127
* [x] #2637
* [x] #2639

joneff:
As of now, there are 5 components in which `@extend` is still present:

* actions
* buttons
* pager
* scheduler
* dropdowntree

All but dropdowntree can be handled within the next release. Closing this one.

PR #2639, opened 2021-07-21
Remove extend from dialog and forms

joneff:
Part of #2551


mohd-akram:
It seems like removing `.k-plus` and `.k-minus` was a breaking change as Kendo React 3.x was using them in the grid detail, so they stopped showing.

Issue #2535, opened 2021-06-08
OrgChart Component


nenchef:
## Meta

* Aha item -- [KUI-415](https://progresssoftware.aha.io/features/KUI-415)
* Design issue -- https://github.com/telerik/kendo-ux-private/issues/333

**Design:**
* [Default theme](https://www.figma.com/file/gWUIux95BHxemqWwOw9ZwB/Default-Theme---Internal-Usage?node-id=162%3A24081)
* [Bootstrap theme](https://www.figma.com/file/k8eCVo9JZeHZlV9o8SXZtd/Bootstrap-Theme---Internal-Usage?node-id=56%3A0)
* [Material theme](https://www.figma.com/file/qpaRti273eNVEEOYOUxojT/Material-Theme---Internal-Usage?node-id=59%3A36324)

**Assignee(s):**

@Juveniel 

## Markup

```html 
<div class="k-orgchart">
    <div class="k-orgchart-container">

        <!-- No Grouping Vertical -->
        <div class="k-orgchart-group k-orgchart-group-v k-vstack k-justify-content-center">
	    <div class="k-orgchart-node-container k-vstack">
	        <div class="k-orgchart-node k-vstack k-align-items-center">

	            <!-- Card -->
		    <div class="k-orgchart-card k-card">
		        <div class="k-card-body k-hbox">
		            <div class="k-avatar-circle k-avatar k-avatar-md">
			        <div class="k-avatar-image"></div>
			    </div>
			    <div class="k-vbox">
			        <div class="k-card-title">Card Title</div>
                                <span class="k-spacer"></span>
			        <div class="k-card-subtitle">Card, Subtitle</div>
			    </div>
			    <span class="k-spacer"></span>
			    <div class="k-card-body-actions">
			        <button class="k-button k-flat k-button-icon">
			            <span class="k-icon k-i-more-vertical"></span>
			        </button>
			    </div>
		        </div>
		    </div>

                    <!-- Horizontal/Vertical Connections -->
	            <div class="k-orgchart-line k-orgchart-line-v"></div>

  		    <!-- Expand/Collapse Button -->
		    <button class="k-orgchart-button k-button k-button-icon">
		        <span class="k-icon k-i-minus"></span>
		    </button>

	        </div>
	        <div class="k-orgchart-node k-vstack k-align-items-center">...</div>
	    </div>
        </div>

        <!-- No Grouping Horizontal-->
[...]

PR #2648, opened 2021-07-23
Orgchart styles

Juveniel:
targets: https://github.com/telerik/kendo-themes/issues/2535
design: https://github.com/telerik/kendo-ux-private/issues/333


packages/default/scss/orgchart/_layout.scss
```diff
@@ -0,0 +1,92 @@
...
+    // OrgChart
+    .k-orgchart {
+        width: 100%;
+        box-sizing: border-box;
+        font-size: $orgchart-font-size;
+        font-family: $orgchart-font-family;
+        line-height: $orgchart-line-height;
+        display: flex;
+        justify-content: center;
+        overflow: auto;
```
joneff, line 12: if everything is in the orgchart container and everything is positioned, there is no real need for the element root element to be flex (block is fine) or have overflow.

packages/default/scss/orgchart/_layout.scss
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    .k-orgchart-container {
+        position: relative;
```
joneff, line 16: if this is the "canvas" of the orgchart, then it should have `width` and `height` set to 100%... Not that the width is not going to be 100%, considering this is a block element.

packages/default/scss/orgchart/_theme.scss
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    .k-orgchart-node-group-subtitle {
+        @include fill( $color: $orgchart-node-group-subtitle-text );
+    }
+
+    // Lines
+    .k-orgchart-line-h,
+    .k-orgchart-line-v {
+        @include fill( $bg: $orgchart-line-bg );
```
joneff, line 36: you can use `color` and make the background `currentColor`

packages/default/scss/orgchart/_layout.scss
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    // Nodes
+    .k-orgchart-group {
+        gap: $orgchart-group-gap;
+    }
+
+    .k-orgchart-group.k-vstack > .k-orgchart-node-container {
+        gap: 0;
```
joneff, line 25: why does the vertical group differ? what's so special about it?

packages/default/scss/orgchart/_layout.scss
```diff
@@ -0,0 +1,92 @@
...
+        font-size: $orgchart-node-group-subtitle-font-size;
+    }
+
+    // Card
+    .k-orgchart-card {
+        width: $orgchart-card-width;
+        border-width: $orgchart-card-border-width;
+
+        .k-card-title {
+            margin-bottom: $orgchart-card-title-margin-bottom;
```
joneff, line 56: on line 40 and 46 you use a different way of setting bottom margin.

packages/default/scss/orgchart/_layout.scss
```diff
@@ -0,0 +1,92 @@
...
+
+        .k-card-body {
+            border-width: $orgchart-card-body-border-width;
+            border-style: solid;
+            flex-grow: 0;
+        }
+        .k-card-body .k-vbox {
+            margin: 0 $orgchart-card-body-vbox-margin-right 0 0;
+            min-width: 0;
+            min-height: $orgchart-card-body-vbox-min-height;
```
joneff, line 73: not a good idea. some one might stack several `k-box` ... btw, use `k-vstack`, like you do above!

joneff (changes requested):

packages/default/scss/orgchart/_layout.scss
Juveniel: The container width and height are dynamically calculated by the component and is centered so that expanding/collapsing nodes does not create a 'jump effect'.

The overflow is needed for creating scroll when the chart is expanded past the set width.

packages/default/scss/orgchart/_layout.scss
Juveniel: The nodes in the vertical groups are connected with lines, thus having gap between them is not applicable.

joneff (approved):

packages/default/scss/orgchart/_layout.scss
Juveniel: Flex styles removed from `k-orgchart` as they were indeed preventing the `k-orgchart-container` element to expand as needed.

joneff (approved):

Issue #2622, opened 2021-07-19
Popover Component


Juveniel:
## Meta

* Aha item -- [KUI-551](https://progresssoftware.aha.io/features/KUI-551)
* Design issue -- https://github.com/telerik/kendo-ux-private/issues/335

**Design:**
* [Default theme](https://www.figma.com/file/gWUIux95BHxemqWwOw9ZwB/Default-Theme---Internal-Usage?node-id=380%3A42303)
* [Bootstrap theme](https://www.figma.com/file/k8eCVo9JZeHZlV9o8SXZtd/Bootstrap-Theme---Internal-Usage?node-id=120%3A27857)
* [Material theme](https://www.figma.com/file/qpaRti273eNVEEOYOUxojT/Material-Theme---Internal-Usage?node-id=106%3A45227)

**Assignee(s):**

@Juveniel 

## Markup

```html
 <div class="k-popover">
      <div class="k-popover-callout"></div>   
      <div class="k-popover-header">Title</div>
      <div class="k-popover-body">Body</div>
      <div class="k-popover-actions k-actions k-hstack k-justify-content-between">Actions</div>     
  </div>
```

Note: After further discussion with @joneff we decided that a more flexible approach would be to have simplified structure, where the header, content and actions will inherit from the Card variables instead of using the k-card specific classes (reflected in the above rendering).

/cc @ag-petrov

PR #2672, opened 2021-07-29
Popover styles

Juveniel:
targets: https://github.com/telerik/kendo-themes/issues/2622
design: https://github.com/telerik/kendo-ux-private/issues/335

We should consider extracting the `callout` in a common module that is reused in Tooltip, Popover, Card, etc. This, however, is not possible with the current PR, as the classes are tightly coupled in several places and this will lead to breaking changes. We could address this after the September release.


packages/default/scss/popover/_layout.scss
```diff
@@ -0,0 +1,82 @@
...
+    .k-popover-callout {
+        margin: 0;
+        width: $popover-callout-width;
+        height: $popover-callout-height;
+        border-width: $popover-callout-border-width $popover-callout-border-width 0 0;
+        border-style: $popover-callout-border-style;
+        position: absolute;
+    }
+
+    .k-popover-callout.k-callout-n {
```
JoomFX, line 58: It might be better (and more consistent) if we nest the `k-callout-n/e/s/w` classes in `k-popover-callout`.
They are already nested in the _theme.scss file.

JoomFX (changes requested):

JoomFX (approved):

PR #2692, opened 2021-08-11
fix(gantt): dependencies svg styles should not be scoped for react only

JoomFX:
This is needed for the Kendo Angular Gantt and probably for the Blazer Gantt in the future.


silviyaboteva (commented):
Hey @JoomFX,

I only have one concern if this is a bug? These styles were created at a time when there was only a jQuery version of the component and we scoped the styles in .k-treelist-scrollable class to avoid breaking changes in the existing component. I suggest renaming the commit to: 
`refactor(gantt): expose dependencies svg styles for Kendo Angular Gantt`.

What do you think?

silviyaboteva (approved):

joneff (approved):

PR #2719, opened 2021-08-19
fix(gantt): add angular gantt timeline styles

JoomFX:
This is needed for the Kendo Angular Gantt ([more details](https://github.com/telerik/kendo-angular-gantt/pull/62#pullrequestreview-732976465)).


JoomFX:
It turned out that it is not really needed.

PR #2753, opened 2021-09-02
fix: add kendo font sizes to the Bootstrap theme

silviyaboteva:
The motivation for this change is that the Typography's fontSize property uses the font-size enum to apply different sizes.


joneff (changes requested):
use `$kendo-font-sizes` as the map in all themes, not just bootstrap. also, this is a fix, not a feat.

joneff (approved):

PR #2830, opened 2021-09-28
Utility classes

silviyaboteva:
In this PR I've extracted the utility classes for setting the theme color into `utils` folder and also added a bunch of border-radius-related utility classes that will be used for the components standardization later.


JoomFX (approved):
Are we going to have a key for `50%` border-radius or the `full` key (`k-rounded-full` class) is enough?

silviyaboteva:
the `k-rounded-full` is enough I think. We can always go and add 50% later if we face a need :)

PaskalevStoyan (approved):

joneff (approved):

PR #2852, opened 2021-10-04
Colorpicker Design Update

Juveniel:


Juveniel:
@JoomFX 

We discussed this very briefly during the fe-sync meeting last week - we aim to remove the misc tests when we have a reference rendering that is adopted in all component flavors. 

As the new design for the ColorPicker is implemented simultaneously in all Kendo flavors, I have extracted the reference rendering and removed the old misc tests which are now obsolete.

PR #2873, opened 2021-10-12
feat(html): add html helper for radiobutton

PaskalevStoyan:


PaskalevStoyan:
TODO: update tests to use radiobutton html helper

PR #2953, opened 2021-11-05
Dropdowngrid v5

PaskalevStoyan:


PaskalevStoyan:
TODO:
 - Refactor table layout

packages/default/scss/table/_layout.scss
```diff
@@ -0,0 +1,241 @@
...
+            border-width: 1px 0 0 1px;
+        }
+    }
+
+    @each $size, $size-props in $kendo-table-sizes {
+        $_font-size: map-get( $size-props, font-size);
+        $_padding-x: map-get( $size-props, padding-x);
+        $_padding-y: map-get( $size-props, padding-y);
+
+        .k-table-#{$size} {
```
JoomFX, line 206: Since most of the code below does the same thing (applies the same padding), can we probably group it together? Something like:

```
.k-table-#{$size} {
    font-size: $_font-size;
    
    .k-table-header .k-table-header-wrap .k-table-header-item,
    .k-table-body .k-table-row .k-table-cell,
    .k-table-list .k-table-row .k-table-cell,
    .k-table-header .k-table-group-header > td,
    .k-table-body .k-table-group-body,
    .k-table-list .k-table-group-body
    .k-table-footer .k-table-footer-wrap .k-table-footer-item {
        padding: $_padding-y $_padding-x;
    }

}
```

packages/bootstrap/scss/table/_layout.scss
```diff
@@ -0,0 +1,9 @@
+@import "~@progress/kendo-theme-default/scss/table/_layout.scss";
+
+@include exports("table/theme/bootstrap") {
+
+    .k-table-body .k-table-group-body,
```
JoomFX, line 5: It seems that this code should be in the _theme.scss file, not in the _layout.scss.
Also, not sure where this color comes from but shouldn't it be in a variable instead of hardcoding it?

packages/material/scss/table/_theme.scss
```diff
@@ -0,0 +1,9 @@
+@import "~@progress/kendo-theme-default/scss/table/_theme.scss";
+
+@include exports("table/theme/material") {
+    .k-table-body .k-table-group-body,
+    .k-table-list .k-table-group-body,
+    .k-table-footer {
+        color: rgba(0, 0, 0, .54);
```
JoomFX, line 7: Same here, the color should probably be in a variable.

JoomFX (changes requested):
Well done. Just a few minor issues to be addressed 😃

JoomFX (approved):

silviyaboteva (changes requested):
Great work so far @PaskalevStoyan 🥳 
I have noticed two things in the PR:
- I have a question about the naming in our docs. Yes, it's true that in the themes, we call the component DropdownGrid. However, our components/widgets are called MultiColumnComboBox. So, the folder in `docs/components/table` should be renamed to  `docs/components/multicolumncombobox`  and the content should be renamed as well. What do you think about that?
- Are there any tests about the sizing of the table? I cannot see any. If not, please add such.

PaskalevStoyan:
> * I have a question about the naming in our docs. Yes, it's true that in the themes, we call the component DropdownGrid. However, our components/widgets are called MultiColumnComboBox. So, the folder in `docs/components/table` should be renamed to  `docs/components/multicolumncombobox`  and the content should be renamed as well. What do you think about that?
* I agree about the renaming of the docs' folder and content. I will rename them.
> * Are there any tests about the sizing of the table? I cannot see any. If not, please add such.
* The test for the sizing options of the table is located in the `table.html` file.

silviyaboteva (approved):

silviyaboteva:
@PaskalevStoyan please check why the build fails. Other than that, I think we are good to go :)

Juveniel (approved):

joneff (approved):
One or two things, but we can target them afterwards

PR #2969, opened 2021-11-17
fix(pager): remove redundant border radius

PaskalevStoyan:
fixes: https://github.com/telerik/kendo-themes/issues/2968


JoomFX (commented):
In my opinion this should not be fixed from the themes.
It seems to be a rendering issue. The `k-pager-nav` that is added on the `k-dropdown-wrap` element is causing the issue. And the issue is present in the Angular Pager only.
In my opinion the `k-pager-nav` class should be removed from the Angular Pager rendering as it is also not present in the React and jQuery Pager.

silviyaboteva:
I agree with @JoomFX 
This is caused by the additional style coming from `k-pager-nav` class. Please remove it from the rendering.

PaskalevStoyan:
I will close this PR and transfer the issue back to kendo-angular repo.

Issue #2960, opened 2021-11-12
FlatColorPicker - cursor: pointer to be removed from the first preview window


preslavapetrova1:
The CSS property `cursor: pointer` should be removed from the first preview window with selector 
`k-coloreditor-preview-color k-color-preview` in the FlatColorPicker component, as it is not clickable.

PR #2992, opened 2021-11-26
fix(color-preview): remove cursor pointer from preview-color element

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/2960


Juveniel (approved):

joneff (approved):
if color preview needs special handling in certain cases, say when in color editor, then the selector should be inside color-editor file ;) we can fix that one later

Issue #2576, opened 2021-06-29
Column menu icon is visible over sticked columns


stnikolova:
**Describe the bug**
In a Grid with sticky columns and column menu enabled, the column menu's icon remains visible over sticked columns on horizontal scroll.

**To reproduce**

1. Go to https://www.telerik.com/kendo-angular-ui/components/grid/columns/menu/#toc-stick-item
2. Start scrolling the example horizontally

**Expected behavior**
When a column is hidden, the column menu icon should not be visible as well.

**Screenshots**

https://user-images.githubusercontent.com/18744873/123785594-6180bc00-d8e1-11eb-900d-fc6983da53e1.mov

**Affected package**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites**
* Kendo UI for Angular

**Affected browsers**
* All

silviyaboteva:
Can you please check if this is a rendering problem? As in the KendoReact, the icon displays ok while scrolling: https://stackblitz.com/edit/react-bdkmco?file=app/main.tsx

PR #2994, opened 2021-11-26
fix(grid): decrease column menu icon z-index

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/2576


silviyaboteva (approved):

joneff (approved):
I am pretty sure there is a reason why we use `5`, but I cannot recall.

PR #3035, opened 2021-12-07
fix(grid): filter button is not visible in ie

Juveniel:
related to: https://github.com/telerik/kendo/issues/12976

The removed code causes the filter button in the Grid header not to be visible due to the right margin adjustment. This is reproducible only when the Grid is sortable (different rendering). 

The code block is no longer needed, as it is targeting older versions of IE.


joneff (approved):
❤️

Issue #2790, opened 2021-09-15
Filter Component Design Revision


Raisolution:
Possible FE work might be needed related to https://github.com/telerik/kendo-ux-private/issues/379

Logging this just in case.

PR #3049, opened 2021-12-10
Adjust styles for angular rendering

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/2790


joneff (approved):
There is no image test for this enhancement, but nothing is changed in the existing tests, so I am bringing it in.

Note: I changed the commit type from `feat` to `fix`, since I didn't see anything new.

Issue #2664, opened 2021-07-28
Unexpected margin before a form field preceded by a hidden field


Dimitar-Goshev:
This is a needed themes enhancement related to adding hidden form field support: https://github.com/telerik/kendo/issues/12665

**To reproduce**

1. Use the following setup for hidden form field:
```
    <form id="myForm"></form>
    <script>
        $("#myForm").kendoForm({
            formData: {
                ID: 1,
                Name: "John Doe"
            },
            items: [{
                field: "ID",
                editor: "hidden",
                validation: { required: true }
            }, {
                field: "Name",
                validation: { required: true }
            }]
        });
    </script>
```
1. This renders as follows:
```
    <form id="myForm" data-role="form" class="k-widget k-form" novalidate="novalidate" data-stop="true">
        <div class="k-form-field  k-hidden">
            <div class="k-form-field-wrap" data-container-for="ID"><input type="hidden" id="ID" name="ID" title="ID"
                    required="required" autocomplete="off" data-bind="value:ID" value="1" class="k-valid"></div>
        </div>
        <div class="k-form-field  "> <label class="k-label k-form-label" for="Name" id="Name-form-label">Name:</label>
            <div class="k-form-field-wrap" data-container-for="Name"><input type="text" id="Name" name="Name"
                    title="Name" required="required" autocomplete="off" aria-labelledby="Name-form-label"
                    data-bind="value:Name" class="k-textbox k-valid"></div>
        </div>
        <div class="k-form-buttons"><button class="k-button k-primary k-form-submit"
                type="submit">Submit</button><button class="k-button k-form-clear">Clear</button></div>
    </form>
```
1. Themes have the following rule for form's first child
```
.k-form > *:first-child, .k-form .k-daterangepicker .k-textbox-container:first-child {
    margin-top: 0;
}
```

And for the rest of the child elements:
```
.k-form .k-form-field, .k-form .k-form-buttons, .k-form .k-daterangepicker .k-textbox-container {
    margin-top: 1rem;
}
```

Since the hidden field is rendered as first child, the actual first visible item has margin-top: 1rem applied.

**Expected behavior**
There is no margin-top before the first field label:
![image](https://user-images.githubusercontent.com/6614183/127322531-e8f6a91a-1770-4927-981d-3dbbd0a480cb.png)

**Screenshots**
![image](https://user-images.githubusercontent.com/6614183/127322681-285547ec-751a-4413-b234-1425bbdd7a71.png)

[...]

PR #3102, opened 2021-12-22
Remove unexpected margin before a form field preceded by hidden field

epetrow:
Fixes: #2664


joneff (approved):

joneff:
This fix is until we drop IE and have a proper fix with flexbox / grid

Issue #3088, opened 2021-12-19
Add default width to searchbox in Grid, TreeList and TaskBoard


jivanova:
The width of the inputs in v5 is width 100%, however, for the integration in the grid and treelist we need a restricted max size.

<details>
<summary>Current Rendering</summary>

``` html

<span class="k-grid-search k-textbox k-input k-input-md k-rounded-md k-input-solid">
  <input autocomplete="off" placeholder="Search..." title="Search..." class="k-input-inner">
  <span class="k-input-suffix"><span class="k-icon k-i-search"></span></span>
</span>

```

</details>

**Current behavior**

![image](https://user-images.githubusercontent.com/7159157/146665626-515cba49-750f-4e6b-975c-b5aca913d7b0.png)

**Expected**

Searchbox should be placed to the right side of the toolbar.

Dimitar-Goshev:
There is a similar issue in Kendo UI for jQuery TaskBoard. Current rendering:
```
<span class="k-searchbox k-textbox k-input k-input-md k-rounded-md k-input-solid" data-uid="ce5a77d9-c477-45e7-828f-0465ed7d776f" id="search-c99e04c5-0747-4f95-9b80-7c67a6f30cff" data-overflow="never">
    <span class="k-input-prefix">
        <span class="k-icon k-i-search"></span>
    </span>
    <input class="k-input-inner" autocomplete="off" aria-label="Search" title="Search" data-command="SearchCommand" placeholder="Search">
</span>
```
The search falls on a new row, instead of being on the same level as Add column button:
![image](https://user-images.githubusercontent.com/6614183/147094024-dd336a16-0d39-489c-b246-538ee678091b.png)

PR #3117, opened 2021-12-24
fix(toolbar): set default width for inputs

Juveniel:
targets: https://github.com/telerik/kendo-themes/issues/3088

needs more work (we have cases like the Editor, Scheduler, Adaptive, where we want the input/pickers to have content width..)


joneff:
few issues with adaptive grid / scheduler and media player, but nothing serious.

joneff (approved):

PR #3137, opened 2022-01-03
Html snapshots

tsvetomir:
* Renames tests in ` tests/visual/src/reference` that use theme "html components" to `.comp.html`
* Adds a script that snapshots component test pages and writes vanilla `.html` output using `@progress/kendo-e2e`

To do:
* [x] Figure out when to run the script.
  Probably in "CI Visual", so the updated HTML snapshots can be included in the screenshots and committed.
* [x] Exclude `.comp.html` sources from screenshots
* [x] Add/update READMEs

ping @joneff


tsvetomir:
This is more or less ready, but it's not something we must merge before the release.

Do you think the overall process makes sense?
* Reference tests are authored as usual, but in a file named `.comp.html` and then the CI build updates both the static `.html` files and the screenshots.
* Browsing the repo gives you an overview of all three, components markup, plain markup and screenshots without needing to clone or run anything.

joneff:
I think it's a step in the right direction. Ultimately, I would like the pages to be full fledged tsx, but that can be done at a later step.

joneff (approved):

tsvetomir:
That would be ideal in terms of development experience, but will take a bit of time and effort. Hence the patchy solution in this PR :)

We can convert `packages/html` into a React app and move the reference markup pages there. The structure in `tests/visual` will remain unchanged and the snapshots will be created from the app.

We get all the goodies like hot reload and fast unit tests, but the files now live in separate folders. This is similar to the suite-specific implementations that are also generated elsewhere.

Juveniel (approved):

tsvetomir:
Seems like a good moment to merge this in order to get the reference markup snapshots in the repo.

There's an idea to refactor the reference examples into a regular React app which will obsolete it, but this is still just an idea.

joneff (approved):

PR #3173, opened 2022-01-07
fix(html): update rendering of the split button to include rounding o…

silviyaboteva:
…n root element

Part of https://github.com/telerik/kendo-themes/issues/2839


joneff (approved):
if the outer wrapper can be focusable, then the inner shouldn't be

Issue #3176, opened 2022-01-07
Expand button is not aligned properly in Safari


dtopuzov:
**Describe the bug**
Expand button is not aligned properly in Safari.

**To reproduce**

1. Open `combobox` demos of jQuery build with themes v5 (it is ok with v4).
1. See the expand button

**Expected behavior**
Expand button aligned properly.

**Screenshots**
![image](https://user-images.githubusercontent.com/6651651/148549326-95abb17d-3257-436a-8831-19b853403193.png)

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material
* theme-tasks

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery

**Affected browsers (please remove the unneeded items)**
* Safari

**Build system information (please remove the unneeded items)**
* Not Applicable

ahmadjubair33:
hey can I solve this isssue?

silviyaboteva:
fixed in `v5-dev`: https://github.com/telerik/kendo-themes/pull/3192

PR #3192, opened 2022-01-10
fix: reset margin for inputs and pickers direct children v5

silviyaboteva:
Fixes: https://github.com/telerik/kendo-themes/issues/3176


joneff (approved):
🤦

PR #3206, opened 2022-01-11
Docs escape values

kspeyanski:
The following PR fixes an issue where the generated markdown is not valid due to calculations inside a variable's value, thus the corresponding website documentation build fails.

The exact issue is reproducible with the value of the `$button-calc-size` variable, which is `calc( #{$button-line-height * 1em} + #{$button-padding-y * 2} + #{$button-border-width * 2} )`. Put into an MDX syntax, this string is not valid unless escaped.

The new-lines around the code-snippets are required to trigger markdown parsing.

### DIFF
--- a/docs/templates/includes/theme-variables-row.md.njk
+++ b/docs/templates/includes/theme-variables-row.md.njk
@@ -1,8 +1,17 @@
 <tr>
     <td>${{ variable.context.name }}</td>
     <td>{{ variable.type }}</td>
-    <td>{% if variable.context.value | is_color === true %}<span class="color-preview" style="background-color: {{variable.context.value}}"></span>{% endif %}{{ variable.context.value }}</td>
-    <td>{% if variable.resolvedValue | is_color === true %}<span class="color-preview" style="background-color: {{variable.resolvedValue}}"></span>{% endif %}{{ variable.resolvedValue }}</td>
+<td>
+{% if variable.context.value | is_color === true %}<span class="color-preview" style="background-color: {{variable.context.value}}"></span>{% endif %}
+
+{% if variable.context.value %}`{{ variable.context.value }}`{% endif %}
+
+</td>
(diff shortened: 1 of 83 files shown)


docs/templates/includes/theme-variables-row.md.njk
```diff
@@ -1,8 +1,17 @@
 <tr>
     <td>${{ variable.context.name }}</td>
     <td>{{ variable.type }}</td>
-    <td>{% if variable.context.value | is_color === true %}<span class="color-preview" style="background-color: {{variable.context.value}}"></span>{% endif %}{{ variable.context.value }}</td>
-    <td>{% if variable.resolvedValue | is_color === true %}<span class="color-preview" style="background-color: {{variable.resolvedValue}}"></span>{% endif %}{{ variable.resolvedValue }}</td>
+<td>
```
silviyaboteva, line 4: May you trim the lines as before, e.g.:
```
<td>{% if variable.context.value | is_color === true %}<span class="color-preview" style="background-color: {{variable.context.value}}"></span>{% endif %}{% if variable.context.value %}`{{ variable.context.value }}`{% endif %}</td>
<td>{% if variable.resolvedValue | is_color === true %}<span class="color-preview" style="background-color: {{variable.resolvedValue}}"</span>{% endif %}{% if variable.resolvedValue %}`{{ variable.resolvedValue }}`{% endif %}</td>
```

So, the generated output doeesn't include so many new lines and produce for example  <td>`1px`</td>

kspeyanski: This is still a `md` file so the new lines won't appear in the final HTML - their only purpose is to trigger code-snippet parsing, so the final formatting will be correct, without new lines.

silviyaboteva (approved):

joneff (approved):

PR #3228, opened 2022-01-13
fix(timeselector): remove outline

JoomFX:


joneff (approved):

joneff:
To clarify why the above is needed: the heuristics of chrome determine that `.k-time-list` should be focused and it goes into `:focus-visible` state, hence the outline on windows and focus ring on mac.

Issue #1572, opened 2020-05-26
Grid detail row container has a separate hover styling


veselints:
**Describe the bug**
Grid detail row container has a separate from its master row hover styling. It should not apply hover styling as it is not a row.

**To reproduce**
Run the following Dojo and hover over the detailed container:
https://dojo.telerik.com/EYeYUSUw/6

**Expected behavior**
The detail container should not have a hover state, or it should display its state together with its master row.

**Screenshots**

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery

**Affected browsers (please remove the unneeded items)**
* All

TeyaVes:
related to: https://github.com/telerik/kendo-themes/issues/1276

PR #3246, opened 2022-01-17
fix(grid): remove separate hover styling on detail row container

epetrow:
Fixes: https://github.com/telerik/kendo-themes/issues/1572


silviyaboteva (changes requested):
You need to apply this also to the selectors with classes `.k-state-hover`

joneff (approved):

Issue #2887, opened 2021-10-15
Include Disabled Variables for NumericTextBox and DateTimePicker 


DaleyKD:
**Is your feature request related to a problem? Please describe.**
I am trying to get the disabled state of the NumericTextBox and DateTimePicker to more closely match our Bootstrap theme. I was hoping I could simply use some kendo variables, but none exist.

**Describe the solution you would like to see implemented**
Like some of the other controls (button and textarea come quickly to mind), let us have some variables for this?

**Describe the alternatives you have considered**
I'm doing the following for now:
``` scss
/* These controls don't have variables for disabled stuff. */
.k-numerictextbox .k-numeric-wrap,
.k-datetimepicker .k-picker-wrap {
    &:disabled,
    &[disabled],
    &.k-state-disabled {
        @include fill($input-color, $input-disabled-bg);
        opacity: 1;

        &::selection {
            @include fill($input-color, transparent);
        }

        .k-select {
            opacity: $disabled-opacity;
        }
    }
}
```

joneff:
fixed by #3268

**A very important note**

Before v5, all inputs had a set of variables of their own -- `$combobox-bg`, `$numeric-text` etc. Now we have high-level components for all inputs (a component you can type in) and pickers (components you can just choose value) and their respective values are `$kendo-input-*` and `$kendo-picker-*`.

In the case of the numeric, you need `$kendo-input-disabled-*` variables, but this will also apply to the rest of the input components.

There is also the matter of applying `opacity` to disabled components, so you will need to reset that. Material is the sole exception, as it already had dedicated color for disabled background.

Finally, while this fix is already in the code base, it will be made available with the next dev version on Monday.

PR #3268, opened 2022-01-21
Add disabled state variables to input and picker high-level components

epetrow:
Related to: https://github.com/telerik/kendo-themes/issues/2887


packages/default/scss/datetimepicker/_theme.scss
```diff
@@ -1,6 +1,21 @@
 @include exports( "datetimepicker/theme" ) {
 
     // Datetime picker
-    .k-datetimepicker {}
+    .k-datetimepicker {
+
+        //Disabled state
+        &:disabled,
+        &[disabled],
+        &.k-state-disabled,
```
silviyaboteva, line 9: As of v5, we've switched from `k-state-*` to more generic  `k-*`. You've noticed that in some modules there are still selectors with `k-state-*` class. But they will be updated in the R2 2022 effort. 

In this PR we work with numeric and datetimepicker that was part of the v5 effort for R1 2022. So the `&.k-state-disabled` selector have to be removed.

silviyaboteva (approved):

joneff (changes requested):
those should be `$kendo-input-` and `$kendo-picker-` variables.

joneff (approved):

PR #3286, opened 2022-01-26
fix(input): conditionally set spinner offset to calculated or auto 

Juveniel:


joneff (changes requested):
The icon offset must remain for some cases. It should be handled probably like `top: if($offset, -1 * $offset, null)`

dimodi:
FYI, the customer confirmed that these styles were indeed breaking CSS minification.

> The icon offset must remain for some cases. It should be handled probably like `top: if($offset, -1 * $offset, null)`

Out of curiosity, how will the `null` expression end up in the CSS? Will the `top` style be skipped altogether or the build will transform it to a valid value such as `auto`?

Juveniel:
@dimodi basically, we have a specific theme variable `$use-input-spinner-icon-offset: false !default;` which is `true\false`. Based on it, we set the value to either `auto` if it is `false`, or to the specified variable `$kendo-input-spinner-icon-offset`.

Thus, to change the spinner offset, one must set the flag to true and provide a value for the `$kendo-input-spinner-icon-offset` variable for the specific theme.

joneff (approved):

joneff:
Null, if assigned correctly, will not render any output. The `if` expression also supports null values i.e:
```scss
selector {
    bottom: if( $use-input-spinner-icon-offset, calc( -1 * #{$kendo-input-spinner-icon-offset}), null );
}
```
in which case this will render noting.

Issue #3279, opened 2022-01-25
Missing border in PivotGrid header 


MilenaCh:
The left vertical border of 'All Periods' header is missing in Material theme:
![pivotgrid4](https://user-images.githubusercontent.com/3844466/150944344-ba373738-7d15-4dfd-96fd-4dba0a8a6f46.PNG)
Demo: https://kendobuild-windows.dev.progress.com/kendo-ui-next/pivotgrid/index?autoRun=true&theme=material-main

PR #3288, opened 2022-01-26
fix(pivotgrid): add vertical borders for the legacy pivot rendeing in…

silviyaboteva:
… Material theme

Fixes: https://github.com/telerik/kendo-themes/issues/3279

This fix is relevant to legacy rendering only!

> Note: After offline discussion with @ivelina-penkova, we came to a conclusion that the vertical borders are needed in the PivotGrid component as there are sub-totals and grand-total columns. If there are no borders, it will be hard to inspect/review the data. 
The design team is OK with the fact that the PivotGrid will differ from the Grid.
[Here is the updated design llink](https://www.figma.com/file/Bun5XXsG0UuzCMz5GyDoHK/Telerik-%26-KENDO-UI-Kit-for-Material-1.9--(UNPUBLISHED)?node-id=0%3A1).


Juveniel (approved):

joneff (approved):
I really wanted to global grid settings and not to deviate from our standard table... Alas this is not the case

PR #3293, opened 2022-01-27
fix(button): font awesome icons in icon button are not properly aligned

JoomFX:
The `.fa` class has `display: inline-block;` that overrides the `display: inline-flex;` in our `.k-icon` class.
Adding more specificity to overcome the issue so we can have the FA icons properly aligned.

fa-icon


joneff (approved):

joneff (commented):
The item is not a bug, because it's something not within our code. it's related to font awesome, so it's an enhancement.

silviyaboteva (approved):

Issue #3311, opened 2022-01-27
Dropdowns in the Toolbar are not wide enough


JoomFX:
**Describe the regression**
Since v5 of the themes there is [a rule](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/toolbar/_layout.scss#L68) that sets the width of the Dropdowns in the Toolbar to `10em`.
However, the width of the Dropdowns in v4 was `12.4em`.
With the new/smaller width there are some visual issues that can be seen in the screenshot and the Stackblitz examples below.

**To reproduce**
[Example with Kendo Themes v4](https://stackblitz.com/edit/angular-asaxgh?file=index.html)
[Example with Kendo Themes v5](https://stackblitz.com/edit/angular-gmsjr9?file=app/app.component.ts)

**Expected behavior**
The Dropdowns in the Toolbar should be wider (probably the same as in v4 - `12.4em`).

**Screenshots**
![dd-width](https://user-images.githubusercontent.com/7753940/151368355-ae05f3e8-f999-4722-8917-50069f5b35c7.png)

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**
* All

**Build system information (please remove the unneeded items)**
* Not Applicable

joneff:
I've explained in detail why we'll not handle this issue -- https://github.com/telerik/kendo-themes/pull/3313#issuecomment-1025627606 -- but here is the gist of it:

* default input width is 20 characters, which is close to 10 em;
* even if we increase width, we are not guaranteed to fit, especially in languages like french
* we have the ability for non-wrapped items, which is the way we can and should handle format dropdown
* we can always use shorter labels for the inputs "Font size" and "Font family" vs "Select ..."
* finally, developers could customize the width of the components with inline styles when our guidelines are not wide enough.

Closing.

PR #3313, opened 2022-01-27
fix(toolbar): increase dropdowns width

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/3311


silviyaboteva (approved):
I'm OK unless the change to 10em was intended for some reason.

Juveniel (changes requested):
@JoomFX by default an input has a [size of 20](https://html.spec.whatwg.org/multipage/input.html#the-size-attribute) which corresponds to the number of characters that a user sees while editing the input. While we cannot use the size attribute in our components, we can set default `width` in the themes that corresponds to that default size (20 characters), which is roughly `10em`.  This is the reason behind the observed change.  

In terms of the proposed change to `12.4em` value - we cannot assume that this width will be sufficient for every component that features a ToolBar, as there are different DropDowns with different character sizes in the components (ImageEditor, FileManager, Pager, etc.) Thus, I believe that a better solution is to increase the width of the pickers in the Editor specifically and not in the ToolBar.

JoomFX:
@Juveniel I'm fine with setting the width in the Editor specifically, however I do believe that doing it in the Toolbar should be fine as well. Especially since the inputs/pickers had this same width (`12.4em`) during all their life and all customers are used to it.
There's no problem at all to add the code in the Editor, I just don't like the idea of adding new code that overrides the code that causes the issue instead of modifying the code that causes the issue.

@joneff @silviyaboteva what do you think?

silviyaboteva:
@Juveniel thank you for clarifying. Now, I remember that we've talked with @joneff about this.

We've made significant changes in the v5 release, so IMO this change is not an issue as we can now say that the default value is a predictable one (20 chars). 

I can suggest in this case to go through the demos of the components and change the widths of the inputs/pickers there for better appearance, and not to change anything from themes.

JoomFX:
@silviyaboteva Well, I don't insist in doing the change in the Toolbar styles, but I do believe that we need to address this issue in the themes.
Please have a look at the issue and the problem we are trying to fix. If we fix the examples only, then the customers will need to add the same custom code to their applications.

joneff:
What @Juveniel said is indeed so -- 20 characters is the default width, which is roughly equal to 10em. For legacy reasons, `12.4em` was transferred blindly from the less to sass themes.

As for the issue we are trying to fix, I understand, but the position stays the same. For the record, in french the width would be vastly insufficient just for the first word which would be "selectionnez / choisir la famille de police"

For "format" we have chosen shorter word e.g. format. I am guessing we do the same for "font family" or "font size". It's a resource after all.

image

That being said, components in toolbars can be sized not only from the themes, but with inline styles. In terms, should a team decides they need a wider box, they can do so by setting inline width. Like @Juveniel 12.4em is not going to be sufficient for all cases, so if a team decides to deviate from our standard (10em) it should do so with caution.

Finally, regarding format dropdown, we can always use non-wraping lines to fit everything.

I am closing this issue.

PR #3339, opened 2022-02-01
Resolve sass variables before markdown generation

### LINKED ISSUE #1610: Create a json file for theme variables
We need a json file containing all the meta information about variables in the theme:

* friendly name
* description
* group
* deprecated version
* deprecated description
* variable type (unit, color, ... )
* variable sub type ( text color, border-color ... )
* value, as is
* value, computed
* parent variable(s), if it uses
* derived (children) variables, if any

The above implies that we have annotations for friendly name and sub type and any additional meta we need that cannot be generated.

At the end the json should look like:
```js
{
    "version": "VERSION", // the semantic version of the theme
    "name": "theme-name", // the name of the theme
    "variables": {
        "$blue": {
            "name": "", // friendly name, from annotation
            "description": "", // description, from annotation
            "deprecated": "VERSION", // the semantic version, from annotation
            "deprecatedDescription": "", // reasoning behind deprecation, from annotation
            "group": "", // group name, from annotation
            "type": "color", // sass variable type, from annotation 
            "subtype": "text-color", // custom subtype, from annotation
            "value": "as is | EXPRESSION -> native css, function, var-ref", // generated
            "computed-value": "computed", // generated
            "parent": "$VARIBALE_NAME || EMTPTY_STRING", // generated
            "children": ["$VAR_NAME", "$VAR_NAME", "EMPTY_ARRAY"] // generated
        }
    },
    ...
}
```

We also need a json schema to validate against.

The resulting file can be used to generated to augment the documentation, create variables hierarchy or if needed to be consumed from other applications like Sass Themebuilder.

### LINKED ISSUE #1611: Integrate variables.json in the documentation
After #1610, integrate the json file data in the documentation, namely:

**Tier 1**
* resolved variable value

**Tier 2**
Additional meta information like subtype (and proper preview based on subtype).

joneff:
This is as convoluted as things get in sass...

1) flatten all themes using baka
2) import said file using the `@use` dart-sass api
3) "compile" said file using dart-sass with custom function that processes each sass variable
4) put everything in a humongous json file
5) during the docs generation, suggest to sassdoc what the variables should be

**Important** that json generation does take some time...

* closes #1610, at least partially
* closes #1611, completely

### DIFF
--- a/docs/sassdoc-theme.js
+++ b/docs/sassdoc-theme.js
@@ -1,17 +1,25 @@
 const fs = require('fs');
 const path = require('path');
-const sassdocExtras = require('sassdoc-extras');
 const nunjucks = require('./nunjucks');
 
 module.exports = function(dest, context) {
     const capitalize = (str) => str[0].toUpperCase() + str.substring(1);
-
-    sassdocExtras( context, 'resolveVariables' );
+    const resolvedVars = JSON.parse( fs.readFileSync(context.json), 'utf-8');
 
     const data = context.data
         .filter((item) => item.access === 'public')
         .reduce((acc, item) => {
             if (item.context.type === 'variable') {
+                let varName = item.context.name;
+
+                // Augment resolved value
+                item.resolvedValue = resolvedVars[varName].value;
+
+                // Augment variable type
+                item.type = item.type !== undefined
+                    ? item.type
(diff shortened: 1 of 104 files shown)


tsvetomir (approved):
🌈 Lovely

Issue #3050, opened 2021-12-10
Input issues related to the new styling options


MilenaCh:
- [x]  Rounded is not applied properly(all inputs) - the proper classes are set to the inputs, but the styling is the same in all rounded modes
- [x] Small size should be 24px.
- [x] Focused style of maskedtextbox and dateinput doesn't correspond to the other inputs(there is no shadow)
- [x] Spin buttons of NumericTextBox don't look good in small size(Material,Default and Classic themes)
- [ ]  The text of floating label is not centered in the small size(all inputs)
- [x] Text selection color is always blue. In the previous version, the color corresponds to the theme's base color(all inputs)
- [ ] Placeholder text disappears when click inside the input and floating label is set(textbox/index) - it's valid for all inputs
- [x] Spin buttons of numerictextbox don't look good in Bootstrap and Classic themes:

<details>
<summary>Image</summary>

![numeric](https://user-images.githubusercontent.com/3844466/146397573-2b13cead-5996-4902-8269-20ab010243fd.PNG)

</details>

- [ ] TextArea small size with rounded full text and selection is cut 

<details>
<summary>Image</summary>

![image](https://user-images.githubusercontent.com/7159157/146952052-9f06ed62-b01f-4300-a369-9558e94ab239.png)

</details>

**Integration issues**

- [x] Missing shadow:
Shadow is not applied to textbox on focus - BreadCrumb (editmode), Filter, PivotGrid(Filter inputs), FileManager(SearchBox), Pager, Scheduler(textaria in edit mode), SpreadSheet(Data Validation, Insert comment, insert hyperlink, export), Grid(filter input, Searchbox)

- [x] [FileManager] Vertical border appears before search icon when searchbox is focused. 
- [x] [Grid, Treelist TaskBoard] Proper width should be set to SearchBox - currently it is expanded on 100% width of the main container
- [x] [TaskBoard] Search icon is located too close to input's border
- [x] Spreadsheet - Filter Menu(http://kendobuild-windows.dev.progress.com/kendo-ui-next/spreadsheet/sorting-filtering)
- [x] broken Filter by condition input (http://kendobuild-windows.dev.progress.com/kendo-ui-next/spreadsheet/sorting-filtering)
![image](https://user-images.githubusercontent.com/6614183/145858044-12cd1f83-1406-4f22-aad5-e23f771bfea6.png)

[...]

PR #3343, opened 2022-02-01
fix(icons): custom icons are not centered

PaskalevStoyan:
Custom icons are not centered because their CSS is overriding our styles.
for more information: https://github.com/telerik/kendo-themes/issues/3050#issuecomment-1006527761


joneff (changes requested):
The scrollview image tests have changed. Please take a closer look why.

Juveniel:
@PaskalevStoyan there is another PR targeting the same - https://github.com/telerik/kendo-themes/pull/3293

joneff:
The fix from @JoomFX is specific to the button. in reality, it can applied to all components. for the time being, we'll merge ivo's PR. However, we must think of long term solution.

Issue #3353, opened 2022-02-02
RTL display breaking when floating label is placed inside a container with [dir="rtl"]


epetrow:
**Describe the bug**

RTL display breaking when floating label is placed inside a container with [dir="rtl"] in Default and Bootstrap themes.

**To reproduce**

1. See screenshots from visual tests below:

Default theme:

![image](https://user-images.githubusercontent.com/95700708/152170444-016715b8-8166-4874-bc78-9c8f968d5f7e.png)

Bootstrap theme:

![image](https://user-images.githubusercontent.com/95700708/152170611-dc2ac96d-8cbf-4514-9d88-87444cc926d3.png)

**Expected behavior**

Floating labels to be properly aligned when placed inside a container with [dir="rtl"]

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**
* All

**Build system information (please remove the unneeded items)**
* Not Applicable

PR #3354, opened 2022-02-02
fix(floating-label): fix rtl display when floatiating label is placed…

epetrow:
… inside container with [dir=rtl]

Fixes: https://github.com/telerik/kendo-themes/issues/3353


silviyaboteva (approved):
<3

joneff (approved):

PR #3355, opened 2022-02-02
fix(dataviz): set default barcode height

tsvetomir:
Ensures that both SVG and Canvas surfaces are equal when no dimensions are set.

Runnable demo: https://dojo.telerik.com/AFaDULeh/2


Stamo-Gochev:
Do we need to do a similar thing for the qrcode for consistency? Right now, its height is around [150](https://dojo.telerik.com/AFaDULeh/6), but this is set as an inline style rather than coming from the themes. There is also a fallback to [200](https://github.com/telerik/kendo-charts/blob/ceb386e9d0d22c1a43f51030e7dda7ff17b1c330/src/qrcode/qrcode.js#L23).

tsvetomir:
Nice catch, I assumed the 200px size is always rendered, but this is not the case in the jQuery version.

Stamo-Gochev:
Any ideas what was the idea behind the check for the default size of 200?
https://github.com/telerik/kendo-charts/blob/ceb386e9d0d22c1a43f51030e7dda7ff17b1c330/src/qrcode/qrcode.js#L247-L254

By the time this check happens, the dimensions of `this.element[0]` are > 0, so this value is never reached.

tsvetomir:
@Stamo-Gochev hmm, not sure, we may need to revisit this. We should keep the default at 150px to maintain compatibility.

joneff (approved):

PR #3359, opened 2022-02-04
fix(stepper): label overflow doesnt work

PaskalevStoyan:
Fixes: https://github.com/telerik/kendo-themes/issues/2956


silviyaboteva (commented):
It seems that the space between the stepper and the content of the wizard test is bigger than previously. Can you check why?

joneff (approved):

Issue #2723, opened 2021-08-23
Progress Bar misplaced when the Wizard height is more than 440px 


ntacheva:
**Describe the bug**

When Wizard Stepper has left or right orientation and you set the component height to 440px and higher, the progress bar is moved away from the steps in downside direction. The higher the Wizard, the bigger the offset of the steps progress bar.

**To reproduce**

1. Go to [Blazor Wizard Configuration demo](https://demos.telerik.com/blazor-ui/wizard/configuration)
1. Use the Dropdown to set the Stepper position to Left/Right
1. Use the dev tools to set height to the div with classes k-widget, k-wizard, k-wizard-vertical and enter value higher than 440px (the main wrapping container of the Wizard - when you provide some value to its Height parameter, the value is applied to that container - this way it can be tested in the live demo).

**Expected behavior**

Progress Bar should stick to the steps regardless of the Wizard height

**Screenshots**
![image](https://user-images.githubusercontent.com/73842592/130463943-267a209f-c6ab-40ff-9664-695a51d1727a.png)

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please remove the unneeded items)**

* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**

* Firefox version: [89]
* Chrome version: [91]
* Edge version: [92]

veselints:
The following style:
```css
.k-progressbar-vertical {
            position: absolute;
            top: calc(#{$stepper-indicator-height} / 2);
}
```

Will destroy the positioning of the progressbar element in jQuery.:
https://dojo.telerik.com/IheWoZuc/4

The `top: ...` style should be supplied to the element as an inline style, just like all the `display: grid` styling is done. The jQuery Stepper uses mutch simpler element calculation logic, which does not include `display: grid`. Hence, any styling relevant to that case should be kept inline.

PR #3368, opened 2022-02-07
Vertical stepper

epetrow:
Fixes: https://github.com/telerik/kendo-themes/issues/2723

Needs changes in the inline styles set via the component. Please check the visual test for reference:
 - [x] Kendo UI for jQuery //cc @veselints 
 - [ ] Kendo UI for Angular
 - [x] KendoReact (https://github.com/telerik/kendo-react-private/pull/1663)
 - [ ] Kendo UI for Vue //cc @zdravkov 
 - [X] Telerik Blazor @KristianMariyanov (https://github.com/telerik/blazor/pull/3870)


silviyaboteva (approved):
🎉

veselints:
@epetrow 
The following style:
```css
.k-progressbar-vertical {
            position: absolute;
            top: calc(#{$stepper-indicator-height} / 2);
}
```

Will destroy the positioning of the progressbar element in jQuery.:
https://dojo.telerik.com/IheWoZuc/4

The `top: ...` style should be supplied to the element as an inline style, just like all the `display: grid` styling is done. The jQuery Stepper uses mutch simpler element calculation logic, which does not include `display: grid`. Hence, any styling relevant to that case should be kept inline.

veselints (changes requested):
The following style:
```css
.k-progressbar-vertical {
            position: absolute;
            top: calc(#{$stepper-indicator-height} / 2);
}
```

Will destroy the positioning of the progressbar element in jQuery.:
https://dojo.telerik.com/IheWoZuc/4

The `top: ...` style should be supplied to the element as an inline style, just like all the `display: grid` styling is done. The jQuery Stepper uses mutch simpler element calculation logic, which does not include `display: grid`. Hence, any styling relevant to that case should be kept inline.

epetrow:
@veselints Thank you for the input, in that case there is no need for any changes in the JQuery widget.

veselints (approved):

KristianMariyanov (approved):

joneff (approved):

Issue #3372, opened 2022-02-08
The $kendo-button-border-radius variable doesn't work


esushilnikov:
**Describe the bug**
The $kendo-button-border-radius variable doesn't affect kendoButton elements.

**To reproduce**

1. Set $kendo-button-border-radius variable to any value

**Expected behavior**
The border-radius should be affected by the value of the variable.

**Screenshots**
![image](https://user-images.githubusercontent.com/7443440/152992240-d564f0fc-0861-4f5a-918e-4035094aee6d.png)

**Affected package (please remove the unneeded items)**
* theme-default

**Affected suites (please remove the unneeded items)**
* Kendo UI for Angular

**Affected browsers (please remove the unneeded items)**
* All
* Not Applicable
* Firefox version: [for example, 72, lts]
* Chrome version: [for example, 80]
* Edge version: [for example, 80]
* Safari version: [for example, 13]
* IE version: [for example, 11]

**Build system information (please remove the unneeded items)**
* Not Applicable
* OS: [for example, Windows, Linux, Mac]
* Node version: [for example, 10.15.3]
* NPM version: [for example, 6.4.1]

**Additional context**
Add any other context about the bug.

PR #3375, opened 2022-02-09
fix(button): add $kendo-button-border-radius variable to layout

epetrow:
Fixes: https://github.com/telerik/kendo-themes/issues/3372


PaskalevStoyan (approved):

tsvetomir (approved):

joneff (approved):
as a side note, we should revisit all usages of border radius to ensure that `$enable-rounded` works

Issue #2784, opened 2021-09-14
Scheduler cuts off translated text inside times-all-day th element


yanmariomenev:
**Describe the bug**
Translating "All Day" in German(for example) cuts off the text in the Scheduler component. For reference ticket number: 1535141
![chrome_bcMkAMmoAH](https://user-images.githubusercontent.com/49076727/133258828-2541a4ad-ebfa-47ff-832a-d2f9684197aa.png)
In German - 'Ganzer Tag'
![chrome_wJCu1kPtaS](https://user-images.githubusercontent.com/49076727/133259153-a201b6c6-8032-40c2-bf50-5ff0f94de10f.png)

**To reproduce**
https://stackblitz.com/edit/angular-2iryu3-yptazw?file=app/app.component.ts

**Expected behavior**
For the overflow not to be hidden.

**Screenshots**

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for Angular

**Build system information (please remove the unneeded items)**
* Not Applicable

PR #3314, opened 2022-01-27
fix(input): excess text is not properly cut

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/3263

PR #3378, opened 2022-02-11
fix(scheduler): add text-overflow to times-all-day th element

epetrow:
Related to this discussion: https://github.com/telerik/kendo-themes/pull/3314#issuecomment-1034972336

Fixes: https://github.com/telerik/kendo-themes/issues/2784


silviyaboteva (approved):

tsvetomir:
Is the English label concatenated by default? See the screenshots.

joneff:
> Is the English label concatenated by default? See the screenshots.

No the english label is not truncated / clipped by default.

The issue here is that in kendo-angular-scheduler the cell that holds "all day" is sized based on the width of the cells that hold the time and there is no additional setting (component or otherwise) to override that width. Unlike jquery, which use the bigger of the two values.

In any case, there is little we can do from the themes alone.

Lastly, due to how clipping and truncation work, in this particular case, truncation does worse job than clipping. I am closing the pr.

Issue #3290, opened 2022-01-26
Component options update based on the design revision


hadjieva:
Changes are expected mainly in the Size and FillMode options of buttons and inputs, so once the suggestions for improvement are provided by the design team, they should be applied to the Themes. Here is the respective design issue: [Component options - Design Revision](https://github.com/telerik/web-components-ux/issues/410) on which this item depends.

joneff:
fixed #3382

joneff:
@hadjieva we'll open another issue for fill mode revision, because they are in different iterations.

PR #3381, opened 2022-02-14
chore: change default border radius in the default theme

epetrow:
Part of: https://github.com/telerik/kendo-themes/issues/3290


Juveniel (approved):

Juveniel:
@joneff, @epetrow should we also use the same metrics for the `Classic` theme, as it is basically based on `Default` ?

joneff (approved):

PR #3384, opened 2022-02-15
chore(html): change default chip rounding to full

Juveniel:
part of: https://github.com/telerik/kendo-themes/issues/3290


Juveniel:
We are not going forward with this change. Seee https://github.com/telerik/web-components-ux/issues/410#issuecomment-1047605090 for more details.

PR #3385, opened 2022-02-15
fix(chiplist): adjust spacing between chips

PaskalevStoyan:


PaskalevStoyan:
This PR will be reopened when the item is planned for the PI.

Issue #3392, opened 2022-02-17
Chip text is cut


Juveniel:
**Describe the bug**

Chip text is partially cut

**To reproduce**
[https://www.telerik.com/kendo-angular-ui/components/buttons/chip/appearance/](
https://www.telerik.com/kendo-angular-ui/components/buttons/chip/appearance/)

**Expected behavior**

The text is not cut at the bottom

**Screenshots**

![image](https://user-images.githubusercontent.com/16972160/154498268-dde9736e-601b-44b6-8f96-f7922f6b218a.png)

**Affected package (please remove the unneeded items)**

* theme-default
* theme-bootstrap
* theme-material
* theme-tasks

**Affected suites (please remove the unneeded items)**

* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**
* All

PR #3397, opened 2022-02-18
fix(chip): fix cut text at bottom of chip

epetrow:
Fixes: https://github.com/telerik/kendo-themes/issues/3392


silviyaboteva (approved):

joneff (approved):
Although there is some displacement in the tests, I believe it's a browser bug in firefox than actual issue

PR #3399, opened 2022-02-21
Themes setup

joneff:
A setup script that:
* installs the dependencies
* initializes husky
* links the packages
* runs bootstrap

**must only be run locally, not on CI**


Juveniel (approved):

Juveniel:
@joneff we should also update the README as well

PR #3404, opened 2022-02-22
Add padding when TextArea is used through directive in Angular

PaskalevStoyan:
When using the TextArea in Angular through the `kendoTextArea` directive, there is no padding applied to the actual `<textarea> element resulting in the text being stuck to the left border. 

image

This is occuring because the actual TextArea component has a wrapper around it (the k-input element) and the actual <textarea> element has the k-input-inner which gives the element padding.

image

As in React, the TextBox (labeled as Input) component doesn't have a wrapper as well, and this scenario is handled with `input.k-input`. 
image

This being said the TextArea directive in Angular also needs padding, being handled with `textarea.k-textarea`.


JoomFX (approved):

packages/default/scss/input/_layout.scss
```diff
@@ -50,13 +50,14 @@
...
+    // Input and Textarea
     .k-input {}
-    input.k-input {
+    input.k-input,
+    textarea.k-textarea {
         padding: $kendo-input-padding-y $kendo-input-padding-x;
     }
 
-
+
```
joneff, line 60: there is trailing space on this line, which means you don't have the correct setup. go to develop, pull and run `npm run setup`. after that, rebase this branch and fix the lint errors.

joneff (changes requested):

joneff (approved):

Issue #3403, opened 2022-02-22
No vertical scrollbar in TextArea


Dimitar-Goshev:
**Describe the regression**
No vertical scrollbar in TextArea with 5.2.0

**To reproduce**

1. Run the following [Dojo snippet](https://dojo.telerik.com/@Dimitar-Goshev/OqAYupOv)

**Expected behavior**
Vertical scrollbar in TextArea as with 5.1.1 - [Dojo snippet](https://dojo.telerik.com/@Dimitar-Goshev/eMileyak)

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material
* theme-classic

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery

**Affected browsers (please remove the unneeded items)**
* All

epetrow:
With 5.2.0 classes with !mportant rule have been added e.g !k-overflow-auto, !k-overflow-x-auto, !k-overflow-y-auto.

Refer to this PR -  https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/utils/_overflow.scss#L55

@telerik/kendo-react-team Could you please add the new class if you haven't done so already?

Dimitar-Goshev:
@epetrow The issue is resolved from Kendo UI for jQuery site.

PR #3407, opened 2022-02-22
Textarea visual test

epetrow:
Related to: https://github.com/telerik/kendo-themes/issues/3403


silviyaboteva:
@epetrow can you please check why so many tests are changed. I couldn't see any differences though.
@joneff is this related to the new Firefox version 97.0.1 from February 17, 2022?

JoomFX:
@silviyaboteva Regarding the changed tests - it is because of the image in the tests (mostly Chip, Avatar, etc).
It does seem to be related to the new FF version.

Juveniel:
I can confirm that the issue is related to the new FF version - this can be verified by examining [the build artifacts here](https://github.com/telerik/kendo-themes/runs/5275823925?check_suite_focus=true) where we can see that the last good commit was running version 96.0. When compared to the CI Visual for the current commit, the version is now 97.0.

I have opened [a new PR ](https://github.com/telerik/kendo-themes/pull/3408)to trigger the image tests. We can rebase the currently opened PRs after it is merged.

silviyaboteva (approved):

joneff (approved):
I would much prefer if this was done by using an `overflow` attribute with initial `auto` value, but the current solution would also suffice.

Issue #3406, opened 2022-02-22
Solid Chip background colors in Material


imasyuko:
There is a slight difference between design and implementation in the background color of solid chips in Material. 

<img width="896" alt="Screenshot 2022-02-22 at 14 16 30" src="https://user-images.githubusercontent.com/81818548/155130605-3445672e-ac0b-4ddb-9187-95701c395e6b.png">

Background colors in normal and hover state in the implementation don't match the ones in [the design file](https://www.figma.com/file/Bun5XXsG0UuzCMz5GyDoHK/Telerik-%26-KENDO-UI-Kit-for-Material-1.9--(UNPUBLISHED)?node-id=190%3A491).

The color progression has to grow as follows:
normal state - $black-8%
hover - $black-16%
focus/selected - $black-24%

PR #3409, opened 2022-02-23
fix(chip): update solid chip background color to match design

epetrow:
Fixes: https://github.com/telerik/kendo-themes/issues/3406


joneff (changes requested):
rebase and change relation to "fixes"

joneff (approved):

Issue #3416, opened 2022-02-28
MultiColumnComboBox has horizontal scroll


Juveniel:
**Describe the bug**

The content of the popup is bigger due to the additional border of the table.

**To reproduce**

[https://demos.telerik.com/kendo-ui/multicolumncombobox/index](https://demos.telerik.com/kendo-ui/multicolumncombobox/index)

**Expected behavior**

The table should have `border-width: 0`.

**Affected package (please remove the unneeded items)**

* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please remove the unneeded items)**

* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React

**Affected browsers (please remove the unneeded items)**

* All

Issue #2801, opened 2021-09-16
Component options for dropdowngrid


joneff:
**Dependencies**

**Component options**

```ts
enum DropdownGridSize { small, medium, large };
enum DropdownGridRounded { small, medium, large, pill };
enum DropdownGridFillMode { solid, flat, outline };

class DropdownGrid {
   size: (null | DropdownGridSize) = DropdownGridSize.medium;
   rounded: (null | DropdownGridRounded) = DropdownGridRounded.medium;
   fillMode: (null | DropdownGridFillMode) = DropdownGridFillMode.solid;
}
```

**Rendering**
* [x] Rendering

```html
<!-- default rendering -->
<span class="k-dropdowngrid k-picker k-picker-md k-rounded-md k-picker-solid">
    <span class="k-input-inner"><span class="k-input-value-text"><span></span>
    <button type="button" class="k-input-button k-button k-icon-button k-button-md k-button-solid k-button-solid-base">
        <span class="k-button-icon k-icon k-i-arrow-s"></span>
    </button>
</span>

<!-- canonical rendering -->
<span class="
    k-dropdowngrid
    k-picker
    k-picker-{size}
    k-rounded-{rounded}
    k-picker-{fillMode}

    {valid && 'k-valid'}
    {invalid && 'k-invalid'}
    {loading && 'k-loading'}
    {required && 'k-required'}
    {disabled && 'k-disabled'}
">
    {showInputIcon && inputIconName !== '' && <span class="k-input-icon k-icon k-i-{inputIconName}"></span>}
    {inputPrefix && <span class="k-input-prefix">...</span>}
    <span class="k-input-inner">
        <span class="k-input-value-icon k-icon k-i-{valueIcon}"><span>
        <span class="k-input-value-text">{valueText}<span>
    </span>
    {inputPrefix && <span class="k-input-suffix">...</span>}
    {showValidationIcon && valid && <span class="k-input-validation-icon k-icon k-i-check"></span>}
    {showValidationIcon && invalid && <span class="k-input-validation-icon k-icon k-i-warning"></span>}
    {showLoadingIcon && loading && <span class="k-icon k-i-loading"></span>}
    {showClearValue && text !== '' && <span class="k-clear-value"><span class="k-icon k-i-x"></span></span>}
    <button type="button" class="k-input-button k-button k-icon-button k-button-{size} k-button-{fillMode} k-button-{fillMode}-base">
[...]

PR #3418, opened 2022-02-28
fix(dropdowngrid): reset data table border width in popup 

Juveniel:
closes: https://github.com/telerik/kendo-themes/issues/3416

Tests Note: we have several separate data-table/table-list tests, thus I have not added a dedicated test for the DDG popup.

Implementation note: this PR will resolve the issue in Kendo jQuery and Angular. However, [the React rendering](https://www.telerik.com/kendo-react-ui/components/dropdowns/multicolumncombobox/) should be updated to closely follow [the spec](https://github.com/telerik/kendo-themes/issues/2801). /cc @silviyaboteva


silviyaboteva (approved):

packages/html/src/dropdowngrid/dropdowngrid.jsx
```diff
@@ -0,0 +1,112 @@
...
+    htmlAttributes.required = required;
+    htmlAttributes.disabled = disabled;
+
+    let dropdownGridClasses = [
+        ownClassName,
+        'k-dropdowngrid'
+    ];
+
+    return (
+        <ComboboxStatic className={dropdownGridClasses} {...props}></ComboboxStatic>
```
joneff, line 56: I will let this one slide. However for R2 we should make this without importing combobox

joneff (approved):

PR #3419, opened 2022-02-28
chore(infra): add gitattributes config

Juveniel:
Set the default file normalization to `LF`, in case contributors do not have `core.autocrlf` set.


TeyaVes (approved):

kspeyanski (commented):
Sorry to jump out of nowhere, but wouldn't this configuration fall-back to whatever the user picked during git installation?:
image

Simply sharing because we had multiple problems related to line-ending in the `kendo-react-private` repository, so we opted to go with the following `.gitattributes` configuration:
https://github.com/telerik/kendo-react-private/blob/develop/.gitattributes

Feel free to steal it!

Juveniel:
@kspeyanski you are correct for the defaults.

In terms of the additional config settings - looking through [the docs](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings) I think that the `text=auto` setting is sufficient for the needs of kendo-themes, as it will leave the images/fonts untouched. Do you have any specific scenario in mind where we need to set the `eol=lf` explicitly?

kspeyanski:
@Juveniel  - Yes!

My biggest pain is with the documentation and more-specifically the `.md` files. As you've probably seen in the `katana` framework we're doing some post-processing of the `.md`'s content through regex matching (take for example the slugs functionality). Having both `\n` (lf) and `\r\n` (cr) makes it confusing the least.

This being said it's not a big deal to leave them as the front-end team prefers, but I was wondering if we can make the EOL a little bit stricter.

Juveniel:
@kspeyanski I have added the relevant eol rules

silviyaboteva (approved):

joneff (approved):

Issue #3421, opened 2022-02-28
The Floating Label should conditionally have pointer-events:none


Xizario:
Currently there is [rule](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/floating-label/_layout.scss#L33):
```css
.k-floating-label-container > .k-label{
pointer-events: none;
}
```
That prevents the label to be clicked. And from accessibility point of view the label should be clickable and the focus should move to the input that is associated with it.
However when there is text in the input with floating label, and the label stays above the input, clicking on the label does nothing. Can be reproduced here: just type some text into the input https://www.telerik.com/kendo-react-ui/components/labels/floating-label/ 

If we just remove the rule, then input can't be focused on mouseDown, so removing it is not an option.

My suggestion to move the `pointer-events: none;` into the [.k-floating-label-container.k-state-empty > .k-label](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/floating-label/_layout.scss#L43).

PR #3429, opened 2022-03-01
fix(floating-label): allow label to be clickable in order to focus as…

epetrow:
…sociated input

Fixes: https://github.com/telerik/kendo-themes/issues/3421


packages/default/scss/floating-label/_layout.scss
```diff
@@ -75,6 +75,7 @@
                 > .k-label {
                     left: auto;
                     right: $floating-label-offset-x;
+                    pointer-events: none;
```
Juveniel, line 78: This is redundant, as the label will receive `pointer-events: none` from the `.k-state-empty > .k-label` selector on lines 40-45. Thus, this line can be removed.

Juveniel (changes requested):

Juveniel (approved):

PaskalevStoyan (approved):

joneff (approved):

Issue #3412, opened 2022-02-25
Update floating label reference rendering with more examples


Stamo-Gochev:
The reference rendering of the floating label:
https://github.com/telerik/kendo-themes/blob/develop/tests/visual/src/misc/floating-label.html

can be updated with more examples about focused, disabled and invalid states.

ng2 
- `k-state-invalid` is rendered for invalid state -https://github.com/telerik/kendo-angular-label/blob/develop/src/floating-label/floating-label.component.ts#L92 - the text of the label is not red

react
- `k-state-error` is rendered for invalid state - https://github.com/telerik/kendo-react-private/blob/develop/packages/labels/src/FloatingLabel.tsx#L156: - makes the label text red
- `k-state-disabled` is rendered for disabled state - https://github.com/telerik/kendo-react-private/blob/develop/packages/labels/src/FloatingLabel.tsx#L149-L157

Started as part of:
https://teams.microsoft.com/l/message/19:1f8c4e9c40004fbf8fe23af65fe52970@thread.tacv2/1644918556215?tenantId=db266a67-cbe0-4d26-ae1a-d0581fe03535&groupId=d825ddf5-c7c7-4f9b-8422-1d8f39041b8d&parentMessageId=1644918556215&teamName=Kendo%20Front%20End%20Team&channelName=General&createdTime=1644918556215

but wasn't fully finished.

/cc @joneff

PR #3438, opened 2022-03-04
test(floating-label): add examples for focus, hover, disabled and inv…

epetrow:
…alid state

Closes: https://github.com/telerik/kendo-themes/issues/3412


tsvetomir:
Please consider organizing these tests under `tests/visual/src`. This will make it easier for product teams to add markup snapshots for regression testing.

silviyaboteva (approved):

joneff (approved):

Stamo-Gochev:
Any reason we decided to drop the "red" text color of the label when its associated input is invalid?

Here is the result from the updated [rendering](https://github.com/telerik/kendo-themes/blob/9c6ffad73cdf1aeca831c602b6d3150aa50bfdb0/tests/visual/src/misc/floating-label.html#L159-L183):
image

There seems to be some variance over this, but when using the label with form field component, it has red text when the value is invalid in ng2:
https://www.telerik.com/kendo-angular-ui/components/forms/forms-guideline/#toc-floating-labels
image

A similar thing happens in react:
https://www.telerik.com/kendo-react-ui/components/inputs/input/forms/#toc-setting-validation-requirements
https://www.telerik.com/kendo-react-ui/components/inputs/numerictextbox/forms/#toc-enforcing-the-validity-state

This is also the case with the blazor [Form](https://demos.telerik.com/blazor-ui/form/validation):
image

This makes me think we are more leaning towards red label text when the input is invalid.

Stamo-Gochev:
Any thoguths on https://github.com/telerik/kendo-themes/pull/3438#issuecomment-1066444797? /cc @telerik/kendo-front-end

epetrow:
The "red" text color is applied to the label only when it is used within a form field: `k-form-error, .k-text-error, .k-form-field-error .k-label { @include fill( $color: $error ); }`

Regarding the second question, this is expected and both classes produce the same behavior.

@Juveniel please, correct me if I am wrong

PR #3439, opened 2022-03-06
fix(calendar): update claendar height for different views

NikoPenev21:
telerik/kendo-ui-core/issues/6576


JoomFX:
With this change each calendar view will have a different height.
I believe that we want to have the same calendar height for all views so it does not "jump" when the view is changed.

@joneff @Juveniel

NikoPenev21:
Not a bug. The calendar has a min-height set so the height is persistent in all views.
For custom scenarios, we can override the $calendar-view-height variable to achieve the desired result.

Issue #3446, opened 2022-03-09
Grid header sort icon spacing does not work


JoomFX:
**Describe the regression**
We have [this selector](https://github.com/telerik/kendo-themes/blob/e032f8763be32e1de0e9c06987b0aa9eedf79587/packages/default/scss/grid/_layout.scss#L658) in the styles but it is no longer valid since there is an additional element (`.k-cell-inner`) in the markup.

**To reproduce**

1. Go to the [Grid Sorting demo](https://www.telerik.com/kendo-angular-ui/components/grid/sorting/)
1. Inspect the markup

**Expected behavior**
The correct selector should be used.

**Screenshots**
![markup](https://user-images.githubusercontent.com/7753940/157406639-990fb38f-e912-45f2-8952-24d8630cd0f7.png)

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**
* All

**Build system information (please remove the unneeded items)**
* Not Applicable

PR #3447, opened 2022-03-09
fix(grid): incorrect header sorting icon selector

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/3446


PaskalevStoyan (approved):

TeyaVes:
I see from the visual tests that after the fix the sorting icon goes quite far from the text in the Bootstrap theme. I think we should change the spacing icon variable and decrease the margin. @JoomFX , what do you think?

Issue #3445, opened 2022-03-08
MultiSelect chips are not styled properly in 5.2.0 dark swatches


ntacheva:
**Describe the bug**

In 5.2.0 MultiSelect chips are not visible in the dark swatches of Default and Bootstrap. They have black background-color and color and are not readable. In the Material dark swatches, the chips' color is white and they can be read but their background is black (could be by design, pointing it out, so it can be checked).

**To reproduce**

1. Go to [MultiSelect - Overview demo](https://demos.telerik.com/blazor-ui/multiselect/overview) https://demos.telerik.com/kendo-ui/multiselect/index?autoRun=true&theme=default-main-dark
1. Click 'Change Theme' dropdown and select Default/Bootstrap Main Dark
1. Select something from the MultiSelect dropdown
1. Chips are not visible

**Expected behavior**

MultiSelect chips should be visible in dark swatches.

**Screenshots**

**Current Default Main Dark (5.2.0)**
![image](https://user-images.githubusercontent.com/73842592/157206970-8ffedd6f-381b-4b3b-bf02-a5031517cf63.png)

**Previous Default Main Dark (4.43.0)**
![image](https://user-images.githubusercontent.com/73842592/157206585-4677122e-ff23-42f6-b332-b9521344a3a5.png)

**Affected package**

* theme-default
* theme-bootstrap
* theme-material

**Affected suites**

* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**

* All

PR #3449, opened 2022-03-09
fix(chip): text color in dark swatches

TeyaVes:
fixes: https://github.com/telerik/kendo-themes/issues/3445


joneff (approved):
I think chips must ultimately no be tied to buttons, but rather base.

PR #3466, opened 2022-03-16
Remove Unused Dependencies

PaitoAnderson:
I encountered an error after upgrading to macOS 12.3 when running `npm install` because this project depends on `sass-variable-parser` and that depends on `node-sass@4.9.3` which expects `python2` to be installed.

This resolves that issue by removing `sass-variable-parser` as a dependency, I didn't see `@joneff/baka` used anywhere either so I remove it as well.

This also changes the package-lock.json format to v2, let me know if that's a problem, I couldn't find a way to preserve the lockfileVersion version.


joneff:
There are a couple of things preventing us from merging directly, so I am creating a temp branch.

You are on point with the dependencies. prior 80e37c4409d673e006d59ed29845c6aa2b5b713c we used both sass-variables-parser and baka.

PaitoAnderson:
Thanks for the quick merge @joneff!

PR #3489, opened 2022-03-28
feat(utils): add aspect-ratio utility classes

TeyaVes:


Juveniel:
I would suggest adding important utils as well (e.g like in the [flex util](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/utils/_flex.scss#L10-L11)).

joneff (approved):

PR #3496, opened 2022-03-29
Colorpalette html helper

Juveniel:


JoomFX (approved):

packages/html/lib/jsx-runtime.js
```diff
@@ -49,7 +49,11 @@ const attrMap = {
     leafclassname: 'leafClassName',
 
     // Template
-    contenttemplate: 'contentTemplate'
+    contenttemplate: 'contentTemplate',
+
+    // ColorPalette
+    palette: 'palette',
```
silviyaboteva, line 55: this might be not needed as `palette` is one word which output attribute is the same.

packages/html/src/colorpalette/colorpalette-tile.jsx
```diff
@@ -0,0 +1,74 @@
...
+    let styles = {
+        'background-color': color,
+        'width': tileSize ? tileSize + "px" : '',
+        'height': tileSize ? tileSize + "px" : ''
+    };
+
+    let ariaAttr = aria
+        ? {}
+        : {};
+
```
silviyaboteva, line 45: do we need to add the legacy rendering as we have in other HTML helpers?

silviyaboteva (approved):

Issue #3523, opened 2022-04-06
SplitButton styles for the popup are not loaded in the ThemeBuilder


Stamo-Gochev:
**Describe the bug**
The popup styles of the SplitButton component are not correctly loaded in the ThemeBuilder, when the SplitButton is the only selected component.

**To reproduce**

1. Go to 
https://winslaveblazor.dev.progress.com/staging-themebuilder/kendo-angular-ui or
https://winslaveblazor.dev.progress.com/staging-themebuilder/blazor-ui
1. Start modifying a theme and select "SplitButton" as the only component

**Expected behavior**
The popup of the SplitButton should be displayed as expected

**Screenshots**
Actual:
![image](https://user-images.githubusercontent.com/1857705/161962493-be9ebef8-b588-4956-9161-909d8bb51693.png)

Expected:
![image](https://user-images.githubusercontent.com/1857705/161962522-f79988bd-3c3b-45ce-8d10-66c579c77e59.png)

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material
* theme-tasks

**Affected suites (please remove the unneeded items)**
* all

**Affected browsers (please remove the unneeded items)**
* All

PR #3524, opened 2022-04-06
fix(button): add missing menu dependency

Juveniel:
fixes: #3523


Juveniel:
Putting this on hold for now, as we need to extract menu list as a separate package to fix the circular dependency issue...

joneff (approved):

Issue #3531, opened 2022-04-08
Focused style is not distinguished in selected toggle buttons in toolbar popup


KristianMariyanov:
**Describe the bug**
When there is overflowing selected toggle buttons in the toolbar popup, the focused styles are not visible.
It's also valid for button group toggle buttons

**To reproduce**
1. Go to this example: https://winslaveblazor.dev.progress.com/blazor-playground-server/toolbar/adaptive
1. Select all of the buttons
1. Resize the screen so some of the buttons in the second toolbar (Toggle Button) are hidden
1. Click on the overflow icon
1. Now the first button is focused and selected, but focused styles are not visible.

**Expected behavior**
Focused styles can be applied with selected styles

**Screenshots**
The Redo 2 button is focused:
![image](https://user-images.githubusercontent.com/6886919/162422822-0ed7e173-0988-421f-92a2-23e522c9a6d8.png)

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please remove the unneeded items)**
* Telerik UI for Blazor
* And all other which uses the referenced rendering

**Affected browsers (please remove the unneeded items)**
* All

**Build system information (please remove the unneeded items)**
* Not Applicable

PR #3546, opened 2022-04-12
Menu item focus state styles

TeyaVes:
fixes: https://github.com/telerik/kendo-themes/issues/3531


Juveniel (approved):

silviyaboteva (commented):
Can you ensure that keyboard navigation will work in this case? As I remember, I made a fix recently reported by the Blazor team: https://github.com/telerik/kendo-themes/commit/ebbadf5025d7d0903d35ef9384143894ebfd4d25

silviyaboteva (approved):

TeyaVes:
@silviyaboteva , the class and the focus state will remain on the li element, it's just the style that would be applied to the inner element, thus, it shouldn't affect the accessibility

Issue #3289, opened 2022-01-26
PanelBar -> scss variable to define item border


meriturva:
See here:

https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/panelbar/_layout.scss#L30

```
> .k-item,
> .k-panelbar-header {
            // TODO
            border-width: 0;
            border-style: solid;
            border-color: inherit;
            display: block;
```

is it possible to have a variable to set the border to `none`?
Actually my customization is:

```
.k-panelbar {
  > .k-item {
      border-style: none;
```

PR #3556, opened 2022-04-20
Expose item border width and style variables

NikoPenev21:
closes : telerik/kendo-themes#3289


joneff (commented):
@NikoPenev21 the correct way to reference an issue is `#123` or `telerik/kendo-themes#123`. 

Name of the pull request should not be based on the commit i.e. the `fix` / `feat` / `chore` prefix should be removed.

Even though this PR closes the actual issue and is not the issue, it should at least have a label which component does it concern

joneff (approved):

joneff:
@Juveniel  lgtm.

Juveniel (approved):

PR #3594, opened 2022-05-04
fix: provide compatibility for both old and new state selectors

kspeyanski:
Related to: https://github.com/telerik/kendo-react-private/pull/1734

The following PR introduces changes, which aim to support the migration to the new state syntax, introduced with `v5`.

So far, we've had the new `k-*` (instead of `k-state-*`) syntax only for ~half of the components. With this PR we want to apply it for All of the components to enable easier transition for the Kendo suits.

Currently, we're applying rendering updates on batches, depending on which components have a reference rendering.
The update consist of:
- Restructuring the DOM
- Migrating `k-state-*` to `k-*`

The second part could be easily applied through `Find And Replace` in Ng/R/V suits, given no per-component preliminary check is required. 

For old-new state classes mapping, i've used the following map:

| Old | New |
|-|-|
|  ["k-state-focus"]| "k-focus" | 
|  ["k-state-focused"]| "k-focus" | 
|  ["k-state-hover"]| "k-hover" | 
|  ["k-state-hovered"]| "k-hover" | 
|  ["k-state-disabled"]| "k-disabled" | 
|  ["k-state-invalid"]| "k-invalid" | 
|  ["k-state-selected"]| "k-selected" | 
|  ["k-state-active"]| "k-active" |


packages/material/scss/floating-label/_theme.scss
```diff
@@ -10,6 +10,7 @@
         .k-invalid + .k-label,
         &.k-state-invalid + .k-label,
         .k-state-invalid + .k-label,
+        .k-invalid + .k-label,
```
Xizario, line 13: duplicate line

packages/default/scss/splitter/_theme.scss
```diff
@@ -15,14 +15,17 @@
...
+    .k-splitbar.k-hover,
     .k-splitbar-horizontal-hover,
     .k-splitbar-vertical-hover {
         color: $splitbar-hover-text;
         background-color: $splitbar-hover-bg;
     }
     .k-splitbar:focus,
     .k-splitbar.k-state-focus,
-    .k-splitbar.k-state-focused {
+    .k-splitbar.k-focus,
```
Xizario, line 26: duplicate

packages/material/scss/floating-label/_theme.scss
kspeyanski: #wontfix

packages/default/scss/splitter/_theme.scss
kspeyanski: #wontfix

Xizario (approved):

TeyaVes (approved):

packages/bootstrap/scss/breadcrumb/_layout.scss
```diff
@@ -9,7 +9,8 @@
         text-decoration: $link-decoration;
 
         &:hover,
-        &.k-state-hover {
+        &.k-state-hover ,
```
grozevg, line 12: Remove the space before the comma.

Issue #3601, opened 2022-05-11
Flat Input Button does not look good when hovered


epetrow:
**Describe the bug**

Currently when hovering on a button inside a focused flat input there is a visible gap between the top border of the input and the button.

**To reproduce**

See screenshot below.

![image](https://user-images.githubusercontent.com/95700708/167846805-b3652ad8-ca96-4f50-a13a-014f3cb189e4.png)

**Expected behavior**

There should be no gap when hovering on a button inside a focused flat input.

**Affected package:**
* theme-default
* theme-bootstrap

PR #3602, opened 2022-05-11
fix(input): set correct  border-width for flat inputs and pickers

epetrow:
Closes: https://github.com/telerik/kendo-themes/issues/3601


TeyaVes:
My only concern here is that the inputs and pickers in flat mode become shorter and there would be a slight vertical misplacement with 1px if positioned next to a button for example. However, I am not sure if this would be a serious issue

epetrow:
> My only concern here is that the inputs and pickers in flat mode become shorter and there would be a slight vertical misplacement with 1px if positioned next to a button for example. However, I am not sure if this would be a serious issue

I don't think there is a noticeable difference, but I might be mistaken.

image

image

joneff (changes requested):
This will make flat inputs a tad smaller. Please think of a different approach.

epetrow:
Closing this for now since there is no obvious fix. We'll revisit this in the future.

Issue #3560, opened 2022-04-22
[Material] [Floating Label] Floating label doesn't behave correctly when associated input is focused or dirty


szyszak:
**Describe the bug:**
Material Theme for Kendo Angular doesn't work well for inputs with appearance set to "outline" combined with floating labels.
Focused inputs with appearance set to "outline" have additional bottom border line stretching beyond the outline. 

**To reproduce:**
Example on Stackblitz:
https://stackblitz.com/edit/angular-cwkjca-phpltk

Angular Material examples for comparison:
https://material.angular.io/components/form-field/overview#form-field-appearance-variants=

**Expected behavior:**
The floating label should reposition directly upwards instead up and to the left. 
It should also stop in the middle of the top border instead of going beyond.
Focused state should not add another border at the bottom.

**Screenshots:**
### Angular Material:

Pristine
![material-pristine](https://user-images.githubusercontent.com/16250531/164793270-fa694ca3-dc9e-4ce4-944d-ed7e70566799.JPG)

Focused
![material-focus](https://user-images.githubusercontent.com/16250531/164793291-d1b38539-54a5-463c-9702-fac00f88f4a2.JPG)

Dirty
![material-dirty](https://user-images.githubusercontent.com/16250531/164793314-564f97dc-84ea-49ad-8d5c-054f693bf2fb.JPG)

### Kendo Angular:

Pristine
![kendo-pristine](https://user-images.githubusercontent.com/16250531/164793327-39596751-4ecc-4bfb-8637-5d6889bda674.JPG)

Focused
![kendo-focus](https://user-images.githubusercontent.com/16250531/164793335-99b27936-c3ce-4222-bbfb-e7e7ac0000f6.JPG)

Dirty
![kendo-dirty](https://user-images.githubusercontent.com/16250531/164793348-9f99e008-926b-49f5-9ed3-a23777190c0a.JPG)

**Affected package:**
* theme-material

**Affected suites:**
* Kendo UI for Angular

**Affected browsers:**
* All

**Build system information:**
* Not Applicable

Issue #3600, opened 2022-05-11
Material focus styles for inputs


TeyaVes:
Currently, in the material theme we have ripple effect on the floating label inputs in flat mode. However, we should add similar focus styles in the rest of the modes and across all inputs. 

Material reference: [https://material.io/components/text-fields](https://material.io/components/text-fields)

TeyaVes:
Already implemented in [Kendo Material theme v. 11.0.0](https://github.com/telerik/kendo-themes/releases/tag/v11.0.0)

PR #3603, opened 2022-05-12
Ensure Floating label behaves correctly when associated input is focused or dirty

epetrow:
Closes https://github.com/telerik/kendo-themes/issues/3560


TeyaVes:
related to : https://github.com/telerik/kendo-themes/issues/3600

TeyaVes:
If we are removing the input-ripple mixin and moving the styles to the input/layout file, I think that we should move the theme styles (e.g bg-color, etc) to the themes file.

TeyaVes (approved):

joneff (approved):

PR #3613, opened 2022-05-19
fix(slider): add React specific styles for slider layout

kikostadinov:
React issue:
https://github.com/telerik/kendo-react/issues/1301


TeyaVes:
@kikostadinov I don't think that using attributes as selectors is what we want. We should rather add a class and use it for adding the specific styles. In this way we would be able to add a visual test for the following functionality and add it easily in the rest of the Kendo suites if needed. In the general case, we add product-specific styles only where this cannot be avoided.

kikostadinov:
Add a visual test once https://github.com/telerik/kendo-themes/pull/3557 is merged.

TeyaVes (changes requested):

packages/html/src/slider/tests/slider-tick-absolute-position.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en" class="k-no-animations">
+<head>
+    <title>Button Size</title>
```
TeyaVes, line 4: Change title here

TeyaVes (changes requested):

packages/html/src/slider/tests/slider-tick-absolute-position.tsx
```diff
@@ -0,0 +1,102 @@
...
+                            <a href="#" className="k-draghandle k-draghandle-end" style={{ left: '33%' }}></a>
+                        </div>
+                    </div>
+                    <Button icon="arrow-e" rounded="full" className="k-button-increase"></Button>
+                </div>
+            </section>
+
+            <section>
+                <div className="k-widget k-slider k-slider-vertical">
+                    <Button icon="arrow-w" rounded="full" className="k-button-decrease"></Button>
```
TeyaVes, line 69: change icons in vertical slider to arrow-n and arrow-s

packages/html/src/slider/tests/slider-tick-absolute-position.tsx
```diff
@@ -0,0 +1,102 @@
...
+                                <span className="k-label">75</span>
+                            </li>
+
+                            <li className="k-tick k-tick-large k-tick-horizontal" style={{ zIndex: 1, position: 'absolute', left: '100%' }}>
+                                <span className="k-label">100</span>
+                            </li>
+                        </ul>
+                        <div className="k-slider-track">
+                            <div className="k-slider-selection" style={{ width: '33%' }}></div>
+                            <a href="#" className="k-draghandle k-draghandle-end" style={{ left: '33%' }}></a>
```
TeyaVes, line 60: Check whether the drag handle needs z-index as well, cause right now it's displayed under the ticks. It should be over them.

TeyaVes (approved):
You can now squash the commits into one, just make sure to drop the visual previews commits- a new one will be automatically generated after the push

joneff (approved):

Issue #2946, opened 2021-11-02
Drawer hierarchical items


Raisolution:
FrontEnd item related to:
- AHA item https://progresssoftware.aha.io/features/KUING2-293
- Design https://github.com/telerik/kendo-ux-private/issues/292#issuecomment-813433893
- Planning item https://github.com/telerik/web-components-planning/issues/10

The idea is to display items on subsequent levels as nested. Similar implementation is available in the PanelBar component through the `k-level-x` CSS classes.

Planned Release date R3 2022 Iteration 1

Xizario:
Is this a followup to https://github.com/telerik/kendo-themes/pull/2423 ? 
From my understanding you want to move the styles from this demo https://www.telerik.com/kendo-angular-ui/components/layout/drawer/hierarchical-drawer/#toc-using-flat-data to the theme itself. Is this correct?

Raisolution:
> Is this correct?

The demo has only a single CSS class to _fake_ level 2 items. But pretty much yes - we will need `k-level-x` CSS classes for multiple levels nesting as in the PanelBar.

Raisolution:
> Planned Release date R1 2022 Iteration 3

Updated the product increment to R3 2022 Iteration 1.

PR #3620, opened 2022-05-25
Add styles for hierarchical drawer items

PaskalevStoyan:
closes: https://github.com/telerik/kendo-themes/issues/2946
related to: https://github.com/telerik/k2/issues/471

~~TBD: When the items get too close to the right border of the Drawer, they move on two levels. Q: should they stay as they are now, or the text should overflow?~~ - discussed and delivered 

image


packages/html/src/drawer/tests/drawer-hierarchical-item-levels.tsx
```diff
@@ -0,0 +1,136 @@
...
+                                    <span className="k-item-text">About Kendo UI</span>
+                                </li>
+                                <li className="k-drawer-item k-state-selected k-level-2">
+                                    <Icon name="palette" />
+                                    <span className="k-item-text">Supported Themes</span>
+                                </li>
+                                <li className="k-drawer-item k-level-0">
+                                    <Icon name="zoom" />
+                                    <span className="k-item-text">Overview</span>
+                                    <Icon name="arrow-chevron-up" className="k-drawer-toggle" />
```
joneff, line 39: Icons must not be clickable.

Add a wrapper around it with the class `k-drawer-toggle`

joneff (changes requested):

packages/html/src/drawer/tests/drawer-hierarchical-item-levels.tsx
PaskalevStoyan: We have discussed this internally with the Angular team, and currently there is no specific need to add the span wrapper at this point, as all events are attached on the list item and not on the icon itself.

As there is a little time until the code freeze, let's merge this PR as it is and if there is a need to wrap the Icon in a wrapper, we will do so in the future. 
@Raisolution , if i'm missing something, please feel free to fill the gaps.

cc: @Juveniel

joneff (approved):

PR #3626, opened 2022-05-30
Combobox structure

joneff:
Note: I moved button structure to `html/meta/button` and changed `ref` to `$ref` (which i might change again if it's an issue).

@rkirilov @tsvetomir

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -20,7 +20,7 @@ const paths = {
     sass: {
         all: "packages/*/scss/**/*.scss",
         assets: "packages/*/scss/**/*.{png,gif,ttf,woff}",
-        themes: "packages/!(html)",
+        themes: "packages/!(html|html-spec)",
         theme: "scss/all.scss",
         swatches: "lib/swatches/*.json",
         inline: "dist/all.scss",
--- a/packages/html-spec/package-lock.json
+++ b/packages/html-spec/package-lock.json
@@ -1,0 +1,6 @@
+{
+  "name": "@progress/kendo-html-spec",
+  "version": "5.8.2-dev.2",
+  "lockfileVersion": 1,
+  "requires": true
+}
--- a/packages/html-spec/package.json
+++ b/packages/html-spec/package.json
@@ -1,0 +1,33 @@
+{
+  "name": "@progress/kendo-html-spec",
+  "description": "Specification for Kendo UI component rendering in JSON format",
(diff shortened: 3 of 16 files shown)


packages/html/meta/button.json
```diff
@@ -198,14 +198,14 @@
         ],
         "children": [
             {
-                "name": "icon",
-                "ref": "icon",
+                "name": "button-icon",
+                "$ref": "icon",
```
tsvetomir, line 202: Nice tweak, makes it really easy to spot.

tsvetomir (approved):

rkirilov:
@joneff we will discuss the structure with the team and will get back to you next week

joneff:
@rkirilov @tsvetomir @kvelikov @dtopuzov  I've updated the json files. See if you like them now better than you did before.

rkirilov:
> @rkirilov @tsvetomir @kvelikov @dtopuzov I've updated the json files. See if you like them now better than you did before.

Looks good to me, but let's wait for @kvelikov review

Juveniel (approved):

joneff:
@kvelikov any thoughts?

kvelikov:
> @kvelikov any thoughts?

I think the proposed structure covers all cases so we're good.

dtopuzov:
Some comments on the testing side.

We plan to have something like `verifyRendering()` function in https://github.com/telerik/kendo-e2e/.
Here are some initial thoughts on the implementation:
- We need spec as npm package
- We will NOT refer to this package in kendo-e2e because we want to allow teams to be compatible with different versions of the spec at different times, we also do not want to release new kendo-e2e just because we have a bug fix in the spec.
- Each product team will refer to the spec [as we do it for a11y spec now](https://github.com/telerik/kendo/blob/master/package.json#L20)
- This means verifyRendering should take spec from the product repo and we need to pass the spec as param to `verifyRendering`
- Currently, we [pass a11y spec as json object ](https://github.com/telerik/kendo-e2e/blob/master/src/a11y/a11y-wct-spec.ts#L21)parameter to `getWCTSpecViolations` and we want something similar for rendering spec
- It will be great if we can bundle all rendering specs in single json file before we publish the npm pacakge (generate single json spec build time)

joneff:
@dtopuzov, @tsvetomir -- for your viewing pleasure: `@progress/kendo-html-spec`.

There are zero dependencies and also a very thin exposure around json files.

PR #3653, opened 2022-06-13
chore(multiviewcalendar): disable outline of focused element telerik/kendo#15502

veselints:
The MultiViewCalendar `k-calendar-view` element should be the focusable element of the widget. Its outline, however, should not be visible as the active date is visually the focused element (using `aria-activedescendant`).


epetrow (approved):

joneff:
on a serious note, we cannot know which element would be desired for the different suites. to facilitate development, we can have `k-no-outline` or `k-outline-0` that deals with said behaviour.

veselints:
The accessibility spec (in progress) requires focus to stay on the `k-calendar-view` element. Hence, all suites will have to comply with that requirement.

joneff:
That wasn't a remark to you as much as it was a remark to us -- for certain scenarios where linear behaviour is require i.e. hide outline on a specific element, such class name would make it much easier.

it will also help determine if the various suites have consciously stepped towards implementing the spec, rather than getting some things "for free".

Issue #3662, opened 2022-06-21
Grid .k-cell-inner element spans outside its parent cell


TeyaVes:
**Describe the bug**
In Kendo Grid .k-cell-inner elements spans wider than th resulting in cutout header cell text if the column is aligned to the right. It is most visible in the Bootstrap theme

**To reproduce**

1. Go to [https://demos.telerik.com/blazor-ui-dev/grid/cell-formatting](https://demos.telerik.com/blazor-ui-dev/grid/cell-formatting)
1. Change theme to Bootstrap
1. See the Price column

**Expected behavior**
.k-cell-inner should be as wide as its parent th

**Screenshots**

![MicrosoftTeams-image (2)](https://user-images.githubusercontent.com/45764088/174817662-1da4d8ae-bb05-4e74-ba96-7c08308b57a3.png)

PR #3664, opened 2022-06-22
Fix grid .k-cell-inner width

TeyaVes:
fixes: https://github.com/telerik/kendo-themes/issues/3662


Juveniel:
@TeyaVes nice cleanup of the code!

I noticed that the sort order element overlaps with the arrow icon (can be seen most clearly in the material theme screenshots):

image


Perhaps we should position the element to avoid this. Could you take a look ?

Juveniel (approved):

Issue #2954, opened 2021-11-08
Show PivotGridV2 scrollbar only when it is needed.


MilenaCh:
Currently, PivotGridV2 scrollbar is always visible. This behavior should be changed and scrollbar should show only when it is needed.

magdalenaan:
To hide the scrollbar when it is not necessary, we needs implementation in all suites, namely add width to col elements.

PR #3672, opened 2022-06-23
fix: hide scrollbars when are not needed #2954

magdalenaan:
related https://github.com/telerik/kendo-themes/issues/2954


TeyaVes (commented):
@magdalenaan, I am concerned this may cause certain misalignments as in the following case:

[https://dojo.telerik.com/arUbETef](https://dojo.telerik.com/arUbETef) and toggle 'ALL GEOGRAPHIES'

pivotgrid

PR #3679, opened 2022-06-28
Fluent icons

TeyaVes:
related to: https://github.com/telerik/kendo-themes-private/issues/2

needs: https://github.com/telerik/kendo-icons/pull/17


Juveniel (approved):
I think we should extract the flip and other layout styles in kendo-icons in the future, as the package can be used separately from the themes.

PR #3682, opened 2022-06-30
chore: add font and icons core variables

magdalenaan:


packages/fluent/scss/core/_variables.scss
```diff
@@ -12,6 +12,50 @@ $use-input-spinner-width: true !default;
...
+    xs: 4px,
+    sm: 8px,
+    md: 12px,
+    lg: 16px,
+    xl: 24px,
+    thin: 2px,
+    hair: 1px
+) !default;
+
+// Icons
```
TeyaVes, line 55: We don't really need this as the vars will be added in the icons module

packages/fluent/scss/core/_variables.scss
```diff
@@ -12,6 +12,50 @@ $use-input-spinner-width: true !default;
 $use-input-spinner-icon-offset: false !default;
 $use-picker-select-width: true !default;
 
+// Spacing
+$kendo-spacer: 1em !default;
+$kendo-spacer-x: $kendo-spacer !default;
+$kendo-spacer-y: $kendo-spacer !default;
+
+$kendo-spacing: (
```
TeyaVes, line 20: We have some paddings (buttons, inputs, etc) that are 6px. Maybe we should add a spacing with 2px step instead of 4px

packages/fluent/scss/core/_variables.scss
```diff
@@ -12,6 +12,50 @@ $use-input-spinner-width: true !default;
 $use-input-spinner-icon-offset: false !default;
 $use-picker-select-width: true !default;
 
+// Spacing
+$kendo-spacer: 1em !default;
```
TeyaVes, line 16: I'm not really sure if we need the spacer. I guess we can add it once we decide that we needed in it for any of the components.

TeyaVes (changes requested):

TeyaVes (approved):

PR #3687, opened 2022-06-30
Expose width property and set default value

NikoPenev21:
closes: telerik/kendo-themes/issues/3576


TeyaVes (approved):

Juveniel:
The commit type should be `fix`, as we would like this to become available in the change log.

PR #3702, opened 2022-07-08
Fluent input

epetrow:
related to: https://github.com/telerik/kendo-themes-private/issues/2


packages/fluent/scss/core/color-system/_theme.scss
```diff
@@ -31,6 +31,11 @@ $kendo-box-shadow-depth-3: 0px 1.2px 3.6px rgba( $black, 0.1 ), 0px 6.4px 14.4px
 /// Equivalent to fluent depth 64.
 $kendo-box-shadow-depth-4: 0px 4.8px 14.4px rgba( $black, 0.18 ), 0px 25.6px 57.6px rgba( $black, 0.22 ) !default;
 
+// Validator
+$kendo-invalid-bg: null !default;
+$kendo-invalid-text: get-theme-color(error, 190) !default;
+$kendo-invalid-border: get-theme-color(error, 190) !default;
```
TeyaVes, line 37: I was thinking that here we should use **get-theme-color-var** function instead of **get-theme-color**. In this way, if the CSS palette var is changed by the customer, the invalid var will be changed as well. I'll update the rest of the vars in the file to use the css vars as well.  What do you think?

packages/fluent/scss/input/_variables.scss
```diff
@@ -0,0 +1,178 @@
...
+        line-height: $kendo-input-line-height-lg,
+        icon-size: calc( #{$kendo-icon-size} + #{$kendo-input-padding-y-lg * 2} ),
+        button-padding-x: $kendo-input-padding-y-lg,
+        button-padding-y: $kendo-input-padding-y-lg
+    )
+) !default;
+
+$kendo-input-bg: var( --kendo-component-bg, initial ) !default;
+$kendo-input-text: var( --kendo-body-text, initial ) !default;
+$kendo-input-border: get-theme-color( neutral, 130 ) !default;
```
TeyaVes, line 74: use get-theme-color-var

TeyaVes (changes requested):

Juveniel (approved):
lgtm

packages/fluent/scss/input/_layout.scss
```diff
@@ -0,0 +1,439 @@
...
+            box-shadow: none;
+        }
+
+        // fix for Safari
+        & > * {
+            margin: 0;
+        }
+
+        &::placeholder {
+            color: var( --kendo-input-placeholder-text, inherit );
```
TeyaVes, line 55: shouldn't we use the sass var as a fallback here and below?

packages/fluent/scss/input/_variables.scss
```diff
@@ -0,0 +1,617 @@
...
+// Width of the border around the input when focused.
+$kendo-input-focus-border-width: 2px !default;
+
+/// Height of he border around the input.
+/// @group input
+$kendo-input-border-height: calc( $kendo-input-border-width * 2 ) !default;
+
+/// Border radius of the input.
+/// @group input
+$kendo-input-border-radius: 2px !default;
```
TeyaVes, line 23: I think here we should rather use var( --kendo-border-radius-md, 0 ) , so that if the global border-radius is changed, the input's border radius will be changed as well.

PR #3746, opened 2022-07-26
Card and popover callouts

TeyaVes:
The rendering of the Popover needs to be updated. An additional k-popover-inner class needs to be added around k-popover-head, k-popover-body, k-popover-actions in case there is a callout added to the popover (similar to the card). Correct rendering

```
<div class="k-popover">
    <div class="k-popover-callout k-callout-s"></div>
    <div class="k-popover-inner">
        <div class="k-popover-header">Title</div>
        <div class="k-popover-body">Body content</div>
        <div class="k-popover-actions k-actions k-actions-stretched">
            <button type="button" class="k-button k-button-md k-rounded-md k-button-flat k-button-flat-base">
                <span class="k-button-text">Action</span>
            </button>
            <button type="button" class="k-button k-button-md k-rounded-md k-button-flat k-button-flat-primary">
                <span class="k-button-text">Primary</span>
            </button>
        </div>
    </div>
</div>
```


Raisolution:
@TeyaVes will this change cause UI discrepancies if a customer is using:
- the updated rendering with the `k-popover-inner` and a current version of the theme and vice versa
- the current rendering with the upcoming R3 theme

In other words will it force clients to update the component and the theme at the same theme, or is it implemented in a backwards compatible way?

TeyaVes:
@Raisolution , with the current styles and new rendering - the popover will not be broken, the only visible difference would be a different border-color of the popover actions

with the current rendering, but the new styles - the popover callouts will look broken.

So in other words, it would be best if both the rendering and the styling is updated. If either of the two is not, the popover won't be unusable, but it won't look okay.

Juveniel (approved):

PR #3758, opened 2022-08-01
fix(floating-label): set correct vertical position for all sizes

epetrow:
The vertical position of the floating label was calculated according to the input's padding-y in size medium and it was not correctly centered for the other sizes.


Juveniel:
We should think of a more consistent approach for handling both the `offset-x` and `offset-y`, as the horizontal spacing currently 'works' as the horizontal padding in all sizes is the same. Any change to the padding of the different input sizes will break the floating label, which is not ideal. 

Properly handling this scenario can require rendering changes to the floating label, thus I am closing this PR for the time being, or at least until we figure out the best approach.

PR #3760, opened 2022-08-02
feat(fluent): add fluent  inputs as individual modules

epetrow:
related to: https://github.com/telerik/kendo-themes/pull/3702


packages/fluent/scss/daterangepicker/_layout.scss
```diff
@@ -0,0 +1,46 @@
...
+        .k-dateinput {
+            width: $kendo-daterange-picker-input-width;
+        }
+    }
+
+
+    // Angular specific
+    kendo-daterange {
+        // @extend .k-daterangepicker;
+    }
```
Juveniel, line 30: remove this block

Juveniel (changes requested):
Use css vars with fallback in the `_layout`  and `_theme` files and make sure that the values correspond to the correct Fluent design guidelines

packages/fluent/scss/datepicker/index.scss
```diff
@@ -0,0 +1,26 @@
+// Module meta
+$_kendo-module-meta: (
+    name: "datepicker",
+    dependencies: (
+        "input",
+        "popup",
+        "calendar"
+    )
```
TeyaVes, line 8: I think we should add a dependency to the Button and Icon here as well

packages/fluent/scss/daterangepicker/_layout.scss
```diff
@@ -0,0 +1,39 @@
+@use "../core/" as *;
+@use "../input/_variables.scss" as *;
+@use "_variables.scss" as *;
+
+@mixin kendo-daterangepicker--layout() {
+
+    // Daterange picker
+    .k-daterange-picker {
+        width: var( --kendo-daterange-picker-width, #{$kendo-input-width} );
```
TeyaVes, line 9: I think here we should add var which is equal to the kendo-input-width:

var( --kendo-input-width, #{$kendo-input-width} )

In this way we gain two things:
1. the width of the daterangepicker can be changed with the Sass variable $kendo-daterange-picker-width, not only with the CSS var
2. If the --kendo-input-width css var is globally changed, it'll apply here as well

packages/fluent/scss/daterangepicker/index.scss
```diff
@@ -0,0 +1,27 @@
+// Module meta
+$_kendo-module-meta: (
+    name: "daterangepicker",
+    dependencies: (
+        "input",
+        "floating-label",
+        "popup",
+        "icon"
```
TeyaVes, line 8: add dependency to button here. There is a case where a swap button in rendered between the inputs

packages/fluent/scss/dropdowngrid/index.scss
```diff
@@ -0,0 +1,32 @@
...
+    dependencies: (
+        "table",
+        "input",
+        "floating-label",
+        "combobox",
+        "popup",
+        "toolbar",
+        "menu",
+        "grid"
+    )
```
TeyaVes, line 13: add button dependency

packages/fluent/scss/dropdowntree/index.scss
```diff
@@ -0,0 +1,27 @@
+// Module meta
+$_kendo-module-meta: (
+    name: "dropdowntree",
+    dependencies: (
+        "floating-label",
+        "dropdownlist",
+        "multiselect",
+        "treeview"
+    )
```
TeyaVes, line 9: add dependency to popup as we are using a var from it

packages/fluent/scss/numerictextbox/index.scss
```diff
@@ -0,0 +1,26 @@
+// Module meta
+$_kendo-module-meta: (
+    name: "numerictextbox",
+    dependencies: (
+        "input",
+        "floating-label",
+        "icon"
+    )
```
TeyaVes, line 8: add button dependency

packages/fluent/scss/datetimepicker/_layout.scss
```diff
@@ -0,0 +1,56 @@
...
+
+    k-datetimepicker {
+        &[dir="rtl"],
+        .k-rtl & {
+            .k-time-tab .k-datetime-selector {
+                transform: translateX(100%);
+            }
+        }
+    }
+}
```
TeyaVes, line 56: the datetime selector does not look okay in the pictures, I'll take a look at it tmrw

TeyaVes (changes requested):

Juveniel (approved):

packages/fluent/scss/datetimepicker/_layout.scss
Juveniel: in an ad-hoc discussion we agreed to resolve this with a separate PR for the Calendar

TeyaVes (approved):

PR #3764, opened 2022-08-03
Fluent scheduler

TeyaVes:
related to: https://github.com/telerik/kendo-themes-private/issues/3

Notes
- Adaptive styles are still missing
- Current day indicators (currently with red color) might change. Waiting for the designers to decide on a color.

Things left to be done in the PR:
- [x] once the toolbar fluent styles are added update the vars in the scheduler
- [x] add docs comments

Angular
- remove inline display: block for scheduler


packages/fluent/scss/scheduler/index.scss
```diff
@@ -0,0 +1,33 @@
...
+$_kendo-module-meta: (
+    name: "scheduler",
+    dependencies: (
+        "forms",
+        "adaptive",
+        "button",
+        "datetimepicker",
+        "dropdownlist",
+        "dialog",
+        "tooltip",
```
Juveniel, line 11: I think we should explicitly refer the "icon" module here

Juveniel (commented):
Looks great!

One thing that I noticed from the screenshots is that the recurrence editor does not look good. Could you please take a look ?

TeyaVes:
> Looks great!
> 
> One thing that I noticed from the screenshots is that the recurrence editor does not look good. Could you please take a look ?

The spacings will come once the radio buttons styles are added. If smth still does not look okay, I'll take a look.

Juveniel (approved):

PR #3778, opened 2022-08-05
Fix fluent variables

JoomFX:


packages/fluent/scss/listbox/_variables.scss
```diff
@@ -29,13 +29,13 @@ $kendo-listbox-line-height: var( --kendo-line-height, normal ) !default;
 
 /// Background color of the listbox component.
 /// @group listbox
-$kendo-listbox-bg: $kendo-component-bg !default;
+$kendo-listbox-bg: var( --kendo-component-bg, #{$kendo-component-bg} ) !default;
```
epetrow, line 32: I think it's better to give a different fallback to `$kendo-listbox-bg`, `-text` and `-border` here as the vars already have the same fallbacks.

Like so $kendo-listbox-bg: var( --kendo-component-bg, transparent ) or initial.

This will ensure there is a fallback if `$kendo-component-bg` is set to null for example.

packages/fluent/scss/listgroup/_variables.scss
```diff
@@ -17,13 +17,13 @@ $kendo-listgroup-line-height: var( --kendo-line-height, normal ) !default;
 
 /// Background color of the listgroup component.
 /// @group listgroup
-$kendo-listgroup-bg: $kendo-component-bg !default;
+$kendo-listgroup-bg: var( --kendo-component-bg, #{$kendo-component-bg} ) !default;
```
epetrow, line 20: These should have a different fallback value as well.

packages/fluent/scss/list/_variables.scss
```diff
@@ -184,13 +184,13 @@ $kendo-list-sizes: (
 
 /// Background color of the list component.
 /// @group list
-$kendo-list-bg: var( --kendo-component-bg, #{$kendo-component-bg} ) !default;
+$kendo-list-bg: var( --kendo-component-bg, transparent ) !default;
 /// Text color of the list component.
 /// @group list
-$kendo-list-text: var( --kendo-component-text, #{$kendo-component-text} ) !default;
+$kendo-list-text: var( --kendo-component-text, transparent ) !default;
```
TeyaVes, line 190: the text and border variables should not fallback to transparent color (as they might become invisible if the variable is not defined). It's okay for the background color, but the text and border should be either inherit or initial.

TeyaVes (changes requested):

TeyaVes (approved):

PR #3806, opened 2022-08-09
[Fluent] Add tooltip styles

epetrow:


packages/fluent/scss/tooltip/_variables.scss
```diff
@@ -0,0 +1,135 @@
...
+        bg: $kendo-tooltip-warning-bg,
+        text: $kendo-tooltip-warning-text,
+        border: $kendo-tooltip-warning-border
+    ),
+    error: (
+        bg: $kendo-tooltip-error-bg,
+        text: $kendo-tooltip-error-text,
+        border: $kendo-tooltip-error-border
+    )
+) !default;
```
Juveniel, line 134: These are no longer needed, as you iterate over the kendo-theme-colors map

Juveniel (changes requested):

packages/fluent/scss/tooltip/_theme.scss
```diff
@@ -0,0 +1,46 @@
...
+        --INTERNAL--kendo-tooltip-text: var( --kendo-tooltip-text, #{$kendo-tooltip-text} );
+        --INTERNAL--kendo-tooltip-bg: var( --kendo-tooltip-bg, #{$kendo-tooltip-bg} );
+        --INTERNAL--kendo-tooltip-border: var( --kendo-tooltip-border, #{$kendo-tooltip-border} );
+
+            color: var( --INTERNAL--kendo-tooltip-text, initial );
+            background-color: var( --INTERNAL--kendo-tooltip-bg, initial );
+            border-color: var( --INTERNAL--kendo-tooltip-border, initial );
+
+        .k-callout {
+            color: var( --INTERNAL--kendo-tooltip-bg, initial );
```
TeyaVes, line 22: The callout can be with an inherit color. In this way if the tooltip bg color is changed, the callout would be the same color. This would also allow you to remove the callout bg from the kendo-themes color map

TeyaVes (approved):

Juveniel (approved):

PR #3815, opened 2022-08-10
Fluent fixes

epetrow:
Related to: https://github.com/telerik/kendo-themes-private/issues/26#issuecomment-1209357864


packages/fluent/scss/input/_layout.scss
```diff
@@ -15,8 +15,8 @@
         border-style: solid;
         outline: 0;
         font-family: var( --kendo-input-font-family, #{$kendo-input-font-family} );
-        font-size: var( --INTERNAL--kendo-input-font-size, 1rem );
-        line-height: var( --INTERNAL--kendo-input-line-height, normal );
+        font-size: var( --INTERNAL--kendo-input-font-size, inherit );
+        line-height: var( --INTERNAL--kendo-input-line-height, #{$kendo-line-height} );
```
TeyaVes, line 19: if the font-size fallback is inherit, shouldn't the line-height be inherit as well?

epetrow: I've changed  these since I noticed when the component size is set to none the input becomes larger.

TeyaVes (approved):

Juveniel (approved):

PR #3841, opened 2022-08-17
Fluent treeview

JoomFX:
related to: https://github.com/telerik/kendo-themes-private/issues/3


packages/fluent/scss/treeview/_layout.scss
```diff
@@ -0,0 +1,160 @@
...
+        *::before,
+        *::after {
+            box-sizing: border-box;
+        }
+    }
+
+
+    // Treeview filter
+    .k-treeview-filter {
+        padding: var( --kendo-treeview-filter-padding, #{$kendo-treeview-filter-padding} );
```
Juveniel, line 34: split to padding-inline and padding-block for consistency

packages/fluent/scss/treeview/_layout.scss
```diff
@@ -0,0 +1,160 @@
...
+        display: flex;
+        flex-direction: row;
+        align-items: center;
+        align-content: center;
+    }
+
+
+    // Treeview item
+    .k-treeview-item {
+        outline-style: none;
```
Juveniel, line 70: move outline to bottom of declaration

packages/fluent/scss/treeview/_layout.scss
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+
+
+    // Treeview sizes
+    @each $size, $size-props in $kendo-treeview-sizes {
+        $_font-size: map-get( $size-props, font-size);
+        $_line-height: map-get( $size-props, line-height);
+        $_item-padding-x: map-get( $size-props, item-padding-x);
+        $_item-padding-y: map-get( $size-props, item-padding-y);
```
Juveniel, line 147: spacing

packages/fluent/scss/treeview/_variables.scss
```diff
@@ -0,0 +1,161 @@
...
+/// Text color of the treeview items.
+/// @group treeview
+$kendo-treeview-item-text: var( --kendo-component-text, inherit ) !default;
+/// Border color of the treeview items.
+/// @group treeview
+$kendo-treeview-item-border: inherit !default;
+
+/// Background color of hovered treeview items.
+/// @group treeview
+$kendo-treeview-item-hover-bg: get-theme-color-var( neutral-20 ) !default;
```
Juveniel, line 89: use --kendo-hover-bg

packages/fluent/scss/treeview/_variables.scss
```diff
@@ -0,0 +1,161 @@
...
+/// Text color of hovered treeview items.
+/// @group treeview
+$kendo-treeview-item-hover-text: get-theme-color-var( neutral-190 ) !default;
```
Juveniel, line 92: use --kendo-hover-text

packages/fluent/scss/treeview/_variables.scss
```diff
@@ -0,0 +1,161 @@
...
+/// Border color of hovered treeview items.
+/// @group treeview
+$kendo-treeview-item-hover-border: inherit !default;
+
+/// Background color of focused treeview items.
+/// @group treeview
+$kendo-treeview-item-focus-bg: $kendo-treeview-item-bg !default;
```
Juveniel, line 99: use css variable to enable customization ( ref to @TeyaVes comment in the Pager PR ). Do so for every reused variable.
```suggestion
$kendo-treeview-item-focus-bg: var( --kendo-treeview-item-bg, #{$kendo-treeview-item-bg} ) !default;
```

packages/fluent/scss/treeview/_variables.scss
```diff
@@ -0,0 +1,161 @@
...
+/// Border color of selected treeview items.
+/// @group treeview
+$kendo-treeview-item-selected-border: inherit !default;
+
+/// Background color of disabled treeview items.
+/// @group treeview
+$kendo-treeview-item-disabled-bg: $kendo-treeview-item-bg !default;
+/// Text color of disabled treeview items.
+/// @group treeview
+$kendo-treeview-item-disabled-text: get-theme-color-var( neutral-90 ) !default;
```
Juveniel, line 125: use --kendo-disabled-text

packages/fluent/scss/treeview/index.scss
```diff
@@ -0,0 +1,26 @@
...
+);
+
+
+// Component
+@forward "_variables.scss";
+@use "_layout.scss" as *;
+@use "_theme.scss" as *;
+
+// Register
+@use "../core/module-system" as module;
```
Juveniel, line 17: I should probably fix the generation script though 💯 

```suggestion
@use "../core/module-system/" as module;
```

Juveniel (changes requested):

packages/fluent/scss/treeview/_variables.scss
JoomFX: As far as I understood the idea when I spoke with @TeyaVes , we should apply this technique when we use variables from other components, so when the user changes the CSS variable in the other component it can be reflected in this component as well. Here's an example:

These variables:
```
$kendo-pager-item-padding-x: $kendo-button-padding-y-md !default;
$kendo-pager-item-hover-bg: $kendo-list-item-hover-bg !default;
```

Should be changed to:
```
$kendo-pager-item-padding-x: var( --kendo-button-padding-y-md, #{$kendo-button-padding-y-md}) !default;
$kendo-pager-item-hover-bg: var( --kendo-list-item-hover-bg, #{$kendo-list-item-hover-bg}) !default;
```

The reason for this is because we want when the user changes the CSS variable, for example `--kendo-list-item-hover-bg` the new color to be automatically applied to the `--kendo-pager-item-hover-bg`.

But in this case (in the case of the Treeview) we just want the initial value of the given SCSS variable to be equal to the value of another SCSS variable (from the same component) when the SCSS compiler compiles the CSS:
```
$kendo-treeview-item-focus-bg: $kendo-treeview-item-bg !default;
```

And from here on the user can customize the values further by using the following CSS variables:
```
--kendo-treeview-item-bg
--kendo-treeview-item-focus-bg
```

I don't know, I got even more confused while writing this comment :)
But if we decide that the correct approach is to always have to the SCSS variable to be assigned the CSS variable (which value represents another SCSS variable), it would mean that in the `_variables.scss` file we should almost never have to one SCSS variable to be assigned another SCSS variable which seems a bit strange to me. 
```
// NOT GOOD
$kendo-treeview-item-focus-bg: $kendo-treeview-item-bg !default;
```

Juveniel: Discussed this in an ad-hoc meeting with @joneff and @epetrow and we decided not to follow this approach as in this case, the value of the sass variable remains unknown ( value could come from a css or sass var ), thus the more clean approach is to stick to standard reuse of the sass variables ( without using the css var ).

Juveniel (approved):

PR #3855, opened 2022-08-22
[Fluent] Progressbar styles

kikostadinov:


packages/fluent/scss/progressbar/_layout.scss
```diff
@@ -0,0 +1,252 @@
+@use "_variables.scss" as *;
+@use "../core/mixins" as *;
+@use "../core/color-system" as *;
+
+@mixin kendo-progressbar--layout {
+
+    // Base
+    .k-progressbar {
+        @include border-radius( $kendo-progressbar-border-radius );
```
TeyaVes, line 9: add CSS var

packages/fluent/scss/progressbar/_variables.scss
```diff
@@ -0,0 +1,76 @@
...
+
+/// Background color of the progressbar.
+/// @group progressbar
+$kendo-progressbar-bg: get-theme-color-var( neutral-30 ) !default;
+/// Text color of the progressbar.
+/// @group progressbar
+$kendo-progressbar-text: var( --kendo-component-text, initial ) !default;
+/// Border color of the progressbar.
+/// @group progressbar
+$kendo-progressbar-border: $kendo-progressbar-bg !default;
```
TeyaVes, line 48: use the CSS variable here with a fallback to the Sass var:

var( --kendo-progressbar-bg, $kendo-progressbar-bg) 

In this way if anyone uses --kendo-progressbar-bg to override the bg color, it'll apply on the border as well.

The same goes for the rest of the sass vars used here

TeyaVes (changes requested):

Juveniel:
@kikostadinov two tings:

I noticed that the circular progressbar has primary-190 stroke ? Is this the color showcased in the design or should it be the primary-130 ?

We need a test page for the ChunkProgressBar Buttons. Can you add a test page for this ?

packages/fluent/scss/progressbar/_layout.scss
```diff
@@ -118,8 +118,8 @@
 
         .k-item {}
         .k-item + .k-item {
-            border-left-width: 1px;
-            border-inline-start-width: 1px;
+            border-left-width: 3px;
```
TeyaVes, line 121: do we need this rule since we have inline-start-width and it does the same?

packages/fluent/scss/progressbar/_layout.scss
```diff
@@ -173,8 +173,8 @@
         }
         .k-item {}
         .k-item + .k-item {
-            border-top-width: 1px;
-            border-block-start-width: 1px;
+            border-top-width: 3px;
```
TeyaVes, line 176: do we need this rule since we have block-start-width and it does the same?

packages/fluent/scss/progressbar/_layout.scss
```diff
@@ -118,8 +118,8 @@
 
         .k-item {}
         .k-item + .k-item {
-            border-left-width: 1px;
-            border-inline-start-width: 1px;
+            border-left-width: 3px;
+            border-inline-start-width: 3px;
```
TeyaVes, line 122: I think we should rather assign a variable here instead of using 3px

Juveniel (approved):

Juveniel:
@kikostadinov after @TeyaVes reviews this, please merge commits in the first feat, re-gen images and rebase the branch with develop.

TeyaVes (approved):

PR #3858, opened 2022-08-23
fix(charts): add chart tooltip popup styles when not a direct child

TeyaVes:
The following styles are not applied if k-popup is inside k-child-animation-container (which should be the correct rendering):

.k-animation-container > .k-child-animation-container > .k-popup


Juveniel (approved):

tsvetomir (approved):
Seems like a safe change 👍🏽 

We don't render `.k-child-animation-container` anywhere in the charts right now. How should that look?

TeyaVes:
@tsvetomir , the k-child-animation-container will be added in blazor to support the transitions - currently the transition is added to the .k-popup element, but as discussed with Vankata the correct approach is to add the transition to the child-animation-container element. So, the rendering would be as following:

```
<div class="k-animation-container">
    <div class="k-child-animation-container" style="transition:.......">
        <div class="k-popup">
             <div class="k-chart-tooltip">....</div>
        </div>
    </div>
</div>
```

In this way we avoid adding transition styles and overrides to the k-popup component.

PR #3886, opened 2022-08-29
feat(popover): add styles to Fluent

magdalenaan:
related https://github.com/telerik/kendo-themes-private/issues/3


packages/fluent/scss/popover/_variables.scss
```diff
@@ -0,0 +1,89 @@
...
+$kendo-popover-callout-border-width: $kendo-popover-border-width !default;
+/// Border style of the popover callout.
+/// @group popover
+$kendo-popover-callout-border-style: $kendo-popover-border-style !default;
+/// Background color of the popover callout.
+/// @group popover
+$kendo-popover-callout-bg: $kendo-popover-bg !default;
+/// Border color of the popover callout.
+/// @group popover
+$kendo-popover-callout-border: $kendo-popover-border !default;
```
TeyaVes, line 89: do we need the $kendo-popover-callout-bg and $kendo-popover-callout-border variables? We don't use them anywhere

packages/fluent/scss/popover/_variables.scss
```diff
@@ -0,0 +1,89 @@
...
+/// Border color of the popover header.
+/// @group popover
+$kendo-popover-header-border: $kendo-dialog-titlebar-border !default;
+
+/// Horizontal padding of the popover body.
+/// @group popover
+$kendo-popover-body-padding-x: $kendo-window-inner-padding-x !default;
+/// Vertical padding of the popover body.
+/// @group popover
+$kendo-popover-body-padding-y: $kendo-card-body-padding-y !default;
```
TeyaVes, line 66: these seem a bit inconsistent to me- the vertical padding is connected to the window-inner-padding-x, while the vertical to the card-padding-y. I would double check with the designer and rather make them the same as the paddings of the card (as in the rest of the themes)

TeyaVes (changes requested):
To me it seems that the variables are inconsistent- some are mapped to the window, others to the dialog, popup or the card. I think we should talk to the designer and map the variables to one (or 2 at most) component, e.g the card

packages/fluent/scss/popover/_variables.scss
magdalenaan: I have remove unnecessary variables.

packages/fluent/scss/popover/_variables.scss
magdalenaan: We had discussed the case with the designer and the conclusion was the following: as Popover in MS Fluent theme looks like the Dialog, it gets paddings from such widget. In our case Dialog applies some paddings from .k-window, so it has no variables for all of them. In the Popover we do not have .k-window className, so we should import some padding variables from the Window widget. For that reason we use paddings from the both widgets.
I have fixed/replaced padding imported from the Card.

Juveniel: This sounds rather strange to me. The card has the same composite elements as the Card (header, content, actions) and also identical colors and padding for the same. If we can set all variables in the Popover to inherit from Card variables, I think this would be a better solution implementation wise and would reduce the complexity and dependencies of the component.

magdalenaan: MS Fluent Popover uses the same paddings as Dialog and Window, not as Card.

packages/fluent/scss/popover/index.scss
```diff
@@ -0,0 +1,27 @@
+// Module meta
+$_kendo-module-meta: (
+    name: "popover",
+    dependencies: (
+        "action-buttons",
+        "dialog",
+        "widnow"
```
TeyaVes, line 7: there is a typo here - window

magdalenaan: fixed

PR #3892, opened 2022-08-30
[Fluent]: Mediaplayer styles

kikostadinov:


packages/fluent/scss/mediaplayer/_theme.scss
```diff
@@ -0,0 +1,30 @@
...
+
+    .k-mediaplayer-overlay {
+        background-color: var( --kendo-mediaplayer-overlay-bg, #{$kendo-mediaplayer-overlay-bg} );
+    }
+
+
+    .k-mediaplayer-titlebar {
+        color: var( --kendo-mediaplayer-titlebar-text, #{$kendo-mediaplayer-titlebar-text} );
+        background-color: var( --kendo-mediaplayer-titlebar-bg, #{$kendo-mediaplayer-titlebar-bg} );
+        text-shadow: 0 0 2px rgba( var( --kendo-mediaplayer-text, #{$kendo-mediaplayer-text} ), .5 );
```
Juveniel, line 23: rgba on css variables does not work. Use a separate variable and point to a color value from the palette (no opacity).

Juveniel (changes requested):

packages/fluent/scss/mediaplayer/_theme.scss
kikostadinov: Looks like the titlebar is not supposed to have text shadow so I had it removed. Thanks!

Juveniel (approved):

Issue #3804, opened 2022-08-09
Missing styles for focused cells in PivotGrid


nstoychev:
Example to check: https://stackblitz.com/edit/react-yvlxco?file=app%2Fmain.tsx

React issue: https://progresssoftware.aha.io/requirements/KUIR-295-5
Angular issue: https://progresssoftware.aha.io/requirements/KUING2-267-5

dtopalov:
outline: none is missing, and the Chrome focused styling interfere with ours:

![image](https://user-images.githubusercontent.com/7533599/188200659-bec32a7b-9d91-4924-b098-60af3a9ef7bf.png)

![image](https://user-images.githubusercontent.com/7533599/188200709-06721ca7-9289-4907-b3b6-329c26cb44ce.png)

PR #3937, opened 2022-09-09
fix(pivotgrid): add outline none to focus cells

PaskalevStoyan:
related to: https://github.com/telerik/kendo-themes/issues/3804

### DIFF
--- a/packages/default/scss/pivotgrid/_layout.scss
+++ b/packages/default/scss/pivotgrid/_layout.scss
@@ -60,6 +60,10 @@
         border-right-width: $pivotgrid-cell-border-width;
         border-bottom-style: solid;
         border-bottom-width: $pivotgrid-cell-border-width;
+
+        &:focus {
+            outline: none;
+        }
     }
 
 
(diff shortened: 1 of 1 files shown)


packages/default/scss/pivotgrid/_layout.scss
```diff
@@ -60,6 +60,11 @@
         border-right-width: $pivotgrid-cell-border-width;
         border-bottom-style: solid;
         border-bottom-width: $pivotgrid-cell-border-width;
+
+        &:focus,
+        &.k-focus {
```
Juveniel, line 65: .k-focus selector not needed here

packages/default/scss/pivotgrid/_layout.scss
```diff
@@ -130,6 +135,11 @@
         white-space: nowrap;
         vertical-align: top;
         overflow: hidden;
+
+        &:focus,
+        &.k-focus {
```
Juveniel, line 140: .k-focus selector not needed here

Juveniel (changes requested):

Juveniel (approved):

PR #3973, opened 2022-09-19
Add functionality to cancel workflows

joneff:
Should cancel workflows if someone pushes in the **same branch**

/cc @tsvetomir @dtopuzov @ivanchev -- we may find this bit useful, especially for long running builds that cost us money.


joneff:
And it does:
﻿
Screen Shot 2022-09-19 at 19 23 48
﻿

dtopuzov (approved):

Juveniel (approved):

PR #3974, opened 2022-09-19
Split visual tests into smaller parts

joneff:
Split tests into smaller chunks to better identify what takes time. True, there is some duplication involved for checking out the repo.


joneff:
It doesn't improve performance but it's clearer what takes time and takes less combined time.

before:
image


after:
image

Juveniel (approved):

PR #3978, opened 2022-09-20
Use child workflows for generating test assets

joneff:
Split `create-snapshots` and `compile-themes` into separate child workflows. that way, they can be reused if needed or optimized without touching the main workflow.


joneff:
Things begin to look better and better

image

Juveniel (approved):

PR #3988, opened 2022-09-21
Fluent fixes 9

kikostadinov:
Related to: https://github.com/telerik/kendo-themes-private/issues/26

**SplitButton**
- [x] The ddl arrow does not gets correct background color on hover in Flat mode and Base theme color 
_**Disclaimer**: Attached the delimiter element to k-button-icon, since both buttons already have pseudo elements and k-split-button has no access to the size classes in order to properly position the delimiter according to them._

**BottomNavigation**
- [x] Solid mode: Selected item should have paddings, currently the icon is next to the top border
image

- [x] Horizontal orientation: item size is wrong , height should be 48px(currently it's 28px)
image


packages/fluent/scss/bottom-navigation/_variables.scss
```diff
@@ -41,7 +45,7 @@ $kendo-bottom-nav-item-min-width: 72px !default;
 $kendo-bottom-nav-item-max-width: none !default;
 /// Minimum height of the bottom navigation item.
 /// @group bottom-navigation
-$kendo-bottom-nav-item-min-height: calc( var( --kendo-icon-size, 1rem ) + #{$kendo-padding-x * 2} - #{$kendo-bottom-nav-padding-x * 2} ) !default;
```
TeyaVes, line 44: the min-height should be related to the icon size. I would suggest replacing kendo-spacing 6 with $kendo-bottom-nav-item-icon-size var that you have defined above

packages/fluent/scss/split-button/_layout.scss
```diff
@@ -52,6 +35,40 @@
...
+                width: var( --kendo-split-button-arrow-delimiter-line-size, #{$kendo-split-button-arrow-delimiter-line-size} );
+                height: var( --kendo-split-button-arrow-delimiter-size, #{$kendo-split-button-arrow-delimiter-size} );
+                position: absolute;
+                top: calc( ( 100% - #{$kendo-split-button-arrow-delimiter-size} ) / 2 );
+                background: var( --kendo-split-button-arrow-delimiter-bg, #{$kendo-split-button-arrow-delimiter-bg} );
+                z-index: 2;
+            }
+        }
+
+        &.k-button-sm {
```
TeyaVes, line 50: I would suggest creating a map for this and defining the sizes in the variables file (similar to the sizes in the button). In this way it'll be easier if we add different button sizes in the future.

packages/fluent/scss/split-button/_variables.scss
```diff
@@ -22,3 +22,13 @@ $kendo-split-button-arrow-delimiter-line-size: 1px !default;
 /// Background color of the split button arrow delimiter.
 /// @group split-button
 $kendo-split-button-arrow-delimiter-bg: get-theme-color-var( neutral-60 ) !default;
+
+/// Horizontal offset of the split button arrow delimiter.
+/// @group split-button
+$kendo-split-button-arrow-delimiter-offset-sm: $kendo-split-button-arrow-padding-x-sm !default;
```
TeyaVes, line 28: do we need these extra variables here?

TeyaVes (changes requested):

TeyaVes (approved):

packages/fluent/scss/bottom-navigation/_variables.scss
```diff
@@ -41,13 +45,13 @@ $kendo-bottom-nav-item-min-width: 72px !default;
...
+$kendo-bottom-nav-item-min-height: calc( $kendo-bottom-nav-item-icon-size + $kendo-bottom-nav-item-icon-size + ($kendo-bottom-nav-item-padding-y * 2) ) !default;
```
Juveniel, line 48: use interpolation here, otherwise this will generate invalid styles

Juveniel (changes requested):

Juveniel (approved):

PR #3989, opened 2022-09-22
Fix Radio and Checkbox ripple

PaskalevStoyan:
related to: https://github.com/telerik/k2/issues/839


Juveniel:
@PaskalevStoyan this does not seem like a valid fix for this issue. The idea of the `.k-ripple-container` class is that it serves as a locator for different elements( buttons, radios, checkboxes ), that can have the ripple effect applied.

PaskalevStoyan:
@Juveniel You're right. There was a simpler approach. The new changes are pushed and are ready for review.

Juveniel:
Things left to do:

* revert tests changes
* move ripple styles customization to respective tests for checkbox/radio

PR #4006, opened 2022-09-28
Improve disabled styling in Fluent theme

epetrow:
Should improve disabled styling across Fluent


packages/fluent/scss/core/mixins/_disabled.scss
```diff
@@ -1,8 +1,7 @@
 // Disabled
-@mixin disabled( $color: null, $bg: null, $border: null ) {
+@mixin disabled( $color: null, $border: null ) {
```
joneff, line 2: you can use `null`

PR #4028, opened 2022-10-05
Allow $scrollbar-size customization

epetrow:
Related to: https://github.com/telerik/kendo-react/issues/1427


Xizario:
@telerik/kendo-react-team @telerik/blazor-team @telerik/kendo-vue-team @telerik/kendo-angular-team 
Make sure you init the --kendo-scrollbar-width var.
In React for example it was used in grid and dropdowns, and now it should be set in the calendar init as well.

Stamo-Gochev:
The blazor team already does this in some scenarios:
https://github.com/telerik/blazor/blob/148c1952edee990ef1db535a06fb5e20848ed74e/js/telerik-blazor/src/common/dom.js#L475-L476
I am not aware why the calendar needs that now, but we use it for the grid, multicolumncombobox and for virtualized scenarios for now.

Issue #4002, opened 2022-09-27
Component options for Pager


Juveniel:
design: https://github.com/telerik/web-components-ux/issues/478
part of: https://github.com/telerik/kendo-themes-private/issues/6

**Component options for Pager**

```ts
enum PagerSize { small, medium, large };

class Pager {
    size: ( null | PagerSize ) = PagerSize.medium
}
```

**Rendering**
* [x] Rendering

<details>

<summary> Pager </summary>

```html
<!-- default rendering -->

Numeric Pager

<div class="k-pager k-pager-md">
    <div class="k-pager-numbers-wrap">
        <button class="k-pager-nav k-pager-first k-button k-button-md k-button-flat k-button-flat-base k-icon-button">
            <span class="k-button-icon k-icon k-i-caret-alt-to-left"></span>
        </button>
        <button class="k-pager-nav k-button k-button-md k-button-flat k-button-flat-base k-icon-button">
            <span class="k-button-icon k-icon k-i-caret-alt-left"></span>
        </button>
        <select class="k-picker k-dropdown-list k-dropdown k-rounded-md k-picker-md">
            <option></option>
        </select>
        <div class="k-pager-numbers">
            <button class="k-button k-button-md k-button-flat k-button-flat-primary">
                <span class="k-button-text"></span>
            </button>
            <button class="k-button k-button-md k-button-flat k-button-flat-primary">
                <span class="k-button-text"></span>
            </button>
            <button class="k-button k-button-md k-button-flat k-button-flat-primary">
                <span class="k-button-text"></span>
            </button>
            <button class="k-button k-button-md k-button-flat k-button-flat-primary">
                <span class="k-button-text"></span>
            </button>
            <button class="k-button k-button-md k-button-flat k-button-flat-primary">
                <span class="k-button-text"></span>
            </button>
            <button class="k-button k-button-md k-button-flat k-button-flat-primary">
                <span class="k-button-text"></span>
            </button>
        </div>
        <button class="k-pager-nav k-button k-button-md k-button-flat k-button-flat-base k-icon-button">
            <span class="k-button-icon k-icon k-i-caret-alt-right"></span>
        </button>
        <button class="k-pager-nav k-pager-last k-button k-button-md k-button-flat k-button-flat-base k-icon-button">
            <span class="k-button-icon k-icon k-i-caret-alt-to-right"></span>
        </button>
    </div>
    <span class="k-pager-sizes">
[...]

PR #4032, opened 2022-10-06
Pager component options

epetrow:
**Do not merge until the updated rendering is addressed in all suites**

closes: https://github.com/telerik/kendo-themes/issues/4002

Completed: 
- [x] Implement component options
- [x] Add $kendo- prefix to variables
- [x] Create JSON structure
- [x]  Docs article containing each mode

Actions required:
- [ ] Add sizing options and update rendering in jQuery
- [ ] Add sizing options and update rendering in MVC & Core
- [ ] Add sizing options and update rendering in Angular - https://github.com/telerik/k2/issues/675
- [ ] Add sizing options and update rendering in React - https://github.com/telerik/kendo-react-private/pull/2017
- [ ] Add sizing options and update rendering in Blazor - https://github.com/telerik/blazor/pull/5519
- [ ] Add sizing options and update rendering in Vue


packages/default/scss/pager/_layout.scss
```diff
@@ -180,7 +139,14 @@
...
             margin-right: 1ex;
-            width: $pager-dropdown-width;
+            width: $kendo-pager-dropdown-width;
+        }
+
+        .k-pager-lg & {
+            .k-dropdown-list,
+            > select {
+                width: 6em;
+            }
```
TeyaVes, line 149: let's add variable for that and add it to the spacing map

docs/components/pager/index.md
```diff
@@ -0,0 +1,149 @@
...
+The Pager enables the user to split a set of data into pages.
+
+## Rendering
+
+The Pager has two types:
+
+### Numeric
+
+```html
+<div class="k-pager k-widget k-pager-md">
```
TeyaVes, line 19: remove k-widget class as it's outdated

packages/bootstrap/scss/pager/_theme.scss
```diff
@@ -1,36 +1,6 @@
 @import "~@progress/kendo-theme-default/scss/pager/_theme.scss";
 
 @include exports( "pager/theme/bootstrap" ) {
```
TeyaVes, line 3: here we can remove the whole mixin since we're not adding any bootstrap specific styles

packages/bootstrap/scss/pager/_variables.scss
```diff
@@ -1,66 +1,79 @@
...
+$kendo-pager-padding-x-sm: k-map-get( $spacing, 1 ) !default;
+$kendo-pager-padding-x-md: k-map-get( $spacing, 2 ) !default;
+$kendo-pager-padding-x-lg: k-map-get( $spacing, 2 ) + k-map-get( $spacing, thin ) !default;
+
+$kendo-pager-padding-y: $kendo-pager-padding-x !default;
+$kendo-pager-padding-y-sm: $kendo-pager-padding-x-sm !default;
+$kendo-pager-padding-y-md: $kendo-pager-padding-x-md !default;
+$kendo-pager-padding-y-lg: $kendo-pager-padding-x-lg !default;
+
+$kendo-pager-item-sm-min-width: $kendo-button-sm-calc-size !default;
```
TeyaVes, line 12: change variable names to $kendo-pager-item-min-width-{size}

packages/classic/scss/button/_variables.scss
```diff
@@ -42,6 +42,9 @@ $kendo-button-line-height-lg: $line-height-lg !default;
 /// Calculated height of the button.
 /// @group button
 $kendo-button-calc-size: calc( #{$kendo-button-line-height * 1em} + #{$kendo-button-padding-y * 2} + #{$kendo-button-border-width * 2} ) !default;
+$kendo-button-sm-calc-size: calc( #{$kendo-button-line-height-sm * 1em} + #{$kendo-button-padding-y-sm * 2} + #{$kendo-button-border-width * 2} ) !default;
```
TeyaVes, line 45: change variable names to $kendo-button-calc-size-{size}

packages/html-spec/src/pager/pager.json
```diff
@@ -0,0 +1,152 @@
...
+                ]
+            }
+        ],
+        "children": [
+            {
+                "name": "pager-nav",
+                "$ref": "button",
+                "selector": [
+                    ".k-pager-nav.k-button"
+                ],
```
TeyaVes, line 55: I'm not sure if we need the .k-button selector here and below

packages/html/src/grid/tests/grid-adaptive-pager.tsx
```diff
@@ -17,24 +17,7 @@ root.render(
...
-                                                <li><a className="k-link">2</a></li>
-                                                <li><a className="k-link">3</a></li>
-                                                <li><span className="k-link k-selected">4</span></li>
-                                            </ul>
-                                        </div>
-                                        <a className="k-link k-pager-nav k-disabled"><Icon name="arrow-e" /></a>
-                                        <a className="k-link k-pager-nav k-pager-last k-disabled"><Icon name="arrow-end-right" /></a>
-                                        <span className="k-pager-info k-label">61 - 77 of 77 items</span>
-                                    </div>
+                                    <Pager size="small" className="k-grid-pager" mobile="small" showPagerInput={false} />
```
TeyaVes, line 20: I think we should keep size="medium" (or omit the size) here, so that the images remain unchanged.

packages/html/src/grid/tests/grid-sticky-columns-rtl.tsx
```diff
@@ -150,25 +150,7 @@ root.render(
...
-                    </div>
-                    <a href="#" className="k-link k-pager-nav k-disabled"><Icon name="arrow-e" /></a>
-                    <a href="#" className="k-link k-pager-nav k-pager-last k-disabled"><Icon name="arrow-end-right" /></a>
-                    <span className="k-pager-sizes k-label">
-                        <DropdownList value="20" />
-                        items per page
-                    </span>
-                    <div className="k-pager-info k-label">1 - 6 of 6 items</div>
-                </div>
+                <Pager size="small" className="k-grid-pager" mobile="small" showPagerInput={false} />
```
TeyaVes, line 153: I would suggest removing the sizing here

TeyaVes (changes requested):

packages/html/src/pager/pager.tsx
```diff
@@ -0,0 +1,233 @@
...
+        } = this.props;
+
+        return (
+            <div
+                {...htmlAttributes}
+                dir={this.props.dir}
+                className={classNames(
+                    className,
+                    'k-pager',
+                    'k-widget',
```
TeyaVes, line 55: remove k-widget from here as well

packages/html-spec/src/pager/pager.json
epetrow: I've removed it from the navigation button selector, but for the other buttons it's needed in my opinion otherwise we will be targeting only their container `k-pager-numbers`.

TeyaVes: If the k-pager-numbers is a container and the button is inside, then the selector should be .k-pager-numbers .k-button

epetrow: You are correct and I left it where it is needed https://github.com/telerik/kendo-themes/pull/4032/files#diff-2a82e99c6c536d38727dbba6094a7916f01be2d93abb746f740ea2bf8dfc111aR71

TeyaVes (approved):

JoomFX (approved):

Juveniel (approved):

magdalenaan:
I have noticed that we use the following three different approaches (class names) for buttons that navigate to the first and last pages in our suites:

- k-i-arrow-end-left / -right
- k-i-caret-alt-to-left / -right
- k-i-seek-w / -e

The first one is initial and in my opinion the most intuitive. I suggest to use the k-i-arrow-end-left/right class name in the referent rendering instead of the k-i-caret-alt-to-left/right .

TeyaVes:
> I have noticed that we use the following three different approaches (class names) for icons of buttons that navigate to the first and last pages in our suites:
> 
> * k-i-arrow-end-left / -right
> * k-i-caret-alt-to-left / -right
> * k-i-seek-w / -e
> 
> The first one is initial and in my opinion the most intuitive. I suggest to use it in our referent rendering instead of the k-i-caret-alt-to-left/right.

@magdalenaan , we've agreed with the product themes that we should use the 'main' name of the icon and not the alias in each component taken from the icons repo list with icons here: https://github.com/telerik/kendo-icons/blob/develop/src/telerik-icons/icons.json , since there is a plan to depricate the aliases at one point. 
Thus, the correct icon names for the navigation are 
`k-i-caret-alt-to-left / -right`

magdalenaan:
Is it necessary to leave the `k-pager-wrap` selector?

Juveniel:
@epetrow please rebase the PR with develop and group styles and variable docs in single commit. 

In addition to the above, please add respective breaking changes labels to the respective commits (for the rendering change of the buttons, for the added size class and for the prefixed variables).

packages/bootstrap/scss/pager/_theme.scss
```diff
@@ -1,36 +1 @@
 @import "~@progress/kendo-theme-default/scss/pager/_theme.scss";
-
```
joneff, line 2: ❤️ I've waited for a long time to delete this!

joneff (approved):
We could improve the pager options a bit, but for now it's good!

PR #4033, opened 2022-10-06
Generate nuget for themes

joneff:
Here be nugets... There are still things to consider:

* [ ] Name of the packages
* [ ] extraction locations
* [ ] sync / populating from package.json
* [ ] action

/cc @telerik/tech-guidance @telerik/blazor-team 

Attached is the resulting nuget generated with `nuget pack package.nuspec` -- [Telerik.Kendo.ThemeDefault.5.8.2-dev.4.nupkg.zip](https://github.com/telerik/kendo-themes/files/9727912/Telerik.Kendo.ThemeDefault.5.8.2-dev.4.nupkg.zip)


packages/default/package.nuspec
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd">
+  <metadata>
+    <id>Telerik.Kendo.ThemeDefault</id>
```
Stamo-Gochev, line 4: Regarding the name:
1. `Telerik.Kendo.ThemeDefault` - this keeps the "Telerik" and "Kendo" prefixes, but I am not sure we actually need this or we can go with the next choice
2. `Telerik.KendoThemes.Default` - the prefix "Telerik" is a reserved word for verification and then we append the name of the entity (kendo themes) and separate the actual theme name (`[Default]`) with a dot
3. `Telerik.KendoTheme.Default` - uses the singular word "theme" that better mathes the npm name - `@progress/kendo-theme-default`. I find having a separation between "Theme" and "Default" (the actual name) better than option 1).

The 3rd option seems to have most advantages to me for now.

joneff: We could go with `Telerik.Blazor.ThemeDefault` or `Telerik.BlazorTheme.Default`.

jivanova: i would vote for 3rd option as well. 

The nuget is beneficial outside of blazor domain, so it's better not to use blazor keyword

ivanchev: We already have packages like `Telerik.Windows.Themes.Fluent.for.Wpf` so I believe that `Themes` should be plural here as well. So from the provided options I prefer the second one. (We also have nugets like `Telerik.Documents..` which is also plural)

joneff: So, to sum up, we have suggestions for:

* Telerik.Kendo.ThemeDefault -- my original suggestion that mirros `@progress/kendo-theme-default`
* Telerik.Kendo.Themes.Default -- a bit deeper naming that is in line with the rest of our offering
* Telerik.KendoThemes.Default and Telerik.KendoTheme.Default -- sort of a mixture between the two

ivanchev: `Telerik` is the sibling brand of `Kendo` for the .net world. In that sense `Telerik.Kendo.` is not a mirror to `progress/kendo` - both `Telerik` and `Kendo` are under `Progress`. It seems odd. I do however see the apparent similarity to `Telerik.Windows.`

Stamo-Gochev: I think that the plural `Windows` is not a result of our thinking whether to use `Window` or `Windows` thus the considerations regarding `Themes vs. Theme` have a different context in my opinion. Having both prefixes - `Telerik.Kendo.<something>` seems unnecessary - if we want to emphasize the Kendo bits, then we can go with the `Kendo.<something>` right away as we have the `Kendo-` prefix reserved as well.  We also have the option for `Progress.<something>` on nuget.

I am under the impression that we want to expose things under the `Telerik` branch as well as build the name around the changing parts - the more "constant" things are Telerik, Kendo and Theme(s), while the last word will be `Default`, `Material`, etc. I am not sure that we want to have things like `Telerik.Kendo.<Recurrence>` or `Telerik.Kendo.<SomethingElse>` - do we have reasons to mix the brand prefixes?

joneff: It's best if we discuss it over a meeting.

ivanchev: Let's not focus on the plural or usage of `Windows` in `Telerik.Windows.Themes`, but rather the plural of `Themes` there. That's what I was mainly referencing.

packages/bootstrap/package.nuspec
```diff
@@ -0,0 +1,38 @@
...
+      <files include="dist/*.scss" buildAction="None" copyToOutput="false" />
+      <files include="scss/**/*.scss" buildAction="None" copyToOutput="false" />
+    </contentFiles>
+    <dependencies>
+      <dependency id="Telerik.Kendo.ThemeCore" version="[$VERSION$]" />
+      <dependency id="Telerik.Kendo.ThemeUtils" version="[$VERSION$]" />
+      <dependency id="Telerik.Kendo.ThemeDefault" version="[$VERSION$]" />
+      <dependency id="bootstrap.sass" version="[5.2.1]" />
+    </dependencies>
+  </metadata>
```
Stamo-Gochev, line 29: Wondering if both Core, Utils, Default and Bootstrap are required here - the corresponding [npm](https://www.npmjs.com/package/@progress/kendo-theme-bootstrap) package is dependent on:
- [@progress/kendo-theme-core](https://www.npmjs.com/package/%40progress%2Fkendo-theme-core) [@progress/kendo-theme-default](https://www.npmjs.com/package/%40progress%2Fkendo-theme-default) [bootstrap](https://www.npmjs.com/package/bootstrap)

The same thing for material depends on Core, Utils and Default, while the [npm](https://www.npmjs.com/package/@progress/kendo-theme-material) package is dependent on:
- [@progress/kendo-theme-core](https://www.npmjs.com/package/%40progress%2Fkendo-theme-core) [@progress/kendo-theme-default](https://www.npmjs.com/package/%40progress%2Fkendo-theme-default)

Should the list of dependencies be the other way around?

Stamo-Gochev (approved):
There is a `[Content_Types].xml` file in the attached sample bundle - not sure if it is required and should be named with this exact casing.

We also discussed:
- [x] rename the `/dist` folder to `/wwwroot`, so that it can expose static assets for blazor directly (otherwise they are not reachable).

packages/bootstrap/package.nuspec
joneff: future compatible tweak

joneff:
New version of default theme attached -- [Telerik.KendoThemes.Default.5.9.1-dev.0.nupkg.zip](https://github.com/telerik/kendo-themes/files/9818846/Telerik.KendoThemes.Default.5.9.1-dev.0.nupkg.zip)

PR #4034, opened 2022-10-07
fix(progressbar): label moves down after initial interaction

kikostadinov:
Related to: https://github.com/telerik/kendo-themes-private/issues/26

"**_Labels moves down after changing the progress from 0 to another value. The behaviour could be observed only initially. To reproduce click on the 'Vote' button on this page_**"


TeyaVes (approved):

epetrow (approved):

TeyaVes (changes requested):
we'll need to fix the label of the vertical progress bar

TeyaVes (approved):

PR #4037, opened 2022-10-08
Introcude color system

joneff:
* use it in nouvelle theme instead of the old one
* remove redundant variables
* prefix variables with `$kendo-`
* use utils internally
* as a result, exclude nouvelle from docs generation
* lastly, include icons from icons package


packages/core/scss/color-system/_functions.import.scss
```diff
@@ -0,0 +1,91 @@
...
+    @return $result;
+
+};
+
+
+@function k-process-variant-matrices( $theme-matrix, $palette-matrix: () ) {
+    $result: ();
+
+    // @debug $theme-matrix;
+    // @debug $palette-matrix;
```
Juveniel, line 47: remove the comments

packages/core/scss/color-system/_palettes.scss
```diff
@@ -0,0 +1,915 @@
...
+        600: #9e2f1c,
+        700: #801f0f,
+        800: #691105,
+        900: #510901,
+        1000: #000000
+    ),
+
+
+    // Tailwind
+    tailwindNeutral: (
```
Juveniel, line 606: do we really need those palettes ? I think we can refrain from adding tailwind and primer palettes here

packages/core/scss/functions/_custom-properties.import.scss
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @return var( --#{$_prefix}#{$_var} );
+}
+
+@function k-vars( $prefix: kendo-, $vars... ) {
+    $vars: k-list-reverse( $vars );
+    $result: null;
+
+    // @debug $prefix;
```
Juveniel, line 31: remove comment

packages/core/scss/color-system/_functions.import.scss
Juveniel, line 47: remove comments

Juveniel (approved):

PR #4055, opened 2022-10-19
Successfully publish nugets

joneff:
We needed just a bit of tweaking, but it's ready: https://github.com/orgs/telerik/packages?repo_name=kendo-themes. It's worth noting that the currently released packages are not production ready yet, because we haven't really tested them. I am, however, confident that we'll be able to make them production ready in time for PI2.

@telerik/blazor-team if need be, we can publish to any nuget feed, since the `nuget-push` script accepts parameters.


Juveniel (approved):
lgtm

matanasov:
The nuget package doesn't contain any css files. It has only scss files. We need to have the theme all.css file as well as all swatch .css files in the package:
https://www.screencast.com/t/Pwau4v6Slxyw

joneff:
@matanasov https://github.com/telerik/kendo-themes/packages/1680584 appears everything is working correctly

Stamo-Gochev:
The nuget package contains CSS files indeed, they are just in a specific folder:
image

We need to check if they can be read from the path `contentFiles/any/any/wwwroot/css` or we need to use something like `contentFiles/wwwroot/css` or even `/wwwroot` for Blazor.

matanasov:
The package seems to install correctly and .css files are now distributed. However they are added to the project as links and are not served from the project folder, but from the .nuget cache folder. Compare the normal .css file location compared to the themes .css file location bellow:

image

Issue #3554, opened 2022-04-19
Focus state  in Default Main Dark swatch is not visible.


Xizario:
Reproducible for buttons and breadcrumb (might be encountered in other components as well) .

Steps to reproduce:
1. Open https://www.telerik.com/kendo-react-ui/components/buttons/button/ , or the Theme Builder
2. Choose Dark Main theme.
3. Focus a button. The border is not visible. (it is visible for primary and flat buttons only)

NikoPenev21:
https://github.com/telerik/web-components-ux/issues/471

PR #4062, opened 2022-10-20
Correct focused button state dark swatch

NikoPenev21:
fixes: #3554


Juveniel:
A few notes from me to take into consideration:

1) Do not put the issue link in the commit body. Instead, add this in the description of the PR e.g. `fixes: #issue`
2) The issue that is to be fixed needs a milestone to be set - this indicates in which release the fix will be available.
3) The related issue is not just related to Buttons, but rather the focus styling in the dark swatches. Thus, we also need to take into consideration focus states in general (inputs, lists, etc. ). You can refer to [the following design guidelines ](https://www.figma.com/file/gWUIux95BHxemqWwOw9ZwB/Default-Theme---Internal-Usage?node-id=5539%3A155480)for the exact states.

joneff (changes requested):
We do not have a picker showing colors with opacity in theme-builder, so exclude the change from `main-dark.json`.

Also, you have two commits with the exact same name and this may make it confusing for customers.

Finally, if you are going to mention a github issue from commit, make sure you are using the correct format and that you are specifying the relation between the commit and issue.

TeyaVes (changes requested):
if we're adding a check in the variables files, then we should remove it from the swatch json

Juveniel:
Closing this for the time being, as we will look into a more streamlined solution with the integration of a color & elevation systems in the themes.

// cc @kspeyanski @JoomFX @magdalenaan

PR #4066, opened 2022-10-26
fix(scheduler): expand vertical grouping to the entire view height

magdalenaan:
related https://github.com/telerik/kendo/issues/8982


Juveniel:
@magdalenaan this fix breaks the standard time column ( see previews ). We should find an alternative fix for this.

magdalenaan:
The issue will be resolved in kendo widget by dynamically calculated height by JavaScript.

Issue #4010, opened 2022-09-30
The hierarchy layout of the Drawer is broken in Material and Bootstrap


mbechev:
**Describe the bug**
In Material and Bootstrap theme the Hierarchical Drawer doesn't provide the needed nesting behavior of the child elements and also breaks the padding of the arrow icons:
![image](https://user-images.githubusercontent.com/33104481/193267867-07158251-c4da-4455-b14d-9bf38f77d34c.png)

Using Default theme:
![image](https://user-images.githubusercontent.com/33104481/193267963-1d41bb5e-d396-41e2-a9c2-7d867461a16f.png)

**To reproduce**
The issue is visible in the [Hierarchical Drawer](https://www.telerik.com/kendo-angular-ui/components/layout/drawer/hierarchical-drawer/) page.

PR #4070, opened 2022-10-27
fix(drawer): hierarchy layout

kikostadinov:
Fixes: https://github.com/telerik/kendo-themes/issues/4010


TeyaVes (commented):
I don't see any changes in the variables file in the classic theme, as well as in the layouts and variables files for fluent. Aren't we adding the fix there as well?

packages/bootstrap/scss/drawer/_layout.scss
```diff
@@ -15,6 +15,12 @@
                     padding-left: $drawer-item-padding-x;
                 }
             }
+
+            // Fallback styles when drawer item is missing k-spacer element
+            .k-i-arrow-chevron-up,
+            .k-i-arrow-chevron-down {
```
Juveniel, line 21: Targeting specific icon classes is not a good idea and flaky. Find another way.

packages/bootstrap/scss/drawer/_variables.scss
```diff
@@ -8,7 +8,7 @@ $drawer-font-size: $font-size !default;
 $drawer-line-height: $line-height !default;
 $drawer-content-padding-x: $padding-x !default;
 $drawer-content-padding-y: $padding-y !default;
-$drawer-item-level-padding-x: k-map-get( $spacing, 4 ) !default;
+$drawer-item-level-padding-x: $drawer-content-padding-x !default;
```
Juveniel, line 11: Do we want to modify the hierarchy items padding when changing the padding of the drawer content ? I think no, thus, better use `$padding-x` here.

Juveniel (changes requested):
One additional thing - it seems from the changed drawer visual tests, that the padding of the first level is bigger than the rest. Also, the change is only visible in the classic, default and fluent themes, but not in bootstrap and material. Can you have a look at this ?

packages/bootstrap/scss/drawer/_layout.scss
joneff: if this is "unification" the class must be present!

packages/bootstrap/scss/drawer/_layout.scss
```diff
@@ -15,6 +15,12 @@
...
+                margin-inline-start: auto;
+            }
         }
 
         .k-rtl & .k-drawer-items,
```
joneff, line 26: use logical padding

joneff (changes requested):

joneff (approved):

joneff:
@TeyaVes can you add the breaking change note in your commit. Apart from that minor thingie, lgtm

TeyaVes (approved):

joneff (approved):

PR #4072, opened 2022-10-29
Nouvelle inputs

joneff:
* use color system for input and picker type components
* ensure all icon related variables are contained in the icon modules
* add function for checking if element is in list
* add function for working with nullable variables


Juveniel (approved):
Note: we should think of a way to have the global css vars(--kendo-font-size-md, --kendo-border-radius-md, etc.) working for all components in order to make customization easier.

Issue #2957, opened 2021-11-11
Kendo Angular Tabstrip - Tab alignment does not work


JoomFX:
**Describe the bug**
With the new Tabstrip rendering (implemented because of the scrollable tabs) the [Tab alignment](https://www.telerik.com/kendo-angular-ui/components/layout/tabstrip/tabs/#toc-tab-alignment) functionality no longer works. It can be fixed by adding `width: 100%;` (for the horizontal tabs) or `height: 100%;` (for the vertical tabs) to the `.k-tabstrip-items` element.

The Tab alignment functionality seems to be available in Kendo Angular only and is implemented by adding inline styles.
It will be better if the inline styles are replaced by our [flex utility classes](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/utils/_flex.scss#L126) so we can keep the universal rendering even if/when the other Kendo flavors implement this functionality.

**To reproduce**

1. Go to the [Tab Alignment demo](https://www.telerik.com/kendo-angular-ui/components/layout/tabstrip/tabs/#toc-tab-alignment).
1. Try to change the Tab alignment

**Expected behavior**
The tab alignment functionality should work.

**Screenshots**
N/A

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for Angular

**Affected browsers (please remove the unneeded items)**
* All

**Build system information (please remove the unneeded items)**
* Not Applicable

Issue #2909, opened 2021-10-22
z-index of k-tabstrip-items-wrapper causes issues


JoomFX:
**Describe the bug**
If there is a Dialog in the Tab content, the `k-tabstrip-items-wrapper` element stays above the overlay because it has a higher `z-index` than the `k-content` element.
For reference - ticket 1539612.

**To reproduce**

1. Go to [this Stackblitz example](https://stackblitz.com/edit/angular-o1ed16-y1vnk3?file=app/app.component.ts)
1. Click on the "Open dialog" button
1. Notice how the tabs are above the overlay

**Expected behavior**
Since we use the `z-index` on the `k-tabstrip-items-wrapper` element for styling purposes only (so the border of the `k-tabstrip-items` element can be visible) we can remove it and add a 1px margin instead.

**Screenshots**
![tabstrip](https://user-images.githubusercontent.com/7753940/138446290-acf0d2c1-fa05-4cc1-8f3c-6d5ed22fafee.png)

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material
* theme-tasks

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**
* All

**Build system information (please remove the unneeded items)**
* Not Applicable

pclancysc:
I disagree.  This is a breaking change since previously the tab container had no z-index and nested elements with z-index value would be treated as if they were at a root level (or whatever child element order had been set).  But now, because the tab content has a z-index value any nested items with a z-index value will be treated as relative from that element and hence all existing code where dialogs were nested inside tabs no longer work as they did.

snisar:
Was this fixed, as I am still seeing an issue when opening a dialog within a tab on the latest release. Thanks

pateldips:
I am having the issue with the same. I am using "@progress/kendo-angular-layout": "^6.5.0"

this is what they have

.k-tabstrip-items-wrapper{
    z-index: 2;
}

I have replaced with this, but top border disappearing because of it. 

.k-tabstrip-items-wrapper{
    z-index: inherit;
}

I got the same issue with Bootstrapper 5.0, they also using as below. 
.input-group .btn{
    z-index: 2;
}

rajaharahap:
you can add this css :

.k-tabstrip-items-wrapper{
[...]

PR #4085, opened 2022-11-01
Various tabstrip improvements

joneff:
* add function to convert between `rgba` and solid colors
* do not overlap tabs
* use tabstrip indicator in fluent (part of the above)
* remove z-index from `tabstrip-items` and `tabstrip-content`
* add utility class names for tabstrip alignment.

@JoomFX please check if #2957 is regressing; also check if #2909 is resolved (it's closed for another reason).


epetrow (approved):

JoomFX:
@joneff Everything works as expected - #2957 is not regressing and #2909 is resolved!

PR #4092, opened 2022-11-07
Unify rendering appbar

kikostadinov:
closes: https://github.com/telerik/kendo-themes-private/issues/52

### Rendering

<details>

<summary> Universal </summary>

```html
<!-- default rendering -->

<div class="k-appbar k-appbar-primary">
	<div class="k-appbar-section">...</div>
	<span class="k-appbar-spacer k-appbar-spacer-sized"></span>
	...
	<span class="k-appbar-spacer"></span>
	<div class="k-appbar-section">
		...
		<span class="k-appbar-separator"></span>
		...
	</div>
</div>
```
</details>

### Visual Tests

`/tests/appbar`

## Deliverables

- [x] Universal rendering
- [x] HTML Component
- [x] HTML JSON Spec
- [x] Rename variables to `$kendo-[varName]`
- [x] Update tests


epetrow:
@kikostadinov A few things I noticed:

1. The commit message test(appbar): add html component should be > feat(html): add appbar html helper
2. I think we should add `k-appbar-inherit` class to the Appbar test because it seems like it does not get the expected color in the Fluent theme
3. Should we add a `k-appbar-actions` class to the Appbar section when there are buttons inside? Currently there is a class that is called just `actions`  @telerik/kendo-front-end, what do you think?

kikostadinov:
@epetrow I have added the changes you requested in 1 and 2 and will address point 3 once we have input from the rest of the team. Thank you.

TeyaVes (commented):
looks good to me, just a few things to point:
1. we agreed that in the html-spec we won't be exposing selectors which are adding layout styles- positioning, display, etc. Thus, I am wondering wether we need the appbar-sticky and fixed, as well as the position-top, bottom
2. I dont think we should add k-appbar-actions class as the contents of the appbar should be left to the client

TeyaVes (approved):

joneff (approved):

PR #4110, opened 2022-11-18
Unify rendering listview

magdalenaan:
closes https://github.com/telerik/kendo-themes-private/issues/49


packages/html-spec/src/listview/listview.json
```diff
@@ -0,0 +1,158 @@
...
+                    ]
+                },
+                {
+                    "name": "horizontal",
+                    "selector": [
+                        ".k-listview-borders-horizontal"
+                    ]
+                },
+                {
+                    "name": "all",
```
TeyaVes, line 80: this should be vertical

packages/html-spec/src/listview/listview.json
```diff
@@ -0,0 +1,158 @@
...
+        {
+            "name": "disabled",
+            "selector": [
+                ".k-disabled"
+            ]
+        }
+    ],
+    "options": [
+        {
+            "name": "layout",
```
TeyaVes, line 18: since layout and flex-direction are utility classes and not component-specific ones, I don't think they should be added to the JSON.

packages/html-spec/src/listview/listview.json
```diff
@@ -0,0 +1,158 @@
...
+            "name": "listview-footer",
+            "selector": [
+                ".k-listview-footer"
+            ]
+        },
+        {
+            "name": "listview-pager",
+            "$ref": "pager",
+            "selector": [
+                ".k-pager",
```
TeyaVes, line 134: we don't need k-pager class, as it'll come from the $ref: pager

packages/html/src/listview/README.md
```diff
@@ -0,0 +1,28 @@
...
+        ...
+    </div>
+
+    <!-- footer -->
+    <div class="k-listview-footer">
+        <!-- template -->
+    </div>
+
+    <!-- pager -->
+    <div class="k-pager k-listview-pager   k-pager-md      k-pager-wrap k-floatwrap">
```
TeyaVes, line 24: we don't need k-pager-wrap and k-floatwrap classes here

packages/html/src/listview/README.md
```diff
@@ -0,0 +1,28 @@
...
+<div class="k-listview k-listview-bordered   k-d-flex">
+
+    <!-- header -->
+    <div class="k-listview-header">
+        <!-- template -->
+    </div>
+
+    <!-- content -->
+    <div class="k-listview-content">
+            <!-- item template with added "k-listview-item" class to direct children -->
```
TeyaVes, line 14: can you please give an example of the item template dom with the k-listview-item class

packages/html/src/listview/README.md
```diff
@@ -0,0 +1,28 @@
+## Listview
+
+```html
+<!-- default rendering -->
+<div class="k-listview k-listview-bordered   k-d-flex">
```
TeyaVes, line 5: remove extra spaces in classes

packages/html/src/listview/listview-item.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export interface ListViewItemProps {
+    className?: string;
+    children?: any;
+    focus?: boolean;
+    selected?: boolean;
+}
+export class ListViewItem extends React.Component<ListViewItemProps> {
+
+    static defaultProps = {
```
TeyaVes, line 12: do we use the defaultProps anywhere here?

packages/html/src/listview/tests/listview-layout.tsx
```diff
@@ -0,0 +1,85 @@
...
+root.render(
+    <>
+        <style>{styles}</style>
+        <div id="test-area" className="k-d-grid k-grid-cols-2">
+
+            <section>
+                <span>Flex layout - row</span>
+                <ListView
+                    borders="vertical">
+                    <ListViewItem ><div>Listview Item 1</div></ListViewItem>
```
TeyaVes, line 27: remove spaces after ListViewItem

TeyaVes (dismissed):
A few comments from me + something that is not quite clear- if we are adding k-d-flex utility class by default to the listview, why don't we add display: flex to .k-listview and remove this utility class? Also, with what kind of configuration are we having grid layout, flex direction, etc. , I couldn't see it in the demos?

magdalenaan:
We have a confuse situation with ListView borders per suites. The jQuery widget has a Boolean property [`bordered`](https://docs.telerik.com/kendo-ui/api/javascript/ui/listview/configuration/bordered), which default value is `true`. This property adds borders around the widget and bottom border to its header and top border to its footer. On the other hand in others suites we do not have such property and ListView widget does not have any borders rendered.

Likewise there are different appearance per our themes design as the follow:
- Default theme - has border around a widget and bottom header border. Footer does not have border.
- Material theme - has bottom header and top footer border. There is not border around the widget.
- Bootstrap and Fluent - all above mentioned elements have borders.

One possible solution is to add a third state to the bordered property - `auto`, which will be default value and will apply different approach by design. With values `true` and `false` all borders will be turn on/off. This approach will follow the design, but will be a breaking change for all suites except jQuery.
Other solution is to make default value of the `bordered` property to be `false`. However this approach requires design revision and changes to make appearance per themes to be consistent. This will also cause breaking change for our jQuery clients.

Please share your ideas or recommendations.
@telerik/kendo-ux

packages/html/src/listview/README.md
magdalenaan: I am not sure if this would not be confused for developers and as result they will be add such element as wrapper to item template content.

packages/html-spec/src/listview/listview.json
magdalenaan: the classes are need for other listView option which is flex-direction

magdalenaan:
> A few comments from me + something that is not quite clear- if we are adding k-d-flex utility class by default to the listview, why don't we add display: flex to .k-listview and remove this utility class? Also, with what kind of configuration are we having grid layout, flex direction, etc. , I couldn't see it in the demos?

We have an [api in jQuery](https://docs.telerik.com/kendo-ui/api/javascript/ui/listview#configuration) for grid or flex layout. Depending on the value we set corresponding class: k-d-flex or k-d-grid.

magdalenaan:
telerik/kendo-a11y could you review the PR?

packages/html/src/listview/listview.tsx
```diff
@@ -0,0 +1,92 @@
...
+    pagerPosition?: 'top' | 'bottom';
+    bordered?: boolean;
+    borders?: 'all' | 'horizontal' | 'vertical';
+    layout?: 'flex' | 'grid';
+    flexDirection?: 'row' | 'row-reverse' | 'col' | 'col-reverse';
+    flexWrap?: 'wrap' | 'nowrap' | 'wrap-reverse';
+    gridColumns?: number;
+    disabled?: boolean;
+    loading?: boolean;
+    item?;
```
TeyaVes, line 23: do we need this?

packages/html/src/listview/tests/listview-loading.tsx
```diff
@@ -1,41 +1,34 @@
...
-            </div>
-
+            <span>Listview with states</span>
+            <ListView
+                header
+                footer
+                pageable
+                loading
+                className="k-d-flex">
+                <ListViewItem ><div>Listview Item</div></ListViewItem>
```
TeyaVes, line 27: remove spaces after ListViewItem here

packages/html/src/listview/tests/listview.tsx
```diff
@@ -1,69 +1,71 @@
...
-                </div>
+                <span>Listview - top pager</span>
+                <ListView
+                    header
+                    footer
+                    pageable
+                    pagerPosition="top"
+                    layout="flex"
+                    flexDirection="col">
+                    <ListViewItem ><div>Listview Item</div></ListViewItem>
```
TeyaVes, line 47: remove spaces

packages/html/src/listview/tests/listview.tsx
```diff
@@ -1,69 +1,71 @@
...
+                <span>Listview - top pager</span>
+                <ListView
+                    disabled
+                    header
+                    footer
+                    pageable
+                    pagerPosition="top"
+                    layout="flex"
+                    flexDirection="col">
+                    <ListViewItem ><div>Listview Item</div></ListViewItem>
```
TeyaVes, line 64: remove spaces

veselints (approved):

packages/fluent/scss/core/helpers/_layout.scss
```diff
@@ -16,6 +16,11 @@
     flex: 1 0 0%;
 }
 
+.k-flex-row { flex-direction: row; }
+.k-flex-row-reverse { flex-direction: row-reverse; }
+.k-flex-col { flex-direction: column; }
+.k-flex-col-reverse { flex-direction: column-reverse; }
+
```
Juveniel, line 23: This should not be added - those helper utilities are [already available in the utils package](https://github.com/telerik/kendo-themes/blob/develop/packages/fluent/scss/utils/_layout.scss), which is used in the fluent theme.

packages/fluent/scss/listview/_layout.scss
```diff
@@ -18,6 +18,14 @@
...
         -webkit-touch-callout: none;
         -webkit-tap-highlight-color: transparent;
+
+        &.k-disabled {
+            & > .k-listview-header,
+            & > .k-listview-footer,
+            & > .k-listview-pager {
+                color: inherit;
+            }
+        }
```
Juveniel, line 28: These styles are not needed. The ListView component API does not provide a way to disable the component.

packages/fluent/scss/listview/_theme.scss
```diff
@@ -8,7 +8,11 @@
             var( --kendo-listview-text, $kendo-listview-text ),
             var( --kendo-listview-bg, $kendo-listview-bg ),
             var( --kendo-listview-border, $kendo-listview-border )
-        )
+        );
+
+        &.k-disabled {
+            color: var( --kendo-listview-disabled-text, $kendo-listview-disabled-text );
+        }
```
Juveniel, line 15: These styles are not needed. The ListView component API does not provide a way to disable the component.

packages/fluent/scss/listview/_variables.scss
```diff
@@ -73,6 +73,10 @@ $kendo-listview-item-selected-text: initial !default;
 /// @group listview
 $kendo-listview-item-selected-bg: get-theme-color-var( neutral-20 ) !default;
 
+/// Text color of disabled listview.
+/// @group listview
+$kendo-listview-disabled-text: var( --kendo-disabled-text, initial ) !default;
+
```
Juveniel, line 79: These styles are not needed. The ListView component API does not provide a way to disable the component.

packages/html-spec/src/listview/listview.json
```diff
@@ -0,0 +1,98 @@
...
+    "selector": [
+        ".k-listview"
+    ],
+    "states": [
+        {
+            "name": "disabled",
+            "selector": [
+                ".k-disabled"
+            ]
+        },
```
Juveniel, line 14: Remove this, as there is no such state for the component.

packages/html-spec/src/listview/listview.json
```diff
@@ -0,0 +1,98 @@
...
+        {
+            "name": "focus",
+            "selector": [
+                ":focus",
+                ".k-focus"
+            ]
+        }
```
Juveniel, line 21: @veselints do we need the entire ListView to be focusable? According to [the current spec](https://github.com/telerik/web-components-ux/blob/master/docs/components/listview/keyboard-navigation.md) - no. If this is indeed the case, we should remove this from the html spec.

packages/html-spec/src/listview/listview.json
```diff
@@ -0,0 +1,98 @@
...
+                },
+                {
+                    "name": "vertical",
+                    "selector": [
+                        ".k-listview-borders-vertical"
+                    ]
+                }
+            ]
+        }
+    ],
```
Juveniel, line 47: I would suggest not to add the `borders` option to the spec, as this is only available in jQuery. The ThemeBuilder does not need to 'know' about this option.

packages/html/src/listview/listview.tsx
```diff
@@ -0,0 +1,91 @@
...
+    itemsCount?: number;
+    hasItemsNumbers?: boolean;
+    focusItem?: Array<number>;
+    selectedItem?: Array<number>;
+    header?: boolean;
+    footer?: boolean;
+    pageable?: boolean;
+    pagerPosition?: 'top' | 'bottom';
+    bordered?: boolean;
+    borders?: 'all' | 'horizontal' | 'vertical';
```
Juveniel, line 16: remove the `borders` prop from the html component as well

packages/html/src/listview/listview.tsx
```diff
@@ -0,0 +1,91 @@
...
+    layout?: 'flex' | 'grid';
+    flexDirection?: 'row' | 'row-reverse' | 'col' | 'col-reverse';
+    flexWrap?: 'wrap' | 'nowrap' | 'wrap-reverse';
+    gridColumns?: number;
```
Juveniel, line 20: @joneff do we want this to go into the component config ?

packages/html/src/listview/listview.tsx
```diff
@@ -0,0 +1,91 @@
...
+    disabled?: boolean;
```
Juveniel, line 21: remove the disabled prop as we do not need it

packages/html/src/listview/listview.tsx
```diff
@@ -0,0 +1,91 @@
...
+            </div>
+        </>;
+
+        return (
+            <div className={classNames(
+                className,
+                "k-listview",
+                {
+                    'k-listview-bordered': bordered,
+                    [`k-listview-borders-${borders}`]: borders,
```
Juveniel, line 68: remove borders config

packages/html/src/listview/listview.tsx
```diff
@@ -0,0 +1,91 @@
...
+                    'k-disabled': disabled
```
Juveniel, line 69: remove disabled

packages/html/src/listview/listview.tsx
```diff
@@ -0,0 +1,91 @@
...
+            flexDirection,
+            flexWrap,
+            gridColumns,
+            disabled,
+            loading
+        } = this.props;
+
+        const pager = <div className={`k-listview-pager k-pager k-pager-md k-listview-pager-${ pagerPosition }`}>
+            Pager{/* TODO - add pager component */}
+        </div>;
```
Juveniel, line 52: Use pager component here

Juveniel (changes requested):
After changes are addressed, please also take care of the commits by:

1) Adding a Breaking Change for prefixing variables. See https://github.com/telerik/kendo-themes/commit/d28ff899bac9660f069cdcf099c26b7297b37712

2) Rename - `test(listview): add html component` to `feat(html): add listview html component`

3) drop the `feat(listview): add unified rendering template` commit (see review comment)

packages/html/src/listview/listview.tsx
joneff: yes, we do. customers **love** properties

packages/html-spec/src/listview/listview.json
joneff: still waiting?

joneff (approved):

packages/html-spec/src/listview/listview.json
joneff: it shouldn't be focusable.

joneff (approved):

Issue #3963, opened 2022-09-15
Getting Warning Usage of '~' in imports is deprecated


raphaelpinel:
**Describe the bug**
I am installing Angular 14.2.2 together with @progress/kendo-theme-material 5.8.1

While installing with `npm install` 
 I am getting a lot (more than 100) of similar warnings:

`Warning: '/Users/<user-name>/<project-path>/node_modules/@progress/kendo-theme-material/scss/action-buttons/_layout.scss' imports '~@progress/kendo-theme-default/scss/action-buttons/_layout.scss' with a tilde. Usage of '~' in imports is deprecated.`

Here is the link to the webpack Sass loader where the deprecation is coming from:
[https://webpack.js.org/loaders/sass-loader/#resolving-import-at-rules](https://webpack.js.org/loaders/sass-loader/#resolving-import-at-rules)

**Affected package (please remove the unneeded items)**
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for Angular

**Build system information (please remove the unneeded items)**
* OS: [Mac]
* Node version: [16.17.0]
* NPM version: [8.15.0]

joneff:
Unlike node, sass, as of now, works with local files first and then tries to resolve the path by means of checking in specific folders or use function, both provided trough configuration.

In other words, we use tilde (`~`) to denote import from package manager folder, because it's a very obvious prefix and is not a valid path that will fallback to said function.

That's more or less the same thing stated in [sass-loader documentation](https://github.com/webpack-contrib/sass-loader/blob/master/README.md#resolving-import-at-rules).

But let's get to the question in hand -- can we make the warnings disappear? Sure, but that will require a major version. We can target it for 2023 R1, together with some other planned changes.

In the mean time, in the highly unlikely scenario that said deprecation turns to "not supported", you can simply pass an importer function `sass-loader`, which is what happens right now, internally.

raphaelpinel:
Thanks @joneff What is the planned schedule for the 2023.1 release?

joneff:
Pardon me for the delay. The themes have a slightly offset release cycle so the product themes have time to implement changes.

The themes will be released on January 9th, 2023 and the product release will follow on January 18th, 2023.

However, starting immediately after 2022 R3 SP2, which is November 28th and December 7th, respectively, we can have those changes applied in a dev version. Considering dev versions are automatically released every monday morning, a dev version with the changes could be out as soon as December 12th. **Could**.

[...]

PR #4118, opened 2022-11-23
fix(material): remove deprecated tilde import

edbzn:
@angular-devkit/build-angular has removed support for deprecated tilde import in v15 which cause the build of an application consuming `@progress/kendo-theme-material/scss/variables` to fail.


joneff:
Alas this will not fix things completely. we have planned to remove ~ imports (#3963) for 2023 R1, which should be out in mid January.

However, we might start earlier, in light of recent events...

I am closing this PR, as it is incomplete.

Issue #1380, opened 2020-03-19
Speeding up test builds


joneff:
There are a couple of bottlenecks

* dart-sass takes a lot of time to compile
* image tests take a lot of time

**Dart-sass compilation time**

To tackle this, we could leave dart-sass compilation only in the build that does not take screenshots. That should reduce the build time by about two minutes, which is a lot.

The only downside I can think of is that if there is a difference between node-sass and dart-sass output, as in computed stylistic content not actual output, we won't catch it. But then again, that would be a bug in either of the compilers.

- [x] compile dart-sass in pull request builds in branches different than master / develop
- [x] compile dart-sass in pushes in master / develop

**Image tests**

We could improve image tests time by starting the tests in parallel. We could also try if we can reduce the time between test runs, currently 200ms.

Logging to the console also takes some time and we could introduce coloring modes to know if something is without difference or with difference.

- [x] run image tests in parallel
- [x] reduce latency between image calls (#1380)
- allow various levels of logging
  - [ ] verbose (current)
  - [ ] compact color code logged message
      - red for changes
      - green for no changes and testarea screenshots
      - blue (or something else) for no changes and full page screenshots
  - [ ] summary (only differences at the end)

**Additional time saving**

We could also have chrome and firefox installed only for the test scenario that makes sense. So for branches running image tests, we need only firefox. For branches running the other tests we need only chrome. That's also going to shave off some time.

- [x] load addons only when needed

**Working locally**

That's not really a bottle neck, but locally, we don't need more than standard lerna. Continue the topic from above, we could simply have our default `package.json` for local environment and then have the other two be ci and ci-images.

- [x] use different build settings / package.json for different scenarios

/cc @eeasss @tsvetomir

joneff:
The other two require improvements in Pastshots.

PR #4125, opened 2022-11-28
Update pastshots

joneff:
* update pastshots so it works on M architecture (see gyoshev/pastshots#23 for other improvements)
* split install step into install and bootstrap steps
* use `--no-aduit` and `--no-fund` to hide messages

Visual tests should run faster...


Juveniel:
@joneff there are some unexpected changes in the generated previews as a result of the update ( e.g Switch toggle button is off by a certain amount ). Could you take a look ?

Juveniel (approved):

joneff:
Related to #1380

PR #4126, opened 2022-11-28
Try speedup screenshot creation by splitting in chunks

joneff:


Juveniel:
Currently this does not cause any significant improvement, thus closing the PR for the time being.

PR #4132, opened 2022-11-29
Update popup styles

joneff:


Juveniel (commented):
I would suggest using `fix(nouvelle): update popup styles`, as the current commit message is ambiguous.

joneff:
> I would suggest using `fix(nouvelle): update popup styles`, as the current commit message is ambiguous.

Done!

Juveniel (approved):

PR #4134, opened 2022-11-29
Consolidate CI in single file with callable workflows

joneff:
~~depends on #4136~~

After some extensive refactoring here is what we have:

* **CI** -- runs `on: push` in all branches, except **develop**, **master** and **v4-dev**
* **CI Develop** -- runs `on: push` in **develop**
* **CI Weekly** -- runs `on: schedule` at 00:20 every Monday

**CI**

Runs visual tests, linting of styles and scripts to pass checks. Apart from the visual tests, the rest of the jobs are reusable. `Status check > Lint scripts`, `Status check > Lint styles`, `Status check > Visual` will be the new checks.

In the future, we'll add accessibility reusable workflow (job) and check as well -- `Status check > A11y`. 

**CI Develop**

Runs linting of styles and scripts. Both are reusable workflows, shared with CI. When we have accessibility workflow, we'll add it here as well.

**CI Weekly**

Runts linting of styles, scripts; builds styles, scripts, docs, test pages; checks docs for disparity. Unlike CI and CI Develop, this build does not reuse workflows, because it runs against node version matrix -- v14 and v16.

When accessibility is added, we'll also add it here.

If both test matrices pass, dev npm release will be triggered. If the latter completes successfully, nuget release will be triggered.


Juveniel:
The order of the  checks does not seem right and is rather confusing, as we have checks with the same name ( e.g  `CI | Lint Scripts` and `CI | CI Lint Scripts`). The process should be easy enough to navigate and troubleshoot issues.

I would suggest combining the workflows as follows:

* CI | Build -> combines `Compile themes` workflow and any other sub-tasks needed for compilation
* CI | Lint -> combines `Lint Styles` and `Lint Scripts`
* CI | Unit -> ...
* CI | Visual -> combines `Render test pages`, `Create screenshots` and `Detect Changes` workflows

The checks should also run in that other to be easy to work with.

Juveniel (approved):

PR #4135, opened 2022-11-29
Improve dependency specification

joneff:
**Move most dependencies to root package.json**

When trying to resolve modules, node will look up, so no need to pollute packages. For themes, that means removing `sass-build`.

For html helpers, that mean leaving only react related as dependencies.

**Fix dependencies**

Previously we marked dependencies between themes with `^`. That's bad in many ways, but mostly because if someone installs an older version, say `5.0.0` of bootstrap theme, the installed dependency of default theme will be the latest possible and so will bootstrap's. We cannot guarantee that any version will work with any future version of it's dependencies...

I am guessing that's why we are seen multiple `color is not a color` reports.

**Remove stale dependencies**

There were a couple of stale / obsolete deps like fse and dss that I removed.

**Fix per theme sass.config file**

The files were using the older syntax.

---

All those changes should provide for improved build time, less dependencies to manage and more stable build artifacts.


JoomFX (approved):

KimBum:
Just found this, while searching for the reason I am having two separate bootstrap versions in my lock file. Any reason as to why you have not bumped the bootstrap dependency yet? When working with such hard lines (not allowing `~`), regular updates are important. Things have been working out for me with 5.2.3 for quite some time now and especially 5.2.2 contains vital fixes to regressions in 5.2.1 🙏

PR #4147, opened 2022-12-02
Update calendar today button rendering

JoomFX:
related to https://github.com/telerik/web-components-planning/issues/116


TeyaVes (commented):
A few things I noticed:
- the today link in the multiviewcalendar-jquery needs to be replaced with flat button as well
- maybe we should consider removing the gap between the Today and prev/next buttons as it does not seem to be needed anymore

JoomFX:
The design team confirmed that the gap should be removed and that they will update the designs.
I have just set the `$calendar-nav-gap` variable to `0` and left the code there so the user can apply whatever gap they want.

TeyaVes (approved):

PR #4148, opened 2022-12-02
Remove styles for fab shape property

TeyaVes:
The FAB shape property is obsolete and needs to be removed from the component and the demos (as it was removed from the Button).

- [ ] Kendo jQuery @telerik/kendo-jquery 
- [ ] Kendo React @telerik/kendo-react-admins 
- [ ] Kendo Vue @telerik/kendo-vue-team 
- [x] Kendo Angular


joneff (approved):
@Juveniel perhaps we should split commits for image tests and screenshots...

PR #4175, opened 2022-12-16
Replace slash with math div

joneff:
A couple of things going on:

* replace slash division with k-math-div to reduce dart-sass and sass-embedded warnings
* reuse k- functions where possible
* as it is mandatory, recreate docs


PaskalevStoyan (approved):

cosi-conda:
When will this fix released? Thank you!

jackdh:
On 6.1.0 I'm still seeing 

```
@function k-math-div( $a, $b )  {
    @return ( $a / $b );
}
```

In `dist\all.scss` which is throwing the deprecation warning.

joneff:
@jackdh that's to be expected -- that's the only instance where we have slash division. The rest of the cases use either `calc` expressions or this function.

True removal of this warning cannot happen while we still support node-sass (but not the import syntax). Only when we drop support for node-sass completely, we can target this last warning.

PR #4180, opened 2022-12-19
Various small improvements in checkbox and radio

joneff:
There is a small difference in the outline of both, but that's about it. And a couple of typo fixes.

Note: the size of the checkmark inside checkbox is expected to stay the same. There is, however, a variable that can control it.


packages/fluent/scss/checkbox/_layout.scss
```diff
@@ -8,10 +8,10 @@
     .k-checkbox {
         margin: 0;
         padding: 0;
-        width: var( --INTERNAL--kendo-checkbox-width, #{ $kendo-checkbox-width });
-        height: var( --INTERNAL--kendo-checkbox-height, #{ $kendo-checkbox-height });
+        width: var( --INTERNAL--kendo-checkbox-width, 1em );
+        height: var( --INTERNAL--kendo-checkbox-height, 1em );
         line-height: initial;
-        border-width: var( --kendo-checkbox-border-width, #{$kendo-checkbox-border-width} );
+        border-width: var( --kendo--checkbox-border-width, #{$kendo-checkbox-border-width} );
```
TeyaVes, line 14: do we need the double dash here?

packages/fluent/scss/checkbox/_layout.scss
```diff
@@ -23,17 +23,35 @@
...
 
 
     // Checkbox indicator
     .k-checkbox::before {
+        @if $kendo-checkbox-indicator-type == "pseudo" {
+            content: "";
+            width: var( --INTERNAL--kendo-checkbox-indicator-size );
+            height: var( --INTERNAL--kendo-checkbox-indicator-size );
+            border-width: 0;
+            border-radius: var( --kendo-checkbox-indicator-border-radius );
```
TeyaVes, line 40: shall we add a fallback here?

packages/fluent/scss/checkbox/_layout.scss
```diff
@@ -58,38 +82,80 @@
...
-            top: var( --INTERNAL--kendo-checkbox-indeterminate-top, #{ $kendo-checkbox-indeterminate-top });
-            left: var( --INTERNAL--kendo-checkbox-indeterminate-left, #{ $kendo-checkbox-indeterminate-left });
+        @if $kendo-checkbox-indicator-type == "image" {
+            background-image: var( --kendo-checkbox-indeterminate-image, #{ $kendo-checkbox-indeterminate-image } );
+            background-size: var( --INTERNAL--kendo-checkbox-indeterminate-size, 100% );
+        }
+
+        @if $kendo-checkbox-indicator-type == "pseudo" {
+            &::before {
+                width: var( --INTERNAL--kendo-checkbox-indeterminate-size );
```
TeyaVes, line 124: add fallback value here or where giving `--INTERNAL--kendo-checkbox-indeterminate-size`var a value

packages/fluent/scss/checkbox/_layout.scss
joneff: that will be fixed.

packages/fluent/scss/checkbox/_layout.scss
joneff: we are fine with border-radius not having a fallback, because it `initial` means `0`.

the width and height I am still thinking on it, since `initial` means `min-content`. perhaps we could add `100%`

joneff:
@TeyaVes comments addressed

PR #4181, opened 2022-12-21
Rename `$spacing` to `$kendo-spacing`

joneff:
Changes, in no particular order:

* used utils from utils package, because the built in utils had a flaw when dealing with dots...
* replaced residual slash division with either `k-math-div` or `math.div`
* prefixed residual functions with our `k-` or the proper `sass:*` namespace for fluent
* added `!default` flag where missing
* for borders, mostly, used static values of 1px or 3px
* for spacing used static value of 5px (fluent grid edit and command cells) where needed; if the value was 7, I rounded it to 8px (splitter)
* added new sizes to spacing map: `1px`, `0.5`, `1.5` ... `7.5`
* replaced named sizes like `sm`, `md`, `lg` with the appropriate one
* replaced `thin` with `.05` and simplified expressions where possible
* removed `hair` size because it was used mostly for borders


TeyaVes (approved):

packages/fluent/scss/action-sheet/_layout.scss
```diff
@@ -1,3 +1,4 @@
+@use "sass:string";
```
Juveniel, line 1: I believe this is not needed

packages/fluent/scss/avatar/_variables.scss
```diff
@@ -41,7 +41,7 @@ $_tc-avatar-matrix: (
     outline: (
         normal: ($white, 100, 100)
     )
-);
+) !default;
```
Juveniel, line 44: do we really need the default flag on private variables ?

packages/fluent/scss/action-sheet/_layout.scss
joneff: Yep. It's not. Previously, it was used for `string.unquote("max(...)")`, but then I remembered dart sass differs between css and math max.

packages/fluent/scss/avatar/_variables.scss
joneff: Probably not, but it wouldn't hurt.

PR #4182, opened 2022-12-22
fix(scheduler): add focus styles

JoomFX:
In [Angular](https://www.telerik.com/kendo-angular-ui/components/scheduler/), [React](https://www.telerik.com/kendo-react-ui/components/scheduler/) and [Vue](https://www.telerik.com/kendo-vue-ui/components/scheduler/) the whole Scheduler is focusable and `box-shadows` is added to it when focused.
When the Scheduler is focused, in Angular the `k-focus` class is added (which has custom styles in the component), while in React and Vue inline style is added instead.

This PR adds the style in question to Kendo Themes.

NOTE 1: We should discuss if we want a different `box-shadow` value for the different themes.
NOTE 2: React and Vue should use the `k-focus` class (instead of inline styles) so they can be in sync with the changes made in the themes.
NOTE 3: Should we add the `k-focus` class to one of the visual tests (probably yes)?
NOTE 4: We need this merged and available for R1 as the styles in the Angular component are removed as part of the CSP effort.

### DIFF
--- a/packages/bootstrap/scss/scheduler/_variables.scss
+++ b/packages/bootstrap/scss/scheduler/_variables.scss
@@ -8,6 +8,8 @@ $scheduler-bg: $component-bg !default;
 $scheduler-text: $component-text !default;
 $scheduler-border: $component-border !default;
 
+$scheduler-shadow: 0 .5px .5px .5px rgba(0, 0, 0, .12) !default;
+
 $scheduler-toolbar-bg: $kendo-toolbar-bg !default;
 $scheduler-toolbar-text: $kendo-toolbar-text !default;
 $scheduler-toolbar-border: $kendo-toolbar-border !default;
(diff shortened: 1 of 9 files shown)


Juveniel:
@JoomFX 

1) The exact box-shadow value for the different themes should be based on the design. Thus, please discuss this with the design team for the exact values.

2) Make sure other teams can adopt the change prior the release.

3) We should have a separate test for this.

4) Will be merged when the PR is ready, but not sooner.

JoomFX:
I'm closing this PR for now because, after an ad hoc meeting with @imasyuko and @dtopalov, we found out that there are quite some illogical and unclear things that prevent us from moving forward and require additional discussions (in both the design team and the component teams - Angular, React, Vue):

1. The Scheduler component in Angular, React and Vue is not focusable (has `tabindex=-1`) but for some reason it gets a visual indicator (box-shadow) that it is focused. We need to figure out why this box-shadows is added. Most probably it is not needed and should be removed.
2. The Scheduler component is only focusable in jQuery. But it is the only one that does not get a visual indicator (box-shadow) that it is focused.
3. Once it is clear if (and why) the Scheduler should have a visual indicator that it is focused, then and only then the Design team will be able to provide the proper design for the different themes.

PR #4076, opened 2022-10-31
Upload unified rendering

TeyaVes:
closes: https://github.com/telerik/kendo-themes-private/issues/48

### Rendering

<details>

<summary> Universal </summary>

```html
<!-- default rendering -->

<div class="k-upload k-upload-async">
  <div class="k-dropzone k-upload-dropzone">
    <div class="k-upload-button-wrap">
      <label for="upload-input"> <!-->optional element<-->
        <button class="k-upload-button k-button k-button-md k-button-solid k-button-solid-base k-rounded-md">
          <span class="k-button-text">Select files...</span>
        </button>
      </label>
      <input id="upload-input">
    </div>
    <span class="k-dropzone-hint k-hidden">Drop files here to upload</span>
    <span class="k-upload-status">
      <span class="k-icon k-i-upload"></span>
      Uploading...
    </span>
  </div>
  <ul class="k-upload-files">
    <!-- single file -->
    <li class="k-file k-file-progress">
      <div class="k-file-single">
        <div class="k-progressbar k-progressbar-horizontal">
          <span class="k-progress-status-wrap k-progress-end"></span>
          <div class="k-selected" style="width: 60%;">
            <span class="k-progress-status-wrap k-progress-end"></span>
          </div>
        </div>
        <span class="k-file-icon-wrapper">
          <span class="k-file-icon k-icon k-i-file-pdf"></span>
          <span class="k-file-state">uploaded</span>
        </span>
        <div class="k-file-info">
          <span class="k-file-name">Document.txt</span>
          <span class="k-file-size">225.68 KB</span>
        </div>
        <div class="k-upload-actions">
          <span class="k-upload-pct">60%</span>
          <button class="k-upload-action k-button k-button-md k-button-flat k-button-flat-base k-rounded-md k-icon-button">
            <span class="k-button-icon k-icon k-i-pause-sm"></span>
          </button>
          <button class="k-upload-action k-button k-button-md k-button-flat k-button-flat-base k-rounded-md k-icon-button">
            <span class="k-button-icon k-icon k-i-cancel"></span>
          </button>
        </div>
      </div>
    </li>
    <!-- multiple files -->
    <li class="k-file k-file-success">
      <div class="k-file-multiple">
        <div class="k-progressbar k-progressbar-horizontal k-hidden">
          <span class="k-progress-status-wrap k-progress-end"></span>
          <div class="k-selected">
            <span class="k-progress-status-wrap k-progress-end"></span>
          </div>
        </div>
[...]

PR #4184, opened 2022-12-23
Upload unified rendering improvements

magdalenaan:
related https://github.com/telerik/kendo-themes/pull/4076/commits

Simplify the new Upload rendering as the following:
- remove `.k-file-single` and `.k-file-multiple` wrappers
- add the `.k-file-multiple` class to the `.k-file` element

### DIFF
--- a/packages/default/scss/upload/_layout.scss
+++ b/packages/default/scss/upload/_layout.scss
@@ -82,29 +82,18 @@
             border-color: inherit;
             list-style: none;
 
-            .k-file-multiple,
-            .k-file-single {
-                width: 100%;
-                display: flex;
-                justify-content: space-between;
-            }
-
-            .k-file-single {
-                align-items: center;
-            }
-
-            .k-file-multiple {
-                align-items: flex-start;
-            }
-
             .k-file {
                 padding: $kendo-upload-item-padding-y $kendo-upload-item-padding-x calc(#{$kendo-upload-item-padding-y} + #{$kendo-upload-progress-thickness});
                 border-width: 0 0 1px;
                 border-style: solid;
                 outline: none;
                 display: flex;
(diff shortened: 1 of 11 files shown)


Juveniel:
Let's wait for @TeyaVes review here. We cannot merge this prior to R1 release, as it's just a couple of working days before the code freeze.

TeyaVes:
Looks good to me. I'm just wondering whether we should keep the k-file-single class added to k-file so that it can be targetted easily. Also, the JSON needs to be updated as well.

magdalenaan:
I discussed the case with the dev and manager of the Angular team and they are OK with removing the DOM element. The `kendo-upload-file-list-single-item` and `kendo-upload-file-list-multiple-items` Angular components will be moved to their parent \<li>. The Angular team can make such updates after the official R1 release.


>  I'm just wondering whether we should keep the k-file-single class added to k-file so that it can be targetted easily. Also, the JSON needs to be updated.

I am not sure if this is necessary as the only reason for keeping the `k-file-multiple` class is one CSS rule for [vertical alignment](https://github.com/telerik/kendo-themes/pull/4184/files#diff-dfb8e081fe71a490159169cb0cbb4939e670c430f8b972b0f6daa45daf06011fR95).

> Also, the JSON needs to be updated.

I have updated the JSON.

Juveniel:
This should not be merged prior to R1, thus marking it as draft for the time being.

Juveniel:
After a discussion in the FE team, we decided not to pursue this change at the moment, as the only benefit that will come out of this is removing one of the elements that we use to wrap the files container. Thus, the cost of updating this in all products is currently not worth pursuing. 

@dtopalov you can also close the related PR in Angular as well.

PR #4194, opened 2023-01-03
Remove spacer variable

joneff:
The spacer variable was removed in favour of the spacing map, which yields consistent results: the spacer variable was either `1rem` (kinda good) or `1em` (not good), where as the spacing map uses increments of `.25rem` or `4px` which is, in the default case, the same.

Also, I had to rename popover variables.

It's expected to have some visual tests generate new results due to `1em` being `14px`...


TeyaVes:
The popover title background is now changed in bootstrap as it has been deleted from the bootstrap_overrides file. Shall we keep it the same as it was before by changing the value of the $kendo-popover-header-bg in the bootstrap theme?

Juveniel (approved):

PR #4235, opened 2023-01-23
CI remove crhomedriver

Juveniel:
In this PR:

* use minimum of `node 16` for all workflows, as there is a bug in prior versions where `--no-optional` installs dependencies regardless of the flag.
* update `pastshots` to prevent installing optional drivers & remove direct `chromedriver` dep
* update `mark checks` on push with correct token type `Bearer`


joneff (approved):

joneff (approved):

joneff:
@tsvetomir do note that we are using `Authorization: Bearer $GH_TOKEN` instead of the previous `Authorization: token $GH_TOKEN`. If we are using the old syntax elsewhere, it may fail!

Issue #3840, opened 2022-08-17
SUGGESTION: variable for floating-label max-width


ShiftedBit:
Hi!

Suggestion: I think it would be great to have a configurable variable for the max-width of the floating label:
[floating-label max-width](https://github.com/telerik/kendo-themes/blob/9060b399c6cf3ca155acb76c6922cd50a59ae890/packages/default/scss/floating-label/_layout.scss#L21)
I would need a 100% here, but I can only hack it by overriding, so variable would be nice. 

Best,
Markus

PR #4240, opened 2023-01-24
Expose floating label max-width

NikoPenev21:
closes: #3840


TeyaVes (changes requested):

TeyaVes (approved):

Juveniel:
@NikoPenev21 squash style commits into one `feat(floating-label): expose label max-width variable` and then the docs commit as `chore: regen docs`.

PR #4249, opened 2023-01-26
Correct position of a cell selection in the Fluent theme

magdalenaan:
Position of cell selection is wrong in the Fluent theme

image

Run [this Dojo](https://dojo.telerik.com/UReCakuT) sample to reproduce the issue.


epetrow (approved):

joneff:
Ideally, we would need a test, because changes in spreadsheet are quite hard to track.

Do you think you can amend one of the existing spreadsheet tests?

PR #4250, opened 2023-01-26
Center vertically text of Spreadsheet cells

magdalenaan:
The text of the Spreadsheet cells is misaligned in all themes in the normal state and in the Fluent theme in the edit state.

Normal mode
image

Edit mode in the Fluent theme
image

Related:
https://github.com/telerik/kendo/pull/17039

Version:
v2023.1.117


packages/fluent/scss/spreadsheet/_layout.scss
```diff
@@ -203,7 +203,7 @@
     // Cell editor
     .k-spreadsheet-cell-editor {
         padding: 0 3px;
-        line-height: var( --kendo-spreadsheet-line-height, $kendo-spreadsheet-line-height );
+        line-height: 20px;
```
TeyaVes, line 206: we should rather update the variable value instead

TeyaVes (changes requested):
please check the material theme in the images, it does not look good

packages/fluent/scss/spreadsheet/_layout.scss
magdalenaan: I kept the variable and changed its value to the hardcoded 20px to be unified with the default value of the spreadsheet row height set dynamically by its JavaScript.

packages/classic/scss/spreadsheet/_variables.scss
```diff
@@ -2,7 +2,7 @@
 $kendo-spreadsheet-border-width: 1px !default;
 $kendo-spreadsheet-font-family: $kendo-font-family !default;
 $kendo-spreadsheet-font-size: $kendo-font-size-md !default;
-$kendo-spreadsheet-line-height: $kendo-line-height-md !default;
+$kendo-spreadsheet-line-height: 20px !default;
```
TeyaVes, line 5: do we need to change this value? As I see in the themes the $kendo-line-height-md = $kendo-line-height = k-math-div( 20, 14 ), which is the same

packages/bootstrap/scss/spreadsheet/_variables.scss
```diff
@@ -2,7 +2,7 @@
 $kendo-spreadsheet-border-width: 1px !default;
 $kendo-spreadsheet-font-family: $kendo-font-family !default;
 $kendo-spreadsheet-font-size: $kendo-font-size-md !default;
-$kendo-spreadsheet-line-height: $kendo-line-height-md !default;
+$kendo-spreadsheet-line-height: 20px !default;
```
TeyaVes, line 5: better use k-math-div( 20, 14 ) for consistency with the line-height

packages/default/scss/spreadsheet/_variables.scss
```diff
@@ -2,7 +2,7 @@
 $kendo-spreadsheet-border-width: 1px !default;
 $kendo-spreadsheet-font-family: $kendo-font-family !default;
 $kendo-spreadsheet-font-size: $kendo-font-size-md !default;
-$kendo-spreadsheet-line-height: $kendo-line-height-md !default;
+$kendo-spreadsheet-line-height: 20px !default;
```
TeyaVes, line 5: same as comment as classic theme

TeyaVes (changes requested):
The big cell with Invoice # .... does not look good in all the themes

Juveniel:
This should not be handled in the themes, as the styles for the cells are generated form the Spreadsheet engine. The engine itself does not work with line-height, which leads to the observed issue. Thus, any changes to this behavior should be addressed in the engine itself.

@magdalenaan closting this PR

PR #4252, opened 2023-01-27
chore(test): update column menu rendering

magdalenaan:
The `.k-popup` class was on the `.k-child-animation-container` in the reference rendering. The correct structure is as the following:
```html
<div className="k-animation-container">
    <div className="k-child-animation-container">
        <div className="k-column-menu k-popup k-grid-columnmenu-popup">
            <div className="k-columnmenu-item-wrapper">
                 ...
```

The `.k-columnmenu-item-wrapper` should be direct child of the `.k-grid-columnmenu-popup`. The redundant `<div>` between them should be removed.

related: https://github.com/telerik/kendo-themes/pull/3858


TeyaVes (changes requested):
Can you please check why the images are changed? 

The commit msg should rather be test(pivotgrid): ...

TeyaVes (approved):

PR #4265, opened 2023-02-01
fix(actionsheet): update selector to match new rendering

kikostadinov:
k-hr is no longer a direct child to k-actionsheet in the new rendering


TeyaVes (changes requested):
can you please check why the border color has changed? It should be grey as before. Also, I would suggest changing the commit msg to fix .... as this is actually a fix.

TeyaVes (approved):

PR #4270, opened 2023-02-03
Fix expand/collapse arrows

magdalenaan:
related https://github.com/telerik/kendo/issues/17078


TeyaVes (commented):
I think that the following icons need to be replaced in the tests as well, so we won't have visual differences, e.g popup-dropdowntree test

TeyaVes (approved):

PR #4285, opened 2023-02-13
Integrate module system in themes

joneff:
First the small bits:

* removed markdown files from source folders, as they did not contribute to anything and were, for the most part, empty;
* added a dependency to sass-unused to verify that I didn't mess something up with the refactoring. at a later point, we can use sass-unused for linting
* aligned the usage of module system in nouvelle theme. that did slow down compilation a bit, but only a bit

and now the good bits:

## What is the module system?

The module system is an abstraction over standard sass functionality that would eventually allow for "import the theme once, output whatever you like" -- in other words, it serves as a centralized entry point for each and every theme when it comes to customization and output.

Our current approach is flawed at best: the theme files are interconnected in a bad way and components import their dependencies, which leads to massive overhead in terms of IO and CPU time. For instance, building components from source can take about a minute! We do have optimizations in place that facilitate our builds and speeds up local development by factor of 60 or so. Still our customers don't use it and pay the 60 seconds price...

The module system, in contrast, allows for describing the dependencies and then including their output in the global output. That way most files are `@import`ed only once and a few others several times.

Note: files are still `@import`ed many times!

Another difference between our current approach, and the module system is that the current approach is implicit i.e. import the files -> styles; where as the module system is explicit i.e. import -> invoke -> styles. There is a built in mechanism for auto invoking, however, it will be removed over time, say for 2024 R1 (v7).

## How to use it?

Two major things of note: 
* the explicit api will be available with 2023 R1 SP1
* the exact api may change slightly 

```scss
// Importing the entire theme
// prior this pr
@import "@progress/kendo-theme-NAME/dist/all.scss";
// module system stage 0 (this pr)
@import "@progress/kendo-theme-NAME/dist/all.scss";
// stage 1 (2023 R2)
@import "@progress/kendo-theme-NAME/dist/all.scss";
// stage 2 (2023 R3)
@import "@progress/kendo-theme-NAME/dist/all.scss";
// stage 3 (2024 R1) (importing all.scss should still work, but this is the preferred way)
@import "@progress/kendo-theme-NAME/index.import.scss";
@include kendo-theme--config( $config: () );
@include kendo-theme--styles();

// Importing individual components
// prior this PR
@import "@progress/kendo-theme-NAME/scss/grid/_index.scss";
@import "@progress/kendo-theme-NAME/scss/window/_index.scss";
// stage 0
@import "@progress/kendo-theme-NAME/scss/grid/_index.scss";
@import "@progress/kendo-theme-NAME/scss/window/_index.scss";
// stage 1
$kendo-auto-bootstrap: true;
@import "@progress/kendo-theme-NAME/scss/grid/_index.scss";
@import "@progress/kendo-theme-NAME/scss/window/_index.scss";
// stage 2 (individual files still work)
@import "@progress/kendo-theme-NAME/index.import.scss";
@include kendo-theme--config( $config: ( $components: ( "grid", "window" ) ) );
@include kendo-theme--styles();
// stage 3 (individual files do not output styles be default)
@import "@progress/kendo-theme-NAME/index.import.scss";
@include kendo-theme--config( $config: ( $components: ( "grid", "window" ) ) );
@include kendo-theme--styles();
```

## What's next

**Those are forward looking statements. Take them with the healthy dose of salt and skepticism. Forward looking = wishful thinking...**

For 2023 R1 SP1 we'll release the module system with the auto invoke flag turned on. Currently it's called `$kendo-auto-bootstrap` but we can change it.

For 2023 R2, the flag will be turned off by default when importing individual component files. Still, importing the entire theme will not change.

For 2023 R3, the component files will no longer import their dependencies. Meaning that even with the auto invoke on `@import THEME/grid` will output just the styles for the grid and nothing more. The correct way to interface with the theme would be to import the index file and specify that only grid is required. This will be a big jump in performance.

For 2024 R1 (v7), component files will have no external file dependencies i.e. you will need to import the core module first, in order to import individual files. We'll also change the name of the index files (entry points) to `index.import.scss` for files that use `@import` syntax and `index.scss` for those that use `@use`. It's important to note that `*.import.scss` will be slightly different from what the sass spec states they should be.

Beyond 2024 is hard to guesstimate what may or may not happen, but ultimately, we'll drop `node-sass`, say after 2025 R1 (v8) and the `*.import.scss` files will be aligned to the sass spec. As long as the sass compiler supports those files, we'll have in the distribution. Probably and a couple of more years after that!

Just to reiterate: those dates are not set in stone! Depending on the actual circumstances we may prolong a phase, or squeeze phases closer. The version in brackets are also tentative. As a component vendor we depend on many external factors and the need for incompatible behaviour (breaking change) may arise from something seemingly unrelated to a developer's tech stack.

Still that's the timeline: integrate with on be default; then off by default; then cleanup imports; then change file names; then drop node-sass. 

One other minor, as of now, thing -- compilation is slowed down by about a second or so for all themes. I am not quite sure why this happens, but this time will improve over time


packages/material/scss/treelist/_layout.scss
```diff
@@ -1,6 +1,8 @@
 @import "@progress/kendo-theme-default/scss/treelist/_layout.scss";
 
-@include exports("treelist/layout/material") {
+
+@mixin kendo-treelist--layout() {
+
```
TeyaVes, line 5: missing @include kendo-treelist--layout-base(); here. I guess this will also fix the differences in the visual tests

joneff:
@Juveniel kendo-theme--config is currently just stubbed. we haven't yet talked about how themes should be customized, hence the mixin is only a stub.

Currently, components should be passed as a separate variable:

```scss
$kendo-components: ( "grid", "treeview" );
$kendo-auto-bootstrap: false;

@import "@progress/kendo-theme-NAME/scss/all.scss"
```

packages/material/scss/treelist/_layout.scss
joneff: thank you!

Juveniel (approved):

TeyaVes (approved):

PR #4290, opened 2023-02-14
fix(pivotgrid): chip menu not properly positioned

JoomFX:
related to https://github.com/telerik/kendo-angular-private/issues/354

### DIFF
--- a/packages/default/scss/chip/_layout.scss
+++ b/packages/default/scss/chip/_layout.scss
@@ -29,6 +29,11 @@
         .k-selected-icon-wrapper {
             display: none !important; // sass-lint:disable-line no-important
         }
+
+        // Adjustment for kendo-icon-wrapper
+        .k-icon-wrapper-host {
+            display: initial;
+        }
     }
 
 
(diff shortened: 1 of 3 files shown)


Juveniel:
@JoomFX can we avoid the custom element `kendo-pivot-chip-menu` in the selector. These tend to cause issues in the long run. The preferred approach is to use a class instead ( could be applied on the element in Angular if necessary ).

JoomFX:
@Juveniel This is the selector we agreed on yesterday with @joneff .
This is the exact thing that we discussed (that we are trying not to use the Angular custom elements in the selectors) but we wanted to scope the selector to the exact element.

Anyway, an alternative selector is the following. Let me know if you want me to commit it:

```
.k-pivotgrid-configurator-content .k-chip .k-icon-wrapper-host {
    display: initial;
}
```

JoomFX:
I agree that handling it in the Chip iteself ( `.k-chip .k-icon-wrapper-host` ) is probably a better solution.
I just tested it and it seems that this approach will not cause any issues/misalignments.

Juveniel (approved):

Issue #3444, opened 2022-03-07
Review the usage of repositioning the parent selector (&)


JoomFX:
**Describe the bug**
A good example are the Slider styles where we have [this selector](https://github.com/telerik/kendo-themes/blob/e1edfd7446141203ec22d38e14033b6eaeb9bced/packages/default/scss/slider/_layout.scss#L343).
When importing the Kendo theme in a class (in order to create a dark theme for example) it is possible that a wrong selector is created.

This import:

```
.dark-theme {
  @import "~@progress/kendo-theme-default/scss/dist/all.scss";
}
```

Might generate the following selector (probably depending on the compiler - node-sass vs dart-sass), which is invalid:

```
.k-slider-horizontal .dark-theme .k-slider-track,
.k-slider-horizontal .dark-theme .k-slider-selection
```

**Expected behavior**
We need to investigate this further and figure out if reworking the Kendo theme styles is a viable solution or if there is another workaround.

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**
* All

**Build system information (please remove the unneeded items)**
* Not Applicable

```[tasklist]
### Tasks
- [x] #4293 
- [x] #4295 
- [ ] #4298 
```

hypetsch:
We are currently also facing this issue. Just if someone also stumbles across it - adding those scss rules fixed it for us:

```scss
.k-slider-horizontal .k-slider-track {
   height: $kendo-slider-track-thickness;
   left: 0;
   margin-top: -$kendo-slider-track-thickness / 2;
   top: 50%;
}

.k-slider-vertical .k-slider-track {
   bottom: 0;
   left: 50%;
   margin-left: -$kendo-slider-track-thickness / 2;
   width: $kendo-slider-track-thickness;
}
```

PR #4293, opened 2023-02-15
Use logical text align

joneff:
Use `start` and `end` instead of `left` and `right` and overrides for RTL. This is probably the easiest part of #3444.

Notes on screenshots:

* calendar: in rtl when week number column is showed, week day labels are now aligned correctly
* scheduler: in month view, day label is aligned correctly
* scheduler: in day / week view, times are now aligned correctly


PaskalevStoyan (approved):

epetrow (approved):
Looks good to me :)

PR #4294, opened 2023-02-16
Unify rendering Scheduler Toolbar Views

epetrow:
Closes: https://github.com/telerik/kendo-themes-private/issues/79


JoomFX:
Can we leave the `k-scheduler-views-wrapper` element? I mean not to remove it.
Currently, all suites have the wrapper so the rendering is unified in that perspective.
I also do not see much benefit of removing it.

Yes, we will have one wrapper less in most suites, but in Angular we will still have a host element (wrapper). All we can do is to remove the class but this does not bring any benefit at all (the wrapper will still be there).
Here's the code in question - [link](https://github.com/telerik/kendo-angular-private/blob/develop/libs/scheduler/src/toolbar/view-selector.component.ts#L66).

@TeyaVes @veselints

epetrow:
@JoomFX Since either the ButtonGroup or Select will be rendered I do not think there is much reason to keep the `k-scheduler-views-wrapper`. If previously it was used to contain both elements now it is not needed anymore. However if there are angular specific limitations that would prevent removing it we agreed that only the class can be removed.

JoomFX:
@epetrow alright, in Angular we will only remove the class.
It is also [a public component](https://www.telerik.com/kendo-angular-ui/components/scheduler/api/ToolbarViewSelectorComponent/) so we cannot refactor it in a way that it does not have its wrapper.

veselints:
The suggestion to remove the `.k-scheduler-views-wrapper` wrapper element is related to the use of the ToolBar component within the Scheduler. In jQuery the ToolBar has its own ButtonGroup item (which is basically the ButtonGroup component rendered in the ToolBar container). If we the `.k-scheduler-views-wrapper` has to be present in the rendering, that would force us to use a template with a ButtonGroup in it instead of the default ButtonGroup already available in the ToolBar. As the wrapper element no longer serves any styling needs, its removal would allow us to use the built-in tool (ButtonGroup) instead of defining custom template for this specific usage.

Juveniel (approved):

PR #4297, opened 2023-02-17
Component specs

TeyaVes:
related to: https://github.com/telerik/kendo-themes-private/issues/81


Juveniel (approved):

tests/action-sheet/action-sheet-list.html
```diff
@@ -260,9 +260,6 @@
                                                     <span class="k-list-item-text">List item 3.3</span>
                                                 </li>
                                             </ul>
-                                            <div class="k-height-container">
-                                                <div></div>
-                                            </div>
```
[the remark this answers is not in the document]
TeyaVes: the k-height-container is added only when a list has virtualization:

`{ virtualization && <div className="k-height-container"><div></div></div> }`

in the following example this is not the case, so imo the container shouldn't be present

packages/html/src/action-sheet/action-sheet.spec.tsx
```diff
@@ -0,0 +1,123 @@
...
+                case ActionSheetFooter:
+                    return false;
+                default:
+                    return true;
+            }
+            return true;
+        })
+        : children.type === ActionSheetItems
+            ? children
+            : <></>;
```
kspeyanski, line 75: Is this based on the current implementation of the `ActionSheet`?

Let's revisit this at a later stage. Overall I would advocate against different rendering modes based on the children or their type (e.g. `string` vs `component`). Seems to be present in multiple components (e.g. List, Page, OrgChart)

packages/html/src/colorpalette/colorpalette.spec.tsx
```diff
@@ -0,0 +1,85 @@
...
+                    .map((color) => {
+                        items.push( <ColorPaletteTile color={color} tileSize={tileSize} /> );
+                    });
+
+                const row = <ColorPaletteRow>{items}</ColorPaletteRow>;
+
+                newChildren.push(row);
+            });
+        }
+    }
```
kspeyanski, line 60: The whole `presets` seems to specific to be part of the rendering spec.

Can we extract this and pass it from outside as a `children` or `data` prop?

packages/html/src/datepicker/datepicker.spec.tsx
```diff
@@ -0,0 +1,117 @@
...
+                <InputLoadingIcon {...props} />
+                <InputClearValue {...props} />
+                <InputSuffix>{suffix}</InputSuffix>
+                <Button
+                    className="k-input-button"
+                    icon="calendar"
+                    rounded={null}
+                    size={size}
+                    fillMode={fillMode}
+                />
```
kspeyanski, line 102: Can we filter the properties which are passed to child component? One of the benefits of the spec would be to give a clear flow of customization properties.

We can stick to the option properties only, as they are the ones that bring value to reading the spec. (e.g. answering the question "if the grid has size='small' should all dropdowns inside have their size modifeid").

I believe this is present on multiple occasions and is not specific to the DatePicker spec only.

packages/html/src/dropdowngrid/dropdowngrid.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+                focus={focus}
+                valid={valid}
+                invalid={invalid}
+                required={required}
+                loading={loading}
+                disabled={disabled}
+                className={classNames(
+                    props.className,
+                    DROPDOWNGRID_CLASSNAME,
+                    'k-combobox'
```
kspeyanski, line 94: Do we intent to use an `input` with `k-combobx` instead of the `Combobox` spec itself? 

While this might not have an effect on the actual rendering of the component, It would imply the decision when developing the component.

packages/html/src/datepicker/datepicker.spec.tsx
```diff
@@ -0,0 +1,117 @@
...
+        required,
+        loading,
+        disabled,
+        opened,
+        ...other
+    } = props;
+
+
+    return (
+        <>
```
kspeyanski, line 75: The `DatePicker` does not use the `Picker` component, while the DropDownList and other dropdowns are using it. I'm not really sure which one would be correct, but can we please double-check that?

kspeyanski (approved):
Kudos to the team for pushing this! Great initiative with a lot of potential and benefits once we adopt it across the suits.

I'm leaving my notes. Feel free to merge the PR without addressing it, as they are mostly questions/long-term.

packages/html/src/datepicker/datepicker.spec.tsx
TeyaVes: the DatePicker is an input with a button, while the `Picker` component is part of these components which use span with classes `k-picker k-picker-solid`, etc. It may seem confusing, but we are basing the input/picker component names on the classes used for styling

packages/html/src/dropdowngrid/dropdowngrid.spec.tsx
TeyaVes: IMO, the `combobox` component includes both the input part as well as the list inside popup part. The `dropdowngrid` on the other hand, consists of an input + grid inside popup. This is why I haven't used the `combobox` spec, which is the same only for the input part. We are using the `k-combobox` class for styling purposes. Does it make sense or not really?

kspeyanski: With the current capabilities of the suits it makes perfect sense. I do believe there is an opportunity there to further improve the "ComboBox" in order to facilitate replacing the content inside the popup - e.g. with a Grid and achieve the DropDownGrid rendering.

PR #4300, opened 2023-02-17
Add styles for toolbar position and orientation and reuse them in other components

joneff:
Orientation: horizontal and vertical
Position: top, bottom, left and right

Those are styles that deal with border placement (`border-width`) and item orientation (`flex-direction`). I've stayed away from "direction", because it might be confused with ltr / rtl direction. Instead I've stayed closed to aria spec, which defines aria-orientation.


JoomFX (changes requested):
In the Material theme the `$kendo-toolbar-border-width` variable is set to `0` (zero), which is the reason for the generated screenshots.
It is a real issue, mostly visible in the Grid ([before screenshot](https://raw.githubusercontent.com/telerik/kendo-themes/ce7b79b7ea0465420200f2abdeb3f682c9d713f6/tests/_output/material/grid/grid-filter-row-angular.png), [after screenshot](https://raw.githubusercontent.com/telerik/kendo-themes/526225f517f53b03e6bcd6137b3c18513e6501b7/tests/_output/material/grid/grid-filter-row-angular.png)), since the bottom border is missing.

joneff:
This PR is outdated. Closing.

PR #4310, opened 2023-02-22
Inner dropzone should not be clickable

epetrow:
`.k-dropzone-inner` is not clickable so the `cursor: pointer` style should be removed


TeyaVes (changes requested):
the cursor: pointer rule should be removed from the fluent theme as well

TeyaVes (approved):

Issue #4323, opened 2023-03-01
Kendo bug reports related to rendering unification and SVG icon util


pepinho24:
Here are some of the issues we encountered after the rendering unification and the introduction of the SVG icon, related to https://github.com/telerik/kendo/issues/16788#issuecomment-1448308977

- [x] 1) Stepper with a disabled element, the line is visible over the icon element. If the `filter: grayscale(0.1)` style is removed, the issue disappears - http://kendobuild-windows.dev.progress.com/staging-test-pages-mvc/KendojQuery/stepper/?theme=default-main

- [x] 2) Slider - SVG icons are not rotated in RTL, seems to be missing transformation styles for RTL - http://kendobuild-windows.dev.progress.com/staging-test-pages-mvc/KendojQuery/slider/?theme=default-main

- [x] 3) ScrollView, SVG icons are small and misplaced - the font icons have `display: table-cell` and `font-size: 4.5em` which is missing for the SVG icon - http://kendobuild-windows.dev.progress.com/staging-test-pages-mvc/KendojQuery/scrollview/?theme=default-main

- [x] 4) TaskBoard - the icons in the Header have `color: inherit` which leads to the icons being blue, not black. Is that the new correct behavior or an issue? The same is observed with the Close button in a Tooltip and the Toggle arrows in Panelbar http://kendobuild-windows.dev.progress.com/staging-test-pages-mvc/KendojQuery/taskboard/?theme=bootstrap-main

PR #4325, opened 2023-03-02
Fix svg icons position

Juveniel:
targets: https://github.com/telerik/kendo-themes/issues/4323


TeyaVes (approved):
the tabstrip carets will be fixed with the following commit: https://github.com/telerik/kendo-themes/pull/4322

PR #4329, opened 2023-03-02
fix(treelist): no spacing between the elements in the angular treelist filter menu

JoomFX:
There is no spacing between the elements in the angular treelist filter menu.
The issue can be seen in [this example](https://www.telerik.com/kendo-angular-ui/components/treelist/filtering/reusable-filter/#toc-filter-menu-with-popup) in the Kendo Angular Trelist documentation.

treelist


Juveniel:
I think we can improve this if we add a dedicated class to the Angular host element, and use it in the themes. Otherwise, we have to add selectors that target custom elements indefinitely. @JoomFX thoughts ?

JoomFX:
@Juveniel I do know that we are trying not to use custom Angular elements as CSS selectors but in my opinion it is not a viable option for all cases/scenarios.
Adding a custom CSS class means double-work for (almost) nothing. In this particular case we need to modify 9 Angular components which are part of different components (Grid and Treelist).
And this is just the tip of the iceberg. Search in the themes repo for custom Angular selectors and you will find them everywhere. Does that mean that we need to add a custom CSS class to all those Angular elements and then update the selectors in the themes?

In my opinion not using the custom Angular elements as selectors should be recommended but not a must.
In cases like this one it should be acceptable to simply extend the current styles with custom Angular selectors instead of doubling our work by first changing the rendering (adding a CSS class) of the components and then refactoring the styles to use CSS classes as selectors.

This should definitely be discussed on a higher/broader level so we can agree on how to proceed from now on.

@dtopalov @stnikolova

Juveniel:
Yes, the recommended approach is not to style custom elements in the themes, as this leads to multiple issues. Here are a few:

* Diverging the rendering of components ( e.g [here](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/grid/_layout.scss#L780-L815) ).
* Duplicating styles that we already have ( e.g the current set of changes ).
* Bug fixes should be done with these specifics in mind, otherwise we have to go back and fix the same bugs all over again.
* Expanding the adoption of this bad practice.  

Taking the above into consideration, extending the styles with the current approach is not really a solution, as it rather shifts the double work to the themes framework.

Furthermore, going for unified rendering across components, we should definitely tackle and resolve this issue on a larger scale. If you have any other suggestion than the proposed one, please share it.

Juveniel:
Merging this for the release. However, we should discuss this further and refactor once we agree on the best approach to handle these cases.

Juveniel (approved):

PR #4332, opened 2023-03-07
Edit Variable Descriptions

yordan-mitev:
As the descriptions of the Sass variables are taken directly from the kendo-themes repo and then are used in the Design System docs, we need to edit them and apply some minor changes in the style.


Juveniel:
@yordan-mitev we should transfer these updates to the variables for all themes, as they are the same ( e.g. packages/bootstrap, packages/material, etc. ).

If you currently don't have the time to address this, please let me know so that I can schedule it in the team.

Juveniel (approved):

PR #4335, opened 2023-03-08
Extend Utils package

magdalenaan:
related https://github.com/telerik/kendo-themes-private/issues/84
related https://github.com/telerik/kendo-themes-private/issues/77

DO NOT MERGE until the `utils-extend-docs` branch is rebased with this one, taking the last changes.

### DIFF
--- a/packages/utils/scss/_variables.scss
+++ b/packages/utils/scss/_variables.scss
@@ -102,14 +102,32 @@ $kendo-utils: (
         video: k-string-unquote("16 / 9")
     ),
     "container": (),
-    "columns": (),
+    "columns": (
+        1: 1,
+        2: 2,
+        3: 3,
+        4: 4,
+        5: 5,
+        6: 6,
+        7: 7,
+        8: 8,
+        9: 9,
+        10: 10,
+        11: 11,
+        12: 12,
+        auto: auto
+    ),
     "break-after": (),
     "break-before": (),
     "break-inside": (),
     "box-decoration-break": (),
-    "box-sizing": (),
+    "box-sizing": (
+        border: border-box,
+        content: content-box
+    ),
     "display": (
         none,
+        contents,
         block,
         inline,
         inline-block,
@@ -119,6 +137,8 @@ $kendo-utils: (
         inline-grid,
         table,
         inline-table,
+        table-row,
+        table-cell,
         list-item
     ),
     "float": (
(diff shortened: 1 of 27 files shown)


packages/utils/scss/_mixins.scss
```diff
@@ -26,3 +26,9 @@
         }
     }
 }
+
+@mixin generate-utils-multiple( $map, $values, $function: "", $important: $kendo-important) {
+    @each $side in $map {
+        @include generate-utils( $side, $side, $values, $function: "", $important: $kendo-important );
+    }
+}
```
Juveniel, line 34: As discussed in our offline meeting, please drop this mixin and define mappings in the `_variables` file. Then, use the standard `generate-utils()` mixin to generate the classes.

packages/utils/scss/layout/_placement.scss
```diff
@@ -0,0 +1,108 @@
...
+/// This is equivalent to `left: 0;`.
+/// @name .k-pos-left
+/// @group position
+/// @contextType css
+
+@mixin kendo-utils--layout--placement() {
+    // Top / right / bottom / left
+    $kendo-utils-placement: k-map-get( $kendo-utils, "placement" ) !default;
+    $kendo-utils-placement-props: (top, right, bottom, left) !default;
+    @include generate-utils-multiple( $kendo-utils-placement-props, $kendo-utils-placement );
```
Juveniel, line 31: use `generate-utils()` here as well

packages/utils/scss/layout/_visibility.scss
```diff
@@ -5,12 +5,6 @@
...
     // Visibility utility classes
     $kendo-utils-visibility: k-map-get( $kendo-utils, "visibility" ) !default;
-    @include generate-utils( visibility, visibility, $kendo-utils-visibility );
-
-    // Legacy aliases
-    .#{$kendo-prefix}visible { @extend .#{$kendo-prefix}visibility-visible !optional; }
-    .\!#{$kendo-prefix}visible { @extend .\!#{$kendo-prefix}visibility-visible !optional; }
-    .#{$kendo-prefix}invisible { @extend .#{$kendo-prefix}visibility-hidden !optional; }
-    .\!#{$kendo-prefix}invisible { @extend .\!#{$kendo-prefix}visibility-hidden !optional; }
+    @include generate-utils( null, visibility, $kendo-utils-visibility );
```
Juveniel, line 9: we should probably not remove those ,as they are being used in the components

Juveniel (changes requested):
There are some unexpected changes in the visual previews. Please take a look and resolve them as well.

joneff:
What motivates the need to change the mixin that generates the utils? In the way it's currently changed, not setting `$kendo-prefix: null` or empty space, together with passing null as a name of utils would generate a class name starting with a dash. Something I am not particularly keen on.

packages/utils/scss/_variables.scss
```diff
@@ -3,7 +3,14 @@ $kendo-important: true !default;
...
 
-$kendo-font-sizes: () !default;
+$kendo-font-sizes: (
+    xs: 10px,
+    sm: 12px,
+    md: 14px,
+    lg: 16px,
+    xl: 20px
+
+) !default;
```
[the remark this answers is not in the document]
magdalenaan: The `$kendo-font-sizes` map is used to generate utilities for font-size - similar to the usage of the `$kendo-border-radii` map.

packages/utils/scss/common/_reset.scss
```diff
@@ -0,0 +1,75 @@
+// TODO: docs
+
+@mixin kendo-utils--common--reset() {
+
+    @if $kendo-reset {
```
joneff, line 5: I think this should be in the core package, not the utils.

joneff (commented):
@Juveniel @magdalenaan  that's my only pondering -- should the reset be part of core or utils. to me, it makes sense to be part of core.

packages/utils/scss/common/_reset.scss
Juveniel: Hmm, indeed, we should make sure to move this to the Core package, as there are arbitrary styles. @magdalenaan could you extract this in a separate PR perhaps, so that we can move forward with this one ?

Juveniel (changes requested):

Juveniel (approved):

packages/utils/scss/common/_reset.scss
Juveniel: resolved by extractring reset styles in a separate PR

PR #4358, opened 2023-03-30
Dropdowntree popup

TeyaVes:
All component's popups should have the following class- **.k-COMPONENT-popup**, e.g .k-dropdowntree-popup. We'll keep the class legacy class k-popup-dropdowntree in the themes until R3.

All products need to make sure that the popups of all dropdowns and pickers have the correct class name.

related to: https://github.com/telerik/web-components-planning/issues/208

@ag-petrov, @nenchef , @Raisolution , @jivanova , @stnikolova , @zdravkov


JoomFX:
@TeyaVes In Angular the `k-popup-dropdowntree` class is used in both the [DropDownTree](https://www.telerik.com/kendo-angular-ui/components/dropdowns/dropdowntree/) and the [MultiSelectTree](https://www.telerik.com/kendo-angular-ui/components/dropdowns/multiselecttree/).
According to [this issue](https://github.com/telerik/kendo-themes-private/issues/68), it should be the same case for React.

We can remove the `k-popup-dropdowntree` class but then we should add the same styles for both `.k-dropdowntree-popup` and `k-multiselecttree-popup`:

```
.k-popup-dropdowntree, // remove legacy class once updated in components
.k-dropdowntree-popup,
.k-multiselecttree-popup {
    ....
}
```

TeyaVes:
@JoomFX , I've added the .k-multiselecttree-popup selector

PR #4387, opened 2023-04-06
SVG icons spec

TeyaVes:
closes: https://github.com/telerik/kendo-themes-private/issues/76

fixes: https://github.com/telerik/kendo-themes/issues/4390

A few notes:
- The breadcrumb text underline does not work for SVG icons, this can be tackled in the item for breadcrumb and tabstrip rendering unificiation
- The text-shadow of the Scrollview icons is not applied when SVG icons are used


inikolova:
I observed the following differences - maybe some of them are intentional, however could we check :) 

- Chip with _icon_ and _small_ size is higher now;
- Badge with icon (Rectangle, Rounded, Pill, Circle) -> icon is not centred;
- Bottom navigation -> icons are smaller (16px/16px);

Juveniel (approved):

PR #4392, opened 2023-04-10
chore: export  form specs

kspeyanski:
Exporting the `defaultProps` due to different defaults per component. 

We will use them in the configurator of the Design System Docs, to populate the default values:
image


TeyaVes (approved):

TeyaVes:
looks good to me, I'll just remove the two commits updating the visual previews

PR #4395, opened 2023-04-12
Column Menu: Tabbed Interface

JoomFX:
closes https://github.com/telerik/kendo-themes-private/issues/57

The task is to fit the Tabstrip component in the Column Menu Popup (to provide rendering and make styling adjustments).
None of the current Column Menu elements have been modified.
Therefore, there are some expected differences between this implementation and what is shown in the designs:

1. In the Default and Bootstrap themes the top left and top right edges of the Popup have to have a border-radius, the same as the one in the Tabstrip active item.
2. The scrollable section in the "Column Visibility" tab is higher than the one in the design.
3. In the Material theme, the Tabstrip has a border bellow the items. That's why this implementation also has a border, while there is no border in the design. Again, the task is to fit the current Tabstrip component (with its current design) in the Column Menu Popup, without modifying the Tabstrip design (and the design of any other of the Column Menu elements).


packages/default/scss/grid/_layout.scss
```diff
@@ -1221,6 +1221,32 @@
...
+
+        .k-tabstrip-items {
+            margin: (-$kendo-tabstrip-item-border-width) (-$kendo-tabstrip-item-border-width) 0;
+
+            .k-item {
+                flex: 1;
+            }
+
+            .k-link {
+                height: $kendo-grid-column-menu-tabbed-tabstrip-item-height;
```
TeyaVes, line 1239: I would suggest increasing the vertical padding, instead of adding variable for the height and setting a height. What if we increase the size of the icon, but the height is set

JoomFX: We decided that overriding the tab height is not needed, therefore the code in questions was deleted.

Juveniel (approved):

PR #4444, opened 2023-04-28
Implement DropDownList option label design

epetrow:
closes: https://github.com/telerik/kendo-themes-private/issues/75

https://progresssoftware.atlassian.net/browse/FE-523


packages/html/src/list/list.spec.tsx
```diff
@@ -100,6 +102,7 @@ export const List = (
                     ['k-virtual-list']: virtualization,
                 }
             )}>
+            { optionLabel }
```
inikolova, line 105: Shouldn't we remove the spacing for consistency?

Juveniel (approved):

PR #4462, opened 2023-05-03
Spec & Utils Templates + tests dev server

kspeyanski:
The following PR is a prototype continuation https://github.com/telerik/kendo-themes/pull/4297 and distributes a bigger pieces of JSX markup, built on top of the component specs. Working name - `templates`. In addition, It implements a `dev-server` to allow easier development of such said templates which builds and hosts everything from the dist/*/tests folder of the `html` package.

The current available [button](https://github.com/telerik/kendo-themes/tree/spec-templates/packages/html/src/button/templates) templates are:
- IconButton
- IconTextButton
- TextButton

The current available [button-group](https://github.com/telerik/kendo-themes/tree/spec-templates/packages/html/src/button-group/templates) templates are:
- IconButtonGroup
- IconTextButtonGroup
- TextButtonGroup
- MixedButtonGroup

and for the [dropdownlist](https://github.com/telerik/kendo-themes/tree/cfaf2ed84727205a551ffb9ba0f86b72db67d71d/packages/html/src/dropdownlist/templates):
- DropdownlistFiltering
- DropdownlistGrouping
- DropdownlistGroupingFiltering
- DropdownlistNormal
- DropdownlistPopup

A couple of notes here:
- I do not anticipate to distribute the `html` itself, but rather a transpiled version of the code. The [JSX syntax](https://github.com/telerik/kendo-themes/blob/cfaf2ed84727205a551ffb9ba0f86b72db67d71d/packages/html/src/dropdownlist/templates/dropdownlist-grouping-filtering.tsx) is flexible enough to generate HTML from it. Checkout the following snippet.

> The usage of `React` is absolutely optional, and any JSX compatible renderer should be able to produce the `html`, but it's really convenient to use `renderToStaticMarkup`:
 
```ts
import ReactDOM from "react-dom/server";
import React from "react";
import { IconButton } from "../../button"; // to become '@progress/kendo-themes-html'

const html = ReactDOM.renderToStaticMarkup(React.createElement(IconButton, {}));
```

Outputs:
```html
<button
  class="k-button k-button-md k-button-solid k-button-solid-base k-rounded-md k-icon-button"
>
  <span class="k-button-icon k-svg-icon k-svg-i-folder"
    ><svg
      aria-hidden="true"
      focusable="false"
      xmlns="http://www.w3.org/2000/svg"
      viewBox="0 0 512 512"
    >
      <path d="..." /></svg
  ></span>
</button>
```

- My gut feeling is we might want to keep component options (`fillMode`, `size`, `rounded` and `themeColor`) and `states` **_out_** of the 'templates' (name TBD). The main reason is that they would introduce an excesive amount of variations which have to be hardcoded, as well as how easy is it to generate them on demand, due to their first-level support from the `.spec` file of the component:

```tsx
import { Button, IconButton } from "../../button";

Button.options.fillMode.map((fillMode) => {
  const html = ReactDOM.renderToStaticMarkup(React.createElement(IconButton, { fillMode }));
});

Button.states.map((state) => {
  const html = ReactDOM.renderToStaticMarkup(React.createElement(IconButton, { [state]: true }));
});
```


tsvetomir:
Not shipping reference markup sounds tempting, but it shifts the responsibility of enumerating the valid configurations to the consuming package. You might get into a situation where suite A and suite B test different things. It's also not guaranteed that all combinations of options make sense.

More complex components may combine not only options, but sub-components (e.g. Toolbar with overflow) that can't be tested by simply enumerating options. So just shipping the bare-bone components might not cut it, you still may need manually curated demos to validate the rendering.

Whether to ship such test cases as JSX or rendered HTML is also questionable. There's definitely some waste and repetition in shipping the markup. On the other hand, consuming them in test should be really simple.

> The usage of React is absolutely optional, and any JSX compatible renderer should be able to produce the html, but it's really convenient to use renderToStaticMarkup.

You really sure about that? The generated code has a hard-coded dependency on `react/jsx-runtime`. The `react-jsx` transform seems be more about [removing the dependency on React in the source code](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html), it's not some kind of industry standard. That's not to say it's not an option, but it would without doubt require a dependency on React v17+.

kspeyanski:
Thank you @tsvetomir for jumping into the conversation.

> Not shipping reference markup sounds tempting, but it shifts the responsibility of enumerating the valid configurations to the consuming package. You might get into a situation where suite A and suite B test different things. It's also not guaranteed that all combinations of options make sense.

I'm still strong on the opinion that we should **not** ship the markup itself, but rather something that generates it easily. A quick math of the simplest button component would reveal the need for 13200 (5 states * 3 sizes * 4 roundness options....etc.) markups predefined. Are you thinking a hybrid approach where we're programmatically generating the markup prior to distribution?

> More complex components may combine not only options, but sub-components (e.g. Toolbar with overflow) that can't be tested by simply enumerating options. So just shipping the bare-bone components might not cut it, you still may need manually curated demos to validate the rendering.

Complex components are something we're actively working on. Our first iteration was to introduce a bigger set of "specs" called templates (part of this PR) which distributes a more-complex scenario requiring no further configuration. Here the configuration work has to be done by the platform consuming the test, e.g.:

```js
const expected = jsxToHTML(DropdownListPopup);
const actual = renderPlatformComponent();
actual.click('.k-button'); // To opening the popup

expect(actual).toMatch(expected);
```

Given our optimistic expectation to be able to produce markup in a Unit test (as opposed to an e2e one) you should not be required to create a separate demo, but rather only configure it on an unit-test level.

Furthermore, the JSX syntax would actually enable us to provide whole sub-components as a regular property.

```js
const expected = jsxToHTML(DropdownListPopup, { Popup: ModifiedPopup }); // <DropDownListPopup Popup={ModifiedPopup} />
const actual = renderPlatformComponent();
actual.click('.k-button'); // To opening the popup

expect(actual).toMatch(expected);
```

> Whether to ship such test cases as JSX or rendered HTML is also questionable. There's definitely some waste and repetition in shipping the markup. On the other hand, consuming them in test should be really simple.

Consuming the JSX in a test case is as simple as we make it, currently (assuming React is a JSX renderer) it's a three-liner, and can easily be extracted into a utility function:

```js
import React from 'react';
import ReactDOM from 'react-dom/server';

const html = ReactDOM.renderToStaticMarkup(React.createElement(DropdownList, {}));
```


> You really sure about that?...

I'm sure it wont work out of the box - however I'm sure we've taken the necessary precautions to enable such functionality if the need arise. The only "replacement" I managed to make is to replace "react" with "preact" through a custom esbuild/webpack resolver.

On the side note, the main thing we're trying to avoid by hiding the "React" part in the specs is introducing confusion that this specs are an actual react components. Imagine someone not on-board with the effort hearing that we're using react and being left with the impression we're rewriting the KendoReact suit or something in those lines.

Hope that helps clarify some of the points, but please let us know if further questions arise, as we're actively seeking feedback on this effort.

Juveniel (approved):

PR #4489, opened 2023-05-16
feat(html): add animation-container and reuse it in tests

Juveniel:


epetrow (commented):
A few comments from my end:

- I noticed that there is a difference in the box-shadow of the popups, is it expected?
- The added Grid test `grid-column-reordering` seems to be missing an `.html` file

epetrow (approved):
LGTM

Issue #4502, opened 2023-05-25
Typography: `k-{element}` styles overwrite the styles of additional utility classes


kikostadinov:
The styles of utility classes `k-m-*`, `k-font-weight-*` and `k-fs-*` are overwritten by the element classes.

The problem can be reproduced in the demo available here:
https://www.telerik.com/kendo-react-ui/components/utils/typography/appearance/

silviyaboteva:
The issue starts from `v6.2.0` where the utility classes are imported prior to the components' styles. Detaching all of the utilities from the themes will be the solution.

PekoPPT:
The issue is no longer replicable and all appearance demos will be returned to the React documentation.

PR #4509, opened 2023-05-29
fix: utils import

kspeyanski:
Related to https://github.com/telerik/kendo-themes/issues/4502 

Seems like `utils` package should be self-contained, no need to import it into each individual component - actually used mixins are coming from the `core` package. Otherwise the import order does not allow overriding specific component styles when an utility is used.

IMO, if we are to distribute the utils together with the theme, their styles should always be loaded at the bottom of the stylesheet. If we're to face inability to style something due to specificities, it would be a different story, but simple-selectors should be easily be styled with any utility class.

cc, @kikostadinov &mdash; I think https://github.com/telerik/kendo-react-private/pull/2463 would be unnecessary after this fix. Thoughts?


kikostadinov:
@kspeyanski hey, yeah this should be enough.

Juveniel:
@kspeyanski can we go forward with moving the import to the end of the dist, but extract the component imports removal for the major release ( R3 ), as we need to also alter docs in docs-seed, katana and DS, and also mark this as an important change for users?

kspeyanski:
Let me try keeping the imports while still forcing the output to have the utils at the end of the file. 

Will follow up once I try it!

kspeyanski:
@Juveniel &mdash; does not seem to be possible due to how the compiler works. When it sees that more than one file  depends on the same module (in our case the `utils` module) it always puts it first in the output as it's treated as a dependency.

zdravkov:
Should we proceed with some fixes or temporary fixes in the ui suites regarding this issue?

kspeyanski:
@zdravkov I'm afraid the proposed fixes in the UI suites (using the `!k-` important prefix) would only introduce another problems, so I cannot advise further. If a workaround for the specific issues has been found, we should probably take no actions at this point.

Juveniel:
This is resolved with detaching the utilities package from the core theme

PR #4518, opened 2023-06-01
chore: reuse core module in all themes

kspeyanski:
The following PR aims to re-use the `core` module across _all_ themes by making it both `@import`-able and `@use`-able. This requires some moving around of variables so they are available in their respective dependents, but overall should not affect the output of the themes.

Most (if not all) image differences are due to fixed input-color in the bootstrap theme affecting both text and icons, due to the usage of the _correct_ color contrast function - this is a fix.


JoomFX (approved):
Awesome!

Extracting the Color System from Fluent into Core is a huge amount of work and a great starting point for the overall "Color System" task. Thank you @kspeyanski !

Juveniel (approved):

Juveniel:
part of: https://github.com/telerik/kendo-themes-private/issues/128

Issue #4510, opened 2023-05-29
[Scheduler] Wrong border color of the Dialog Titlebar in the popup edit form in Angular


JoomFX:
**Describe the bug**
The border of the Dialog Titlebar in the Scheduler popup edit form in Angular inherits a wrong color.

**To reproduce**

1. Go to [this Stackblitz example](https://stackblitz.com/edit/angular-m335cv?file=src%2Fapp%2Fapp.component.ts).
1. Edit one of the existing events or try to create a new event.

**Expected behavior**
The border should get the correct color.

**Screenshots**
![border](https://github.com/telerik/kendo-themes/assets/7753940/8071c8ee-3327-403b-9c66-594d49a71c4f)

**Affected package (please remove the unneeded items)**
* theme-default

**Affected suites (please remove the unneeded items)**
* Kendo UI for Angular

**Affected browsers (please remove the unneeded items)**
* All

PR #4528, opened 2023-06-02
fix(scheduler): wrong border color of the dialog titlebar in the popu…

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/4510


Juveniel:
@JoomFX Currently the Scheduler edit dialog is rendered inside the component, thus causing the observed issue. Instead of patching this in the themes, I would rather suggest to implement the fix in the Scheduler component, where the edit dialog should be rendered outside of the Scheduler - similar behavior is already implemented for [the Grid external edit](https://www.telerik.com/kendo-angular-ui/components/grid/editing/external-editing/).

JoomFX:
@Juveniel Yes, the root cause of the issue is that the Dialog is rendered inside the Scheduler component.
This is a limitation in Angular, we have to contain everything in the component itself.
The [Grid example](https://www.telerik.com/kendo-angular-ui/components/grid/editing/external-editing/) you are referring to is just an example where the Edit Dialog is a separate/stand-alone component that is not built-in in the Grid component. It does not come as a feature of the Kendo Angular Grid component.

In my opinion [the current styles](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/scheduler/_layout.scss#L25) that we have are fine.
The problem is that the `.k-scheduler-edit-dialog` element is not a direct child of `.k-scheduler`. It is not a direct child because it is a separate (internal) component and each and every Angular component has a wrapper.

Juveniel (approved):

PR #4531, opened 2023-06-07
Unify rendering Skeleton

kikostadinov:
closes https://github.com/telerik/kendo-themes-private/issues/142


packages/html/src/skeleton/skeleton.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+        ...other
+    } = props;
+
+    return (
+        <span
+            {...other}
+            className={classNames(
+                props.className,
+                SKELETON_CLASSNAME,
+                'k-skeleton-',
```
epetrow, line 35: Remove this as it results in `k-skeleton-` being rendered as an additional class

packages/html/src/skeleton/tests/skeleton.tsx
```diff
@@ -10,25 +10,24 @@ const style = `
...
+            <Skeleton shape={'circle'} style={{ width: '50px', height: '50px' }} />
+            <Skeleton shape={'rect'} style={{ width: '300px', height: '200px' }} />
 
             <div className="k-skeleton-container k-d-grid">
-                <span className="k-skeleton k-skeleton-circle" style={{ gridColumn: "1 / 3", gridRow: "1 / 3" }}></span>
-
-                <span className="k-skeleton k-skeleton-text" style={{ gridColumn: "3 / 10", gridRow: "1 / 2" }}></span>
-                <span className="k-skeleton k-skeleton-text" style={{ gridColumn: "3 / 7", gridRow: "2 / 3" }}></span>
+                <Skeleton shape={'circle'} style={{ gridColumn: "1 / 3", gridRow: "1 / 3" }} />
+                <Skeleton shape={'text'} style={{ gridColumn: "3 / 10", gridRow: "1 / 2" }} />
```
epetrow, line 24: Since we pass `text` as  a defaultProp  do we need to define it here and below?

epetrow:
Also rename the last commit to `"chore(html): update captcha spec"`

packages/default/scss/skeleton/_layout.scss
```diff
@@ -22,13 +22,6 @@
     }
 
 
-    // Legacy alias
-    .k-placeholder-line {
-        @extend .k-skeleton !optional;
-        @extend .k-skeleton-text !optional;
-    }
```
Juveniel, line 29: This is used in the Grid component virtualization. If you want to deprecate the legacy alias, you need to also update the other tests, where the Skeleton should be used, and provide product guidelines for the change in the themes issue.

kikostadinov: I've decided to revert the change, since it is affecting another component and will require more time to resolve.

Juveniel: We still want to do this change and streamline the usage of the Skeleton, as this is the essence of the task itself.

JoomFX (changes requested):
Regarding [this commit](https://github.com/telerik/kendo-themes/pull/4531/commits/d0c1cef615a93a7cf6ae157d43bfa8f3ee18d02b) - please replace the Skeleton rendering (`<span className="k-skeleton k-skeleton-text k-skeleton-pulse"></span>`) with the actual component (`<Skeleton />`).

JoomFX (approved):

epetrow (approved):
looks good to me

packages/default/scss/grid/_layout.scss
```diff
@@ -90,7 +90,7 @@
                 outline: none;
             }
 
-            .k-placeholder-line {
+            .k-skeleton {
                 width: $kendo-grid-cell-placeholder-line-width;
             }
```
Juveniel, line 95: Do we really need to enforce 80% width for the Grid ? We can verify this with the design team and act accordingly.

Juveniel (approved):

PR #4468, opened 2023-05-04
Update drag and drop to match design guidelines

epetrow:
Closes: https://github.com/telerik/kendo-themes-private/issues/39

PR #4532, opened 2023-06-07
Unify rendering TileLayout

kikostadinov:
closes https://github.com/telerik/kendo-themes-private/issues/137


epetrow:
Just a few things I noticed so far:

- Why are we removing the `<div class="k-card-title></div>` element from the rendering? If it is intentional this leads to a smaller font size of the header -> `14px` which differs from the design -> `16px`
- Since we are unifying the rendering, do we need product specific tests e.g. `tilelayout-react`?
- If we are removing the `k-cursor-move` class from the TileLayout header in the reference rendering maybe it is a good idea to add a separate test to showcase it? It was changed from `k-cursor-grab` as part from [this](https://github.com/telerik/kendo-themes/pull/4468) effort
- I don't think we should add both `k-resize` and `k-resize-x, k-resize-y` classes on the same element e.g. `k-resize k-resize-x`
- In general when passing string props, we tend to wrap them in double quotes e.g. `columnWidth="1fr"`

packages/html/src/tilelayout/tests/tilelayout.tsx
```diff
@@ -1,93 +1,29 @@
...
+                rowHeight="1fr"
+                gap="16px"
+                padding="16px"
+                style={{ width: '800px' }}
+            >
+                <TileLayoutItem header="Title goes here" body="Content goes here" style={{ order: "0", gridColumnEnd: "span 1", gridRowEnd: "span 1" }} focus />
+                <TileLayoutItem header="Title goes here" body="Content goes here" style={{ order: "1", gridColumnEnd: "span 1", gridRowEnd: "span 1" }} />
+                <TileLayoutItem header="Title goes here" body="Content goes here" style={{ order: "2", gridColumnEnd: "span 1", gridRowEnd: "span 2" }} />
+                <TileLayoutItem header="Title goes here" body="Content goes here" style={{ order: "3", gridColumnEnd: "span 1", gridRowEnd: "span 1" }} />
+                <div className="k-layout-item-hint" style={{ order: "4", gridColumnEnd: "span 1", gridRowEnd: "span 1" }}></div>
```
Juveniel, line 19: we could extract this as a component e.g `<TileLayoutItemHint />`

tests/_output/bootstrap/tilelayout/tilelayout-resizing.png
Juveniel, line 1: the non resizable Card seems to be removed from the test - is this intentional ?

tests/_output/bootstrap/tilelayout/tilelayout.png
Juveniel, line 1: overall the test looks very different from the initial one, especially the layout hint. Could we use similar markup here ?

JoomFX (approved):

Juveniel (approved):

PR #4540, opened 2023-06-12
chore: add combobox templates

inikolova:
Adds common templates for the Combobox component


packages/html/src/combobox/templates/combobox-popup.tsx
```diff
@@ -0,0 +1,16 @@
+import { List, ListItem } from "../../list";
+import Combobox from "../combobox.spec";
+
+export const ComboboxPopup = (props) => (
+    <Combobox
+        opened={true}
+        popup={
+            <List>
+                <ListItem>List item</ListItem>
+                <ListItem selected>List item</ListItem>
```
kspeyanski, line 10: Let's try to keep templates as simple as possible. In this case the `selected` is obsolete, and there is no indication from the name of the template, that the second list item will be selected.

If needed to replicate some of the current tests, we can pass additional props to the actual template (e.g. `selected={[1]}`) to further configure the template to also add selection.

Juveniel: I would suggest to add the states tests to the `List` component templates/tests ? If we have state tests for the List there is no need to add similar once for other components. Thoughts ?

Juveniel (approved):

Issue #4556, opened 2023-06-16
SassError: Index out of range


JoomFX:
**Describe the bug**
There are several users experiencing this issue after upgrading from an older Kendo Themes version:

```
./src/styles.scss - Error: Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
SassError: Index out of range
   ╷
23 │     @return function-exists( $name );
   │             ^^^^^^^^^^^^^^^^^^^^^^^^
   ╵
  node_modules\@progress\kendo-theme-core\scss\functions\_meta.import.scss 23:13     k-meta-function-exists()
  node_modules\@progress\kendo-theme-utils\scss\_mixins.scss 4:19                    generate-utils()
  node_modules\@progress\kendo-theme-utils\scss\layout\_aspect-ratio.scss 29:5       kendo-utils--layout--aspect-ratio()
  node_modules\@progress\kendo-theme-utils\scss\layout\index.import.scss 17:5        kendo-utils--layout()
  node_modules\@progress\kendo-theme-utils\scss\index.import.scss 25:5               kendo-utils()
  node_modules\@progress\kendo-theme-default\scss\utils\_index.scss 26:9             @content
  node_modules\@progress\kendo-theme-core\scss\module-system\index.import.scss 35:9  module-render()
  node_modules\@progress\kendo-theme-default\scss\utils\_index.scss 25:5             kendo-utils--styles()
  node_modules\@progress\kendo-theme-default\scss\utils\_index.scss 31:5             @import
  node_modules\@progress\kendo-theme-default\scss\icons\_index.scss 2:9              @import
  src\styles.scss 4:9                                                                root stylesheet
``` 

Some of the users were able to work around the issue by changing the order of their `@import`'s, but it is a bug in Kendo Themes that was introduced in v6.2.0.

Here is a list of some of the users (tickets and issues) that experience this issue:
https://github.com/telerik/kendo-themes/issues/4231
ticket 1612296
ticket 1612746
ticket 1606635
ticket 1605984

**To reproduce**
I was able to reproduce the issue by importing Angular Material Theming before Kendo Theme Default in the **styles.scss** file of a brand new Angular application (most probably there are other `@import`'s causing the issue, it is not limited to Angular Material Theming only):

```
@import '@angular/material/theming';
@import "@progress/kendo-theme-default/dist/all.scss";
```

If you change the order of the imports (if you import Kendo Theme Default first) the issue is not present.

[...]

PR #4557, opened 2023-06-16
fix(core): k-meta-function-exists breaks the compilation is some specific scenarios

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/4556

Another possible solution is to replace the default value of the `$function` parameter [here](https://github.com/telerik/kendo-themes/blob/develop/packages/utils/scss/_mixins.scss#L1) with something else, but not an empty string. It might be something like `$function: "false"` like this:

```
@mixin generate-utils( $name, $props, $values, $function: "false", $important: $kendo-important ) { ... }
```


Juveniel (approved):

bedwards-cms:
I am having this issue, so I am hoping next release will fix it?  I can't use the ocean-blue-a11y theme at the moment. 
image

bedwards-cms:
I confirmed that the code added in this PR did resolve my issue.

PR #4560, opened 2023-06-22
Adaptive pager

magdalenaan:
closes: https://github.com/telerik/kendo-themes-private/issues/161


packages/html/src/grid/tests/grid-adaptive-pager.tsx
```diff
@@ -12,7 +12,7 @@ export default () =>(
                         <div className="k-view k-widget">
                             <div className="k-stretched-view k-content">
                                 <div className="k-grid k-grid-md k-grid-display-block k-grid-mobile k-editable">
-                                    <Pager className="k-grid-pager" mobile="small" />
+                                    <Pager className="k-grid-pager" adaptive info={false} pageSizes={false} />
```
Juveniel, line 15: are info and pageSize always hidden in adaptive mode ?

Juveniel: i suggest to leave it as is for now

tests/_output/default/pager/pager-adaptive.png
inikolova, line 1: According to the design, there shouldn't be any margins around the page dropdown/select element and the Prev/Next buttons in the smallest size

inikolova: After offline sync with the designers, there is a decision that the PageSizes dropdown won't be visible; also there will be margins around the page dropdown/select element and the Prev/Next buttons. The design has been updated https://www.figma.com/file/VCIOKDxHOw7qYkC4ks5Ctp/Adaptive-Story?type=design&node-id=2015-210910&mode=design&t=EhnykUsopnUJ2wVB-0

inikolova (approved):

Juveniel (approved):

Issue #1929, opened 2020-09-29
Conversational UI improvements


silviyaboteva:
**Describe the solution you would like to see implemented**
- [x] Based on client feedback (support ticket #1485002), the attachments need to be aligned based on the user who attaches them. 

Currently:
<img width="412" alt="Screenshot 2020-09-29 at 11 09 08" src="https://user-images.githubusercontent.com/17604855/94530814-4c88d680-0244-11eb-9499-c00d56576fe4.png">

Proposed solution:
<img width="408" alt="Screenshot 2020-09-29 at 11 08 48" src="https://user-images.githubusercontent.com/17604855/94530842-5c081f80-0244-11eb-99bc-a425efd0dbb6.png">

- [ ] Based on this issue (https://github.com/telerik/kendo-ux-private/issues/88), attachments rendered in the Chat view does not display User related information (name of the user and avatar). The user name and avatar could be displayed on the top of the card deck.
![72417715-1c66d980-3782-11ea-8ea4-0b6cdb9882be](https://user-images.githubusercontent.com/17604855/94531090-b3a68b00-0244-11eb-9d81-ecb8b7cd0ae3.png)

inikolova:
In v12.0.0 of the themes the Chat component was fully redesigned and its rendering and styles were improved. The first point in this thread was addressed with the Chat revamp, the second one is related to a design issue, which is closed now. Therefore I am closing this issue.

PR #4568, opened 2023-07-01
Unify Chat rendering

epetrow:
closes: https://github.com/telerik/kendo-themes-private/issues/140

Targets first bullet from https://github.com/telerik/kendo-themes/issues/1929#issue-710876110, second one needs design.


inikolova:
Almost everything looks good to me. However, I have concerns with [align attachments based on user who attaches them](https://github.com/telerik/kendo-themes/pull/4568/commits/278632a9c0d2295c290752d70b75e6341ac1705e)- this is not an universal solution and will work only in specific cases (for example it won't work in Angular, as there is an additional host element). My suggestion is to add an additional class for the _alt_ attachments and use it in the CSS selector.

Juveniel (approved):

PR #4600, opened 2023-07-07
Update lockfileVersion to 2

epetrow:


joneff (approved):
There are two separate things in this PR, but it's ok.

PR #4605, opened 2023-07-10
chore: add multiselect templates

inikolova:
Adds common templates for the MultiSelect component and simplify tests


packages/html/src/multiselect/tests/multiselect-solid.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+                    <div dir="rtl">
+                        <MultiSelectChipActionRemove { ...{ [state]: true }}>{state}</MultiSelectChipActionRemove>
+                    </div>
+                </>
+            ))}
+
+            <div>invalid + focus</div>
+            <div>
+                <MultiSelectChipActionRemove invalid focus>invalid + focus</MultiSelectChipActionRemove>
```
Juveniel, line 50: This looks like a MultiSelect with a value - `MultiSelectChipActionRemove` sounds rather confusing to me in terms of what the template aims to showcase.

inikolova: I agree :) I renamed the template to _multiselect-value.tsx_ and respectively the component to **MultiSelectValue**

Juveniel (approved):

PR #4610, opened 2023-07-12
chore: add switch templates

inikolova:
Adds common templates for the Switch component and simplify tests


packages/html/src/switch/switch.spec.tsx
```diff
@@ -11,6 +11,7 @@ const states = [
 
 const options = {
     size: [ Size.small, Size.medium, Size.large ],
+    roundness: [ null, "sm", "md", "lg", "full" ]
```
Juveniel, line 14: use from misc - [see Button for ref](https://github.com/telerik/kendo-themes/blob/develop/packages/html/src/button/button.spec.tsx#L16)

packages/html/src/switch/switch.spec.tsx
```diff
@@ -11,6 +11,7 @@ const states = [
 
 const options = {
     size: [ Size.small, Size.medium, Size.large ],
+    rounded: [ null, Roundness.small, Roundness.medium, Roundness.large, Roundness.full ]
```
Juveniel, line 14: do not use null here, but rather in the templates same as the button. 
Btw, the label for the same option should be 'none', as that is what the component API name is.

Juveniel (approved):

PR #4614, opened 2023-07-13
chore: add list templates

inikolova:
Adds common templates for the List component and simplify tests


Juveniel (approved):
LGTM, we could probably start with these templates( standard spec & angular spec ) and then try to unify both in a separate effort.

@kspeyanski thoughts ?

kspeyanski:
@Juveniel - correct! Having both specs is crucial at the moment so we can implement the actual tests. Once the tests are in-place we can push the rendering unification through them.

PR #4621, opened 2023-07-15
Color system POC

JoomFX:
This is a basic POC for the Color System implementation.

### Requirements
As per the specification ([here](https://github.com/telerik/kendo-themes-private/issues/126) and [here](https://github.com/telerik/kendo-themes-private/issues/127)) there are some main requirements:

- The core of the color system should be extracted in the `kendo-theme-core` package (a big part of this requirement is completed with [this PR](https://github.com/telerik/kendo-themes/pull/4518))
- Each theme should implement theme specific palettes (the currect state of the palettes can be seen [here](https://github.com/telerik/web-components-ux/issues/769))

### The Approach
The main goal of this approach is to try to keep the current architecture of the themes unchanged or to introduce as little changes as possible.
We already have a basic Color System in the Fluent theme. If we try to implement it in the other themes we will not need to do any major changes in the Fluent theme.
Here is a very basic, top-level overview of the main steps:

1. Implement the Color System with Palettes in each theme
2. Update the global `_variables.scss` file as well as all other `_variables.scss` files in each theme so they do not use hardcoded colors but colors from the palettes instead.

SIDE NOTE: Since we want to have theme-specific palettes we might want to re-introduce the palettes and variables in the Fluent theme ([commit](https://github.com/telerik/kendo-themes/pull/4518/files#diff-ad9ed92d20b1b941b716b128199591cb7a518d450bcd5d6c4e70690441a58e15)).

At this point all themes should have a basic Color System, the same (or very similar) to the one in the Fluent theme.
From here on the Color System itself can be extended and developed further. We already have some very useful [functions](https://github.com/telerik/kendo-themes/blob/develop/packages/core/scss/color-system/_functions.import.scss). New ones can be added whenever there is a need for that.

### Next steps
Once we have the basic Color System implemented in all themes, the next step would be to make sure that no colors outside of the palettes are used. In order to do that we should either drop some of the color functions that we currently use  or modify them so they return another color from the palettes.
As a reference, in the Fluent theme we do not use any of the color functions (`k-color-tint`, `k-color-shade`, `k-try-shade`, etc, which all boil down to the `k-color-level` and `k-color-mix` functions).

Another step is to convert the styles of all themes so they can be customized with CSS vars as well (not only with SCSS variables). We could use the approach from the Fluent theme (passing a CSS var with SCSS var as a fallback - `var( --kendo-color-primary, #{$kendo-color-primary} )`).

### Swatches
Since the main goal of this approach is to keep the current architecture mostly unchanged, the swatches will be build and used as they currently are. We would still rely on the `$kendo-is-dark-theme` toggle.


Juveniel:
@JoomFX should we close this in light of the variables refactoring task ?

JoomFX:
@Juveniel yep, I'm closing it.

PR #4622, opened 2023-07-17
chore: add treeview templates

inikolova:
Add common templates for the TreeView component and simplify tests


packages/html/src/treeview/templates/treeview-checkboxes.tsx
```diff
@@ -0,0 +1,28 @@
+import { Treeview, TreeviewItem, TreeviewGroup } from "../../treeview";
+
+export const TreeviewCheckboxes = (props) => (
+    <Treeview
+        children={[
+            <TreeviewItem text="Root 1" showCheckbox />,
```
Juveniel, line 6:
```suggestion
            <TreeviewItem text="Root 1" showCheckbox />
```

packages/html/src/treeview/templates/treeview-icons.tsx
```diff
@@ -0,0 +1,30 @@
+import { Treeview, TreeviewItem, TreeviewGroup } from "../../treeview";
+
+export const TreeviewIcons = (props) => (
+    <Treeview
+        children={[
+            <TreeviewItem text="Root 1" />,
```
Juveniel, line 6:
```suggestion
            <TreeviewItem text="Root 1" />
```

tests/_output/bootstrap/treeview/treeview.png
Juveniel, line 1: the root item in RTL scenarios is not spaced properly, as there needs to be a checkbox there

Juveniel (changes requested):

packages/html/src/treeview/templates/treeview-checkboxes.tsx
inikolova: The _TreeviewItem_ are defined as children, so I think we need the comma

packages/html/src/treeview/templates/treeview-icons.tsx
inikolova: The _TreeviewItem_ are defined as children, so I think we need the comma

PR #4623, opened 2023-07-17
chore: add time selector templates

inikolova:
Adds common templates for the Time selector component and simplify tests


packages/html/src/time-selector/index.ts
```diff
@@ -2,3 +2,6 @@ export * from './time-selector.spec';
 export * from './time-selector-column';
 export * from './time-selector-header';
 export * from './time-selector-fast-selection';
+export * from './templates/time-selector-all';
```
Juveniel, line 5: how about `time-selector-normal` ?

inikolova: I am not sure, as there aren't any default columns; _time-selector-all_ defines all the possible columns

Juveniel (approved):

PR #4637, opened 2023-07-27
Global box-shadow

magdalenaan:


magdalenaan:
In this state of this draft pull request, I offer the following:
1) As we discussed on the "Effect & Typography Utils Classes" meeting, the Utils package should use Default variables for shadows - even if the theme is not added to the page. This requires adding the Default theme dependency to itself and to the Utils and Fluent packages.
   I was considering to move shadows variables to the Core package, to avoid circular dependencies (Default -> Utils -> Default), but this would compromised the consistency of the workflow. Furthermore we discussed with @kspeyanski that as it is about library, this solution is OK. For this reason I suggest leaving the variables in the Default theme package.

2) Adding a new Shadow module in each theme, which will consume its shadow [**map** variable](https://github.com/telerik/kendo-themes/blob/global-utils/packages/default/scss/shadows/_variables.scss). Every key-value pair corresponds to different shadow depth (elevation) and consists of list of shadows with its depth and color. As there are some few cases when we need to change some of the default color, I have added [function ](https://github.com/telerik/kendo-themes/blob/global-utils/packages/core/scss/functions/_shadow-color.import.scss)for customizing it. Structure of the variables map allows to use generate-utils() mixin from the Utils package, so we can generate utilities classes in every theme with the corresponding variables.

I would appreciate any feedback and comment.

magdalenaan:
In the meeting with the FE team, we decided to continue with the solution where we have default variables for global systems placed in the Core package. So the development will continue in the [global-shadow-core](https://github.com/telerik/kendo-themes/tree/global-shadow-core) branch.

PR #4638, opened 2023-07-31
Dock manager styles

epetrow:
closes: https://github.com/telerik/kendo-themes-private/issues/88


packages/default/scss/dock-manager/_layout.scss
```diff
@@ -0,0 +1,231 @@
...
+        display: flex;
+        flex-flow: row nowrap;
+        position: relative;
+        border-width: $kendo-dock-manager-border-width;
+        border-style: $kendo-dock-manager-border-style;
+
+        .k-pane,
+        .k-splitbar {
+            box-sizing: border-box;
+        }
```
Juveniel, line 14: can we extract this in the splitter styles ?

packages/default/scss/dock-manager/_layout.scss
```diff
@@ -0,0 +1,231 @@
...
+
+        &::before {
+            height: 0;
+        }
+    }
+
+    .k-toolbar-left {
+        border-width: 0 $kendo-dock-manager-border-width 0 0;
+    }
+
```
Juveniel, line 111: probably address this with https://github.com/telerik/kendo-themes/pull/4300/commits/ce7b79b7ea0465420200f2abdeb3f682c9d713f6

packages/default/scss/dock-manager/_layout.scss
```diff
@@ -0,0 +1,231 @@
...
+        border-radius: $kendo-dock-manager-dropping-area-border-radius;
+    }
+
+    .k-tabstrip-prefix {
+        display: inline-flex;
+        padding: $kendo-dock-manager-tabstrip-prefix-padding;
+        align-items: center;
+        justify-content: center;
+        cursor: move;
+    }
```
Juveniel, line 147: cane we move this to the TabStrip styles ?

packages/default/scss/dock-manager/_theme.scss
```diff
@@ -0,0 +1,77 @@
...
+
+    // DockManager
+    .k-dock-manager {
+        border-color: $kendo-dock-manager-border;
+    }
+
+    .k-splitbar-horizontal,
+    .k-splitbar-vertical {
+        border-color: $kendo-dock-manager-splitbar-border;
+    }
```
Juveniel, line 11: just use `border-color: inherit` - see filemanager

packages/default/scss/dock-manager/_layout.scss
```diff
@@ -0,0 +1,231 @@
...
+
+        .k-splitbar-horizontal {
+            border-inline-width: $kendo-dock-manager-splitbar-border-width;
+            border-inline-style: $kendo-dock-manager-splitbar-border-style;
+        }
+
+        .k-splitbar-vertical {
+            border-block-width: $kendo-dock-manager-splitbar-border-width;
+            border-block-style: $kendo-dock-manager-splitbar-border-style;
+        }
```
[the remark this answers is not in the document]
epetrow: I don't think we can since we have to apply either block or inilne border width and style depending on the orientation of the Splitbar

packages/default/scss/dock-manager/_layout.scss
```diff
@@ -0,0 +1,231 @@
...
+
+    }
+
+        // Needed because of a bug in Chrome - Issue 1473554
+        // A fix will be shipped likely in milestone 118 and then we can revisit this
+        @supports  (not (-ms-ime-mode: none)) and (not (overflow: -webkit-marquee)) and (not (-moz-appearance:none)) {
+            .k-dock-manager-toolbar .k-toolbar-button {
+                padding-block: $kendo-dock-manager-toolbar-button-padding-y;
+                padding-inline: $kendo-dock-manager-toolbar-button-padding-x;
```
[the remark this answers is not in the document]
epetrow: We need the variables for the time being otherwise the buttons will not be the correct size -- they do not share the same padding across all themes. What I've done now is ommit the `!default` option so they are not customizable since we won't be keeping these styles for long. Let me know what you think.

packages/default/scss/dock-manager/_layout.scss
```diff
@@ -0,0 +1,231 @@
...
+        .k-tabstrip-content,
+        .k-splitter {
+            border-width: 0;
+        }
+
+        .k-toolbar .k-separator-horizontal {
+            width:  $kendo-line-height-em;
+            height: 0;
+            border-width: $kendo-dock-manager-separator-border-width 0 0;
+        }
```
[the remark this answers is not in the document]
epetrow: Actually they are needed because [here](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/toolbar/_layout.scss#L81) we explicitly state that the separator in the Toolbar should be vertical

packages/default/scss/dock-manager/_layout.scss
```diff
@@ -0,0 +1,231 @@
...
+        border-style: $kendo-dock-manager-border-style;
+
+        .k-pane,
+        .k-splitbar {
+            box-sizing: border-box;
+        }
+
+        .k-pane-flex {
+            flex-direction: column;
+        }
```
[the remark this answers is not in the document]
epetrow: I think these classes set the direction of the Splitter itself and how it's children will be layed out

packages/bootstrap/scss/dock-manager/_variables.scss
```diff
@@ -0,0 +1,113 @@
+///  The width of the border around the DockManager component.
+/// @group dock-manager
+$kendo-dock-manager-border-width: 1px !default;
+/// The style of the border around the DockManager component.
+/// @group dock-manager
+$kendo-dock-manager-border-style: solid !default;
```
inikolova, line 6: In other components we don't add a variable for border-style - instead it is directly set to the component. Example: 
https://github.com/telerik/kendo-themes/blob/8b90f3aef2d8742d9fea94474dabb38bd7553c1d/packages/default/scss/window/_layout.scss#L7

packages/default/scss/dock-manager/_index.scss
```diff
@@ -0,0 +1,38 @@
...
+$_kendo-module-meta: (
+    name: "dock-manager",
+    dependencies: (
+        "splitter",
+        "toolbar",
+        "tabstrip"
+    )
+);
+
+// Dependencies
```
inikolova, line 14: Shouldn't we add @import "../common/_index.scss"; as a dependency?

packages/default/scss/dock-manager/_layout.scss
```diff
@@ -0,0 +1,230 @@
...
+
+        .k-pane-tabbed {
+            padding-block-start: $kendo-dock-manager-tabbed-pane-padding-y;
+            padding-block-end: $kendo-dock-manager-tabbed-pane-padding-x;
+            padding-inline: $kendo-dock-manager-tabbed-pane-padding-x;
+        }
+
+        .k-splitbar {
+            border-width: 0 1px;
```
inikolova, line 122: Maybe we could use a variable for the splitbar border width instead of hardcoding it?

packages/bootstrap/scss/dock-manager/_variables.scss
Juveniel: The variable just exposes the possibility of customization. I agree though that consistency is nice, however, in this particular case it is not something that should be removed.

packages/default/scss/dock-manager/_index.scss
Juveniel: Not needed, as the component does not utilze any module from `common`.

packages/default/scss/dock-manager/_layout.scss
Juveniel: In general, overriding the splitbar here is not the best solution. This needs to be extracted in the Splitter styles and we should not apply any modifications in other components. Taking this into account, for the time being it is better to not expose variables for customizing this override.

inikolova (approved):

JoomFX (approved):

PR #4640, opened 2023-08-01
feat(orgchart): add toolbar with search and export options

inikolova:


Juveniel:
Closing this for the time being. We can re-visit when we start working on this.

PR #4647, opened 2023-08-02
chore: add floating-label templates

zhpenkov:
Add common Floating Label template and simplify tests.


inikolova (approved):

tests/_output/fluent/floating-label/floating-label-outline.png
Juveniel, line 1: tests seem huge - can we somehow reduce/split them in some sensible way ?

inikolova: In my opinion, we could remove the Combobox component from these tests and leave only the Textbox component, as the main idea is to demonstrate the Floating label itself

inikolova (approved):

PR #4651, opened 2023-08-08
chore: add toolbar templates

zhpenkov:
Add common Toolbar templates and simplify tests.

**Note:**
A `toolbar-popup.spec` has been added.

I have simplified and removed the item states in the `toolbar-popup`, `toolbar-popup-size` as well as `toolbar-popup-legacy` tests, to make those tests only showcase the `popup.spec` behavior.


packages/html/src/toolbar/tests/toolbar-sizes.tsx
```diff
@@ -11,7 +11,7 @@ export default () =>(
 
             <span className="k-colspan-all k-col-span-full">Small</span>
             <section className="k-colspan-all k-col-span-full">
-                <Toolbar size="small" resizable>
+                <ToolbarAdaptive size="small">
```
inikolova, line 14: Please simplify this test and show all sizes from the Toolbar options. Like here:
https://github.com/telerik/kendo-themes/blob/develop/packages/html/src/pager/tests/pager-sizes.tsx#L7

packages/html/src/toolbar/templates/toolbar-adaptive.tsx
```diff
@@ -0,0 +1,7 @@
+import ToolBar from "../toolbar.spec";
+
+export const ToolbarAdaptive = ({ children, ...other }) => (
```
inikolova, line 3: In my opinion, this template should be named ToolbarResizable instead as the only option it gets is _resizable_.

packages/html/src/toolbar/templates/toolbar-angular-adaptive.tsx
```diff
@@ -0,0 +1,7 @@
+import ToolbarAngular from "../toolbar-angular.spec";
+
+export const ToolbarAngularAdaptive = ({ children, ...other }) => (
```
inikolova, line 3: Same as above

packages/html/src/toolbar/tests/toolbar-popup.tsx
inikolova, line 1: Adding toolbar-popup.spec is great :) In my opinion, however, the toolbar-popup test should demonstrate only the overflowing controls in the command overflow popup. To show tools with popups, I would suggest to add additional test - toolbar-tools-popup (or sth). What do you think @zhpenkov @kspeyanski?

inikolova (approved):

PR #4654, opened 2023-08-09
fix(fab): keep absolute position for FAB

magdalenaan:
The Floating Button is not positioned correctly in the Fluent theme.

Demo to reproduce: https://dojo.telerik.com/axAWENEW


Juveniel:
there is something wrong with the preview, @magdalenaan could you take a look ?

epetrow (approved):
Overall looks good, I am just wondering if we could do something about the base state as we do not seem to have styles for `k-fab-solid-base` for each theme

JoomFX (approved):

Issue #4655, opened 2023-08-09
[ColumnMenu][ColumnChooser] Update focus-related styling for list items


dtopalov:
**Is your feature request related to a problem? Please describe.**
The `ColumnMenu` and `ColumnChooser` component specs specify the following behavior and attributes:

- the internal input type "checkbox" elements are no longer focusable and interactive
- the wrapping role "option" elements are focusable, navigable and interactive:

https://github.com/telerik/web-components-ux/blob/master/docs/components/column-chooser/aria.json#L41-L93
https://github.com/telerik/web-components-ux/blob/master/docs/components/column-menu/aria.json#L157-L192

Thus, the following styling will need to be added to the respective column list items to cater for the default browser focused outline, and for adding the Kendo-specific focus indicators:

```
.k-column-list-item:focus {
    box-shadow: inset 0 0 0 2px rgba(0, 0, 0, 0.12);
    outline: 0;
}
```

Expected outcome:

![image](https://github.com/telerik/kendo-themes/assets/7533599/9bb1609c-2234-4234-9b83-8c548caa5603)

**NOTE:** As the current a11y spec is not entirely implemented across suites, this may potentially lead to undesired results in components where the focusable item is still the checkbox.

PR #4657, opened 2023-08-10
feat(grid): add focus state styles for the column list item in the column menu

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/4655


tests/_output/fluent/grid/grid-column-menu.png
Juveniel, line 1: @JoomFX the focus in the "Columns grouped expanded" scenario in Fluent does not look as expected. Could you please take a look and see if we can fix this ?

JoomFX: @Juveniel I don't know if we have designs for this exact scenario, but I would expect the focus style to be the same as the one for the List item:

focus

In my opinion the focus style is expected.
Apart from the above, the focus styles for the Column List Item [were already present](https://github.com/telerik/kendo-themes/blob/develop/packages/fluent/scss/grid/_theme.scss#L490) in the Fluent theme.
Let me know if I'm missing something.

Juveniel: image

Juveniel: this is due to the scrollbar that is unfortunately invisible in the visual tests, thus not an actual issue

Juveniel (approved):
change commit to `fix`, otherwise lgtm

PR #4660, opened 2023-08-16
chore(spec): remove spin button default prop

zhpenkov:
To ensure consistency between the component's **spec** behavior and its suites implementation, the **_defaultProps_** of the **DateInput** component no longer includes **_showSpinButton_**.

**_Reasons_:**
- By default, the **DateInput** component in every suite does not come with spinner buttons.

- While spinner buttons are available as an option in suites such as **React**, **Angular**, and **Vue**, they are not implemented in others like **JQuery** and **Blazor** at all. 

**Note:**

- **DateRangePicker**'s **spec** depends on **DateInput**, but the **_reasons_** above are applicable to this component as well. 

- **DateInput** component no longer returns **InputClearValue** in its **spec**, because the returned component is nowhere to be included in the suites' **DateInput** default rendering.


kspeyanski (changes requested):
Please enable the `showSpinButton` in all tests, or at least the `dateinput` ones, so we can have more coverage with the screenshots.

kspeyanski (approved):

PR #4661, opened 2023-08-16
Spreadsheet v2

Juveniel:
targets: https://github.com/telerik/kendo-themes-private/issues/156


JoomFX (approved):
Looks great!
Just a few quick notes/questions:

- Should we remove the `k-widget` class from the rendering in the different visual test files?
- The rendering of the Scrollable Tabs at the bottom worries me a bit. The Prev and Next buttons are after the `k-tabstrip-items` element, while in the Tabstrip rendering it is - Prev button, then the `k-tabstrip-items` element, then the Next button.
- Should we add a RTL test?

tests/_output/default/spreadsheet/spreadsheet.png
inikolova, line 1: Do you think it's a good idea to remove the top border from _k-spreadsheet-sheets-bar_? (the design suggests selected tab with no top border)

tests/_output/fluent/spreadsheet/spreadsheet-tools.png
inikolova, line 1: This png is cropped - I think we should decrease the inline padding of the #test-area for this test

tests/_output/material/spreadsheet/spreadsheet-filtering.png
inikolova, line 1: When running the examples locally, spreadsheet menu and add buttons in the k-spreadsheet-sheets-bar are very misaligned in the Material theme. I know there is no any specific alignment set in this case, however maybe we could add some/
Screenshot 2023-08-29 at 16 23 14

packages/html/src/spreadsheet/tests/spreadsheet-row-contextmenu.tsx
```diff
@@ -0,0 +1,538 @@
...
+                                </li>
+                            </ul>
+                            <Button className="k-tabstrip-prev" fillMode="flat" icon="caret-alt-left" rounded={null} size={null} />
+                            <Button className="k-tabstrip-next" fillMode="flat" icon="caret-alt-right" rounded={null} size={null} />
+                        </div>
+                    </div>
+                </div>
+            </div>
+
+            <Popup id="spreadsheet-context-menu-row" className="k-menu-popup">
```
inikolova, line 525: Shouldn't we add _k-context-menu-popup_ class too?

tests/_output/default/spreadsheet/spreadsheet-print.png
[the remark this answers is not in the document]
Juveniel: This is somewhat expected, as the screenshots have limited size on the CI. However, I have split the test in one column, so that everything is visible.

tests/_output/default/spreadsheet/spreadsheet.png
Juveniel: The observed behavior is due to the rendered scrollbar, which is not visible in the screenshots. Nothing to do here at the moment.

tests/_output/fluent/spreadsheet/spreadsheet-tools.png
Juveniel: Same as the print dialog. I have increased the width of the preview as much as possible.

tests/_output/material/spreadsheet/spreadsheet-filtering.png
Juveniel: This works as expected on the CI and is rather an issue when compiling the changes from the branch locally.

packages/html/src/spreadsheet/tests/spreadsheet-row-contextmenu.tsx
Juveniel: Added

Juveniel:
@JoomFX k-widget is removed. In terms of the sheets bar scrollable buttons - according to the design team, the best solution is to put the scrollable buttons on one side for this scenario. Also, it turns out this is a quite common scenario. Thus, we should rather think for improvement in the TabStrip, so that we are able to position buttons on the sides, or split them on each side etc. 

RTL test - skipping this for the time being as we still need to take care of the legacy rendering. Once we drop this, a test for this will be added.

inikolova (approved):

PR #4662, opened 2023-08-17
Update icons package

inikolova:


epetrow (commented):
I noticed that the lockfileVersion has changed, could you please check?

PR #4665, opened 2023-08-21
chore: add RadioGroup & RadioItem Component spec

zhpenkov:
**RadioGroup** and **RadioItem** specs were added to ensure consistency and unify the **theme** component rendering with the **suite's** implementation.

**Note:**
The tests have been updated according to the new specs.
New test for **RadioGroup** has been added.


kspeyanski (changes requested):
Please add the `name` attribute to the `RadioButtonLabel` templates so single selection is enforced through the browser.

kspeyanski (approved):
I'm really liking how those specs shaped out. Kudos!

packages/html/src/radio/tests/radio-group.tsx
```diff
@@ -0,0 +1,87 @@
...
+    }
+`;
+
+export default () => (
+    <>
+        <style>{styles}</style>
+        <div id="test-area" className="k-d-grid">
+            <span>Normal - vertical</span>
+            <span>With Label - vertical</span>
+            <span>Inside Large - vertical</span>
```
inikolova, line 15: I believe here is a typo -> should be `Inside Label` (instead of `Inside Large`)

packages/html/src/radio/tests/radio-group.tsx
```diff
@@ -0,0 +1,87 @@
...
+                <RadioItem>
+                    <RadioButtonInsideLabel name="inside-label-vertical">Second</RadioButtonInsideLabel>
+                </RadioItem>
+                <RadioItem>
+                    <RadioButtonInsideLabel name="inside-label-vertical">Third</RadioButtonInsideLabel>
+                </RadioItem>
+            </RadioGroup>
+            <span>Normal - horizontal</span>
+            <span>With Label - horizontal</span>
+            <span>Inside Large - horizontal</span>
```
inikolova, line 51: I believe here is a typo -> should be `Inside Label` (instead of `Inside Large`)

packages/html/src/radio/radio-group.spec.tsx
```diff
@@ -0,0 +1,47 @@
...
+    } = props;
+
+    return (
+        <ul
+            {...other}
+            className={classNames(
+                props.className,
+                RADIOGROUP_CLASSNAME,
+                {
+                    [`k-list-${layout}`]: layout === 'horizontal',
```
inikolova, line 33: This will render only _**k-layout-horizontal**_ class (when layout is set to horizontal). I think we should render k-list-vertical class by default, as well as when layout is set to vertical

inikolova (approved):

PR #4667, opened 2023-08-23
chore: add CheckboxGroup & CheckboxItem Component spec

zhpenkov:
**CheckboxGroup** and **CheckboxItem** specs were added to create consistency and unify the theme component rendering with the suite's implementation.

**Note:**
The tests have been updated according to the new specs.
A new test for **CheckboxGroup** has been added.

I have changed the rendered **HTML** tree structure for the **Checkbox** **_Group_** & **_Item_** to be `(ul > li)`, rather than
 `(div > span)`, to achieve similarities between the **theme's** and the **suite's** rendering.


tests/checkbox/checkbox-size.html
```diff
@@ -22,133 +22,133 @@
         <span></span>
         <div>
             <span class="k-checkbox-wrap">
-                <input type="checkbox" class="k-checkbox k-checkbox-sm k-rounded-md k-checked" checked="">
+                <input id="ch1" type="checkbox" class="k-checkbox k-checkbox-sm k-rounded-md k-checked" checked="">
```
inikolova, line 25: id-s should be unique, please update the _**checkbox-size**_ test and remove the duplication

packages/html/src/checkbox/checkbox-group.spec.tsx
```diff
@@ -0,0 +1,47 @@
...
+    } = props;
+
+    return (
+        <ul
+            {...other}
+            className={classNames(
+                props.className,
+                CHECKBOXGROUP_CLASSNAME,
+                {
+                    [`k-list-${layout}`]: layout === 'horizontal',
```
inikolova, line 33: I think we should render **_k-list-vertical_** class by default, as well as when _layout_ is set to _**vertical**_

packages/html/src/checkbox/templates/checkbox-label.tsx
```diff
@@ -2,7 +2,7 @@ import Checkbox from "../checkbox.spec";
 
 export const CheckboxLabel = ({ children, id, ...other }: any) => (
```
kspeyanski, line 3: This is unrelated to this PR, but can we please use the opportunity to rename the file/template to `CheckboxWithLabel`, as the current name applies it's the "Label of the Checkbox", but it's actually a "checkbox **and** a label"

packages/html/src/checkbox/checkbox.spec.tsx
```diff
@@ -35,7 +35,7 @@ const defaultProps = {
 };
 
 export const Checkbox = (
-    props: CheckboxProps & CheckboxState & React.HTMLAttributes<HTMLSpanElement>
+    props: CheckboxProps & CheckboxState & React.HTMLAttributes<HTMLInputElement>
```
kspeyanski, line 38: Nice catch!

kspeyanski (approved):
Spot on!

inikolova (approved):

PR #4668, opened 2023-08-25
chore: update lockfileVersion to 3

inikolova:
Closes https://github.com/telerik/kendo-themes-private/issues/174


Juveniel:
@joneff, @dtopuzov any concerns here?

dtopuzov:
No concerns for lockfileVersion v3 format, the only concern is that we should ensure we use `npm@9` both locally and on CI.

If we have lockfileVersion v3 and `npm@8` when we call `npm ci` it will re-generate the lock file because older node generate lockfileVersion v2 and this re-generate will result in something that will happen if we run `npm i` (lock will not be respected).

dtopuzov (approved):

Juveniel (approved):

PR #4670, opened 2023-08-28
feat(common): introduce indicators helper and move the current time styles there

JoomFX:
related to https://github.com/telerik/kendo-angular-private/issues/665

The Current Time Marker feature is [already implemented in the Kendo jQuery Gantt](https://docs.telerik.com/kendo-ui/api/javascript/ui/gantt/configuration/currenttimemarker).
However, it uses the `k-current-time` class which is supposed to be used in the Scheduler only (styles [here](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/scheduler/_layout.scss#L540) and [here](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/scheduler/_theme.scss#L13)).

This PR introduces a new common helper called "indicators" and moves the current time styles there by making them component agnostic.
Any suggestions about where the current time SCSS variables should be located are welcome. The most obvious possibilities are:

1. In the **_indicators.scss** file together with the styles (they are currently located there). In general, we try not to mix the styles with the variables declaration.
2. In the global **_variables.scss** file of each theme. We try to keep very generic, high level configuration variables there, so it does not feel right to place the current time variables there.
3. To introduce a new **_variables.scss** file in the **common** folder and place the current time variables there.


epetrow (approved):
Looks good, my opinion on the question regarding the SCSS variables is not to expose them and keep the same approach that we have for the other common "helpers".

PR #4673, opened 2023-08-31
chore(html): remove states from pager spec

kikostadinov:
Related to: https://github.com/telerik/web-components-planning/issues/136

- removed `hover` and `focus` states, since they are unnecessary and are causing errors in the rendering tests
- `disabled` and `selected` states still remain, since they are used in the initial state of the Pager


packages/html/src/pager/pager.spec.tsx
```diff
@@ -108,6 +108,7 @@ export const Pager = (
                             'k-pager-numbers'
                         )}>
                         <Button
+                            selected
```
Juveniel, line 111: Any particular reason to leave the `selected` state ? In genera, if we remove the states from the specs, we should not remove them selectively.

kikostadinov: Pager component has one `disabled` and one `selected` button by default in all suites, so if we remove these states, we will have to add exceptions in the test, so it seamed reasonable to leave them. Though, I am okay to remove them as well.

Juveniel (approved):

PR #4683, opened 2023-09-04
chore: add dialog templates

zhpenkov:
Add common Dialog templates and simplify tests.

### DIFF
--- a/packages/html/src/dialog/templates/dialog-modal.tsx
+++ b/packages/html/src/dialog/templates/dialog-modal.tsx
@@ -1,0 +1,13 @@
+import Dialog from "../dialog.spec";
+
+export const DialogModal = ({ children, ...other }: any) => (
+    <>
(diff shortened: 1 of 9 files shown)


packages/html/src/dialog/templates/dialog-modal.tsx
```diff
@@ -0,0 +1,13 @@
+import Dialog from "../dialog.spec";
+
+export const DialogModal = ({ children, ...other }: any) => (
+    <>
+        <div className="k-dialog-wrapper">
+            <div className="k-overlay"></div>
```
Juveniel, line 6: any specific reason why this is not part of the spec ?

zhpenkov: **Comment Update:**
We had a discussion and made the decision to implement it at a later time.

inikolova (approved):

PR #4691, opened 2023-09-11
chore(deps): upgrade lerna to v7

Juveniel:
In this PR:

- upgrade `lerna` to v7
- migrate to [`npm workspaces`](https://docs.npmjs.com/cli/v7/using-npm/workspaces) instead of [deprecated `@lerna/bootstrap`](https://lerna.js.org/docs/legacy-package-management#migrating-from-lerna-bootstrap-lerna-add-and-lerna-link-in-lerna-v7-and-later)
- simplify repo setup( now just `npm install` and `husky init` )
- remove outdated/unused tasks


lerna.json
```diff
@@ -29,5 +22,6 @@
       "verifyAccess": false,
       "yes": true
     }
-  }
+  },
+  "$schema": "node_modules/lerna/schemas/lerna-schema.json"
```
joneff, line 26: Usually, `$schema` is on the top.

package.json
```diff
@@ -44,33 +55,30 @@
...
     "lint-staged": "lint-staged --verbose",
-    "docs": "lerna run docs",
-    "docs:check": "npm run docs && lerna exec --ignore @progress/kendo-themes-html -- git diff --exit-code --quiet -- docs/",
+    "docs": "npm run docs -ws --if-present",
+    "docs:check": "npm run docs -ws --if-present -- git diff --exit-code --quiet -- docs/",
     "render-test-pages": "node ./scripts/render-test-pages.mjs",
     "create-screenshots": "./build/create-screenshots.sh",
     "embed-assets": "gulp assets",
     "create-component": "gulp create-component",
-    "update": "npx npm-check-updates --upgrade",
```
joneff, line 73: You will be surprised, but this bit is useful.

packages/fluent/sass.config.js
```diff
@@ -6,11 +6,11 @@ module.exports = {
         {
             file: 'scss/all.scss',
             outFile: 'dist/all.css',
-            compiler: 'sass-embedded',
+            compiler: 'sass',
```
joneff, line 9: why?

packages/html/package.json
```diff
@@ -50,5 +50,6 @@
   },
   "devDependencies": {
     "@types/react": "^18.0.2"
-  }
+  },
+  "main": "index.js"
```
joneff, line 54: `main` is pretty important field. consider moving it up the tree for visibility. fields at the end of the file, tend to be ignored.

package.json
```diff
@@ -44,33 +55,30 @@
   },
   "scripts": {
     "start": "node ./scripts/start-dev-server.js",
-    "setup": "bash ./scripts/setup.sh",
+    "setup": "npm ci && npx husky install",
```
joneff, line 58: `prepare` could do the job. See docs -- https://docs.npmjs.com/cli/v10/using-npm/scripts#npm-ci

joneff (approved):

lerna.json
Juveniel: moved to top

packages/fluent/sass.config.js
Juveniel: The CI build fails, as the `sass-embedded` os dependencies are options, thus running `npm ci --no-optional` causes an issue. Also, we have actually used `sass` up till now, as the root config has this inconsistency: [see here](https://github.com/telerik/kendo-themes/blob/develop/sass-swatches.config.js#L22-L27).

Thus, I am eliminating the issue, as well as standardizing on the used approach. Also, `sass-embedded` is now [part of the primary `sass` distributable](https://github.com/sass/dart-sass/releases/tag/1.63.0).

package.json
Juveniel: Removing in favor of dependabot ( grouped updates and versioning strategy ).

packages/html/package.json
Juveniel: moved

epetrow (approved):

PR #4692, opened 2023-09-11
Setup integration tests

zhpenkov:
Closes - [issue](https://github.com/telerik/kendo-themes-private/issues/123)

We have implemented integration tests for the most popular building tools, such as **WebPack**, **Gulp-PostCss**, **Parcel** and **Vite**.

### Project structure

```
├───gulp-sass
│   └───styles
├───parcel
│   └───styles
├───vite
│   └───assets
│       ├───bootstrap
│       ├───classic
│       ├───default
│       ├───fluent
│       ├───material
│       └───nouvelle
└───webpack
    ├───src
    └───styles
```
The tests for each build tool are located in `integrations` folder in the main themes directory.
Each build tool package contains the bare minimum code, required to compile each Kendo theme.

### Execution
The workflow for all integration tests will be called in the weekly CI which runs every Monday.

### Local execution

1. navigate to the desired build-tool package (e.g `integrations/webpack`)
2. run `npm i`  and then `npm run build`

You can also build each tool separately from `integrations` - e.g. `npm run  build:gulp` or build all packages together with `npm run build` within the `integrations` folder.

To build all packages from the root of the themes run `npm run test:integrations`

All output **CSS** & files go to `/dist` folder in each webtool package

_NOTE_: Currently there are deprecation warnings emmited on every build which is not ideal. Disabling them with `quietDeps` does not seem to work, as it seems that it may not consider local imports as dependencies ( or at least that's what I think ). What I've tried is listing the path to `node_modules` for every package in `includePaths: [ '../../packages/default/node_modules', '../../packages/bootstrap/node_modules', etc ]`, but unfortunatelly that did not work.

### DIFF
--- a/.github/workflows/ci_weekly.yml
+++ b/.github/workflows/ci_weekly.yml
@@ -83,3 +83,7 @@ jobs:
       - name: Check docs
         run: |
           npm run docs:check
+
+      - name: Test integrations
+        run: |
+          npm run test:integrations
--- a/package.json
+++ b/package.json
@@ -79,7 +79,8 @@
     "create-screenshots": "./build/create-screenshots.sh",
     "embed-assets": "gulp assets",
     "create-component": "gulp create-component",
-    "test-contrast": "node ./scripts/test-contrast.mjs"
+    "test-contrast": "node ./scripts/test-contrast.mjs",
+    "test:integrations": "npm run build --prefix integrations"
   },
   "engines": {
     "node": "^18"
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 node_modules/
 dist/
 modules/
+.parcel-cache/
 
 # Log files
 npm-debug.log
--- a/integrations/gulp-sass/gulpfile.js
+++ b/integrations/gulp-sass/gulpfile.js
@@ -1,0 +1,17 @@
+const gulp = require('gulp');
+
+const sass = require('gulp-sass')(require('sass-embedded'));
+
+const sassOptions = {
+    precision: 10,
+    outputStyle: 'expanded',
+    includePaths: '../../node_modules'
+};
+
+gulp.task('compile', function(done) {
+    return gulp.src('./styles/*.scss')
+        .pipe(sass.sync(sassOptions).on('error', function(err) {
+            done(err);
+        }))
+        .pipe(gulp.dest('./dist'));
+});
--- a/integrations/gulp-sass/package-lock.json
+++ b/integrations/gulp-sass/package-lock.json
@@ -1,0 +1,4464 @@
+{
+  "name": "gulp-sass",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "devDependencies": {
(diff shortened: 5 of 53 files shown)


integration/sass-compilation/webpack/styles/bootstrap.scss
```diff
@@ -0,0 +1,6 @@
+@import "@progress/kendo-theme-bootstrap/scss/all.scss";
+// @import "@progress/kendo-theme-classic/scss/all.scss";
```
inikolova, line 2: Maybe we should remove these comments?

.github/workflows/_themes-integration.yml
```diff
@@ -0,0 +1,72 @@
+name: Integration
```
inikolova, line 1: I would suggest to use _Themes integration_ as a name for this workflow - this way we will be consistent with the other workflows naming, where the actual workflow name matches the .yml file name (but without dashes and underscores)

.github/workflows/ci_weekly.yml
```diff
@@ -83,3 +83,7 @@ jobs:
       - name: Check docs
         run: |
           npm run docs:check
+
+      - name: Test integration
```
Juveniel, line 87: integrations
```

File: .github/workflows/ci_weekly.yml [hunk 4]
```diff
@@ -83,3 +83,7 @@ jobs:
...
+        run: |
+          npm run test:integration
```
Juveniel, line 89: integrations

integrations/gulp-postcss/gulpfile.js
```diff
@@ -0,0 +1,28 @@
+const gulp = require('gulp');
+const postcss = require("gulp-postcss");
+const autoprefixer = require("autoprefixer");
```
Juveniel, line 3: i would avoid using postcss, autoprefixer and postcss-calc dependencies for the purpose of these tests

integrations/gulp-postcss/gulpfile.js
```diff
@@ -0,0 +1,28 @@
...
+const sassOptions = {
+    precision: 10,
+    outputStyle: 'expanded',
+    includePaths: '../../node_modules'
+};
+
+gulp.task('compile', function() {
+    return gulp.src('./styles/*.scss')
+        .pipe(sass.sync(sassOptions).on('error', sass.logError))
+        .pipe(postcss(postcssPlugins))
```
Juveniel, line 26: remove

package.json
```diff
@@ -79,7 +79,8 @@
     "create-screenshots": "./build/create-screenshots.sh",
     "embed-assets": "gulp assets",
     "create-component": "gulp create-component",
-    "test-contrast": "node ./scripts/test-contrast.mjs"
+    "test-contrast": "node ./scripts/test-contrast.mjs",
+    "test:integration": "npm run prebuild --prefix integrations && npm run build --prefix integrations"
```
Juveniel, line 83: integrations

dtopuzov (approved):

PR #4695, opened 2023-09-13
chore: add colorgradient templates

zhpenkov:
Add common **ColorGradient** templates and simplify tests.

**Notes:**
I have removed **hover** and **readonly** states from color-gradient.spec.

**Reasons**
`hover` and `readonly` states did not provide any styles to their respective kendo classes for this component.
The two states are not used by any of the suites, except `readonly` in **Angular**, where this option is implemented by attaching an attribute to the component, not by adding a kendo class - [link](https://www.telerik.com/kendo-angular-ui/components/inputs/colorgradient/readonly-state/).


inikolova:
I would suggest to keep the readonly and hover states - yes, at this point they did not provide any styles to their respective kendo classes, however this will be illustrated in the screenshots.

epetrow (approved):

PR #4697, opened 2023-09-14
chore: add colorpicker templates

zhpenkov:
Add common **ColorPicker** templates and simplify tests.

**Notes:**
I have added `popup` and `opened` props in **colorpicker's** spec as it is a component using **popup** with a specific class
 `k-color-picker-popup` to display its open state content (similar to dropdown components).


inikolova:
I would suggest to rename the colorpicker-gradient/ colorpicker-palette templates and add _popup_ in their names. As an example: colorpicker-popup-gradient / colorpicker-popup-palette

inikolova:
Also, I would suggest to separate the colorpicker.tsx test - in my opinion, it would be better to have one test, which demonstrates different sizes and another - different states.

epetrow (approved):

packages/html/src/colorpicker/tests/colorpicker-states.tsx
```diff
@@ -0,0 +1,53 @@
...
+        flex-direction: column;
+    }
+`;
+
+export default () => (
+    <>
+        <style>{styles}</style>
+        <div id="test-area" className="k-d-grid k-grid-cols-4">
+
+            <section>
```
inikolova, line 25: Maybe we could demonstrate the different options of the <ColorPickerNormal /> in a separate test -> no color, opacity, icon, icon opacity

inikolova:
Please rename the colorpicker-dropdown.tsx to colorpicker-opened.tsx for consistency

tests/_output/bootstrap/colorpicker/colorpicker-opened.png
inikolova, line 1: The popup is clipped. In the tests you should add position: relative to the .k-animation-container

inikolova (approved):

PR #4699, opened 2023-09-15
chore: add editor templates

zhpenkov:
Add common **Editor** templates and simplify tests.

**Notes:**
Should we consider adding more templates to this component?

Added  `disabled` state to **editor** spec.


inikolova (approved):

inikolova:
Suggestion for additional templates - editor with different [rendering modes](https://www.telerik.com/kendo-react-ui/components/editor/rendering-mode/), however we should update the spec, which is not part of this PR. Also, we could add [disabled](https://www.telerik.com/kendo-angular-ui/components/editor/api/EditorComponent/#toc-disabled) state and demonstrate it in a test.

inikolova (approved):

PR #4703, opened 2023-09-19
chore: add listview templates

zhpenkov:
Add common **ListView** templates and simplify tests.

**Notes:**
Added `disabled` state in listview spec.


tests/_output/bootstrap/listview/listview.png
inikolova, line 1: I think we could leave only one _Listview-top pager_ in this test, as they look pretty the same.

packages/html/src/listview/templates/listview-normal.tsx
```diff
@@ -0,0 +1,3 @@
+import { ListView, } from '../../listview';
```
inikolova, line 1: Remove the unnecessary comma

inikolova (approved):

Issue #4717, opened 2023-09-27
[Grid, TreeList] Nested icons broken by k-font-icon overriding k-icon-modifier styling


dtopalov:
When the new `k-font-icon` class is added to the icons indicating Grid grouping and/or column reordering actions, the smaller icon styling (coming from `k-icon-modifier`) is overridden by `k-font-icon`:

![image](https://github.com/telerik/kendo-themes/assets/7533599/a3508053-ab2d-4c92-98e9-b68fb4c89e88)

When new fonts package CSS is loaded before the theme, it looks as expected (but this is hardly an option for customers):

![image](https://github.com/telerik/kendo-themes/assets/7533599/6fe1ed99-8e17-4dec-bb5c-b9971afebb95)

PR #4719, opened 2023-09-28
fix(icon): modifier does not have the correct size

Juveniel:
fixes: https://github.com/telerik/kendo-themes/issues/4717

### DIFF
--- a/packages/default/scss/icons/_layout.scss
+++ b/packages/default/scss/icons/_layout.scss
@@ -38,16 +38,14 @@ $ki-icon-size: $kendo-icon-size;
         margin: .25em;
     }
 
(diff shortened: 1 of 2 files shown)


inikolova (approved):

JoomFX (changes requested):
The issue here is that the `k-font-icon` styles override the `positions: absolute;` and `font-size: 0.5em` of the `k-icon-modifier`.

We should either make the selector stronger (for example `.k-icon.k-icon-modifier`) or use `!important` for `positions: absolute;` and `font-size: 0.5em`.

JoomFX (approved):

Issue #1949, opened 2020-10-06
Add styles for bottom navigation component


joneff:
## Meta

<details>

* Aha item -- [KUI-466](https://progresssoftware.aha.io/features/KUI-466)
* Cross-team issue -- telerik/k2#263
* Design issue -- telerik/kendo-ux-private#197

</details>

**Design**
* [Default theme](https://xd.adobe.com/view/8e1c8b07-28f9-43a6-a96d-21fc158cbc2d-db4f/)
* [Bootstrap theme](https://xd.adobe.com/view/f3778c49-6b14-447c-85cb-d9fa3bb7b603-a624/)
* [Material theme](https://xd.adobe.com/view/70ef723a-5ffb-41e2-89c6-00a8c1507bdf-25bc/)

**Assignee(s)**
* @Juveniel is the main developer

---

## Markup

```html

<nav class="k-bottom-nav k-bottom-nav-primary k-bottom-nav-flat k-bottom-nav-border k-bottom-nav-item-flow-vertical k-pos-fixed">
    <!-- Item element can be one of the following - BUTTON, A, SPAN -->
    <span class="k-bottom-nav-item k-state-selected">
        <span class="k-bottom-nav-item-icon k-icon k-i-email"></span>
        <span class="k-bottom-nav-item-text">Prices</span>
    </span>
    <span class="k-bottom-nav-item k-state-disabled">
        <span class="k-bottom-nav-item-icon k-icon k-i-email"></span>
        <span class="k-bottom-nav-item-text">Heat Map</span>
    </span>
    <span class="k-bottom-nav-item" href="#">
        <span class="k-bottom-nav-item-icon k-icon k-i-email"></span>
        <span class="k-bottom-nav-item-text">Portfolio</span>
    </span>
    <span class="k-bottom-nav-item">
        <span class="k-bottom-nav-item-icon k-icon k-i-email"></span>
        <span class="k-bottom-nav-item-text">Search</span>
    </span>
</nav>
```

<details>
<summary>Framework specific markup</summary>

Markup is unified across frameworks.

</details>

## API

* `BottomNav.fill= 'solid` -> `<nav class="k-bottom-nav-solid" />` | `('solid' or 'flat')` | `Default: 'flat'`
* `BottomNav.themeColor = 'secondary'` -> `<nav class="k-bottom-nav-secondary" />` | `Default: 'primary'`  |
[...]

PR #4721, opened 2023-09-28
chore: add bottom-navigation templates

inikolova:
Adds common templates for the BottomNavigation component and simplify tests

**Note**: I changed the default value of the NavBar **fillMode** to **_flat_**, following the recommendations here:
https://github.com/telerik/kendo-themes/issues/1949. Also, the default value in all Kendo UI suites is _flat_ too.


Juveniel:
test contrast is failing - need to [update the test name here ](https://github.com/telerik/kendo-themes/blob/develop/scripts/test-contrast.mjs#L39)in order to be excluded from testis ( it is an exception ).

zhpenkov (approved):

zhpenkov (approved):

PR #4760, opened 2023-10-24
chore: add upload templates

zhpenkov:
Adds common templates for the Upload component & update tests.
I have moved the states from the current test to a separate one.

**Note:**
Should i add more templates for `uploading`, `done` and `failed` Upload status ?


inikolova (changes requested):
In my opinion we should add a template for UploadFileMultiple too

packages/html/src/upload/tests/upload-states.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export default () => (
+    <>
+        <style>{styles}</style>
+        <div id="test-area" className="k-d-grid k-grid-cols-1">
+
+            {[ 'normal', ...Upload.states ].map((state) => (
+                <section>
+                    <span>{state}</span>
+                    <Upload async actions status="upload" {...{ [state]: true }}>
```
inikolova, line 24: Use UploadNormal here

packages/html/src/upload/tests/upload-states.tsx
```diff
@@ -0,0 +1,31 @@
...
+                        <UploadFile name="Image1.jpg" icon="file-image" state="uploaded" size="106.43 KB"></UploadFile>
```
inikolova, line 25: The `state="uploaded"` attribute is not valid and I am not sure we need it

packages/html/src/upload/tests/upload-states.tsx
```diff
@@ -0,0 +1,29 @@
+import { Upload, UploadChunk } from '../../upload';
```
inikolova, line 1: In my opinion, here we could use UploadNormal, as disabled Upload does not allow to upload files

packages/html/src/upload/tests/upload.tsx
```diff
@@ -1,71 +1,64 @@
-import { Upload, UploadFile, UploadFileMultiple, UploadFileInfo } from '../../upload';
+import { UploadFile, UploadFileMultiple, UploadFileInfo, UploadChunk, UploadBatch } from '..';
 
 
-export default () =>(
+export default () => (
     <>
         <div id="test-area" className="k-d-grid k-grid-cols-2">
```
inikolova, line 7: Please add titles to the columns -> Chunk Upload and Batch Upload

inikolova (approved):

PR #4761, opened 2023-10-24
chore: add window templates

zhpenkov:
Adds common templates for the Window component & update tests.


inikolova:
In my opinion we don't need **window-with-action-buttons** template; we should demonstrate this option in a test (similar to the [Dialog component](https://github.com/telerik/kendo-themes/blob/develop/packages/html/src/dialog/tests/dialog.tsx)).

tests/_output/bootstrap/window/window-rtl.png
inikolova, line 1: Title is cropped - please check

packages/html/src/window/tests/window.tsx
```diff
@@ -1,8 +1,9 @@
-import { Window } from '../../window';
+import { WindowNormal } from '..';
 
 const style = `
     .k-window {
         position: relative;
+        width: 300px;
```
inikolova, line 6: Maybe we could increase this width?

zhpenkov: I decided to remove the width limitation completely as it was and leave it to scale according to the content it has. **WDYT**.

packages/html/src/editor/tests/editor-table-wizard.tsx
```diff
@@ -22,7 +22,7 @@ export default () =>(
         <style>{styles}</style>
         <div id="test-area" className="k-d-grid k-grid-cols-2">
             <section>
-                <Window className="k-editor-window k-editor-table-wizard-window" title="Table properties" actions={[ 'minimize', 'maximize', 'close' ]} actionButtonsAlign="start" actionButtons={
```
inikolova, line 25: Editor table wizard window buttons differ from the WindowNormal default buttons

inikolova (approved):

PR #4762, opened 2023-10-24
chore: add loader templates

zhpenkov:
Adds common templates for the Loader component & update tests.

**Note:**
Added the different animations of `loader` and `loader-container` as templates (different animations - different **HTML** renders to test). The component might as well have only `loader-normal` & `loader-container-normal` templates - **WDYT**?

Is the different **HTML** rendering of a component's behavior a pattern we should consider for creating a template?


zhpenkov:
> In my opinion the component templates should be loader-normal & loader-container-normal

We further discussed and decided to add templates for _**loader-infinite**_ and _**loader-converging**_, in addition to those mentioned earlier.

Juveniel (approved):

Issue #4704, opened 2023-09-19
ListBox with top or bottom toolbox overlaps other elements


gdenchevprog:
**Describe the bug**

When the toolbar action buttons are rendered at the top or bottom of the ListBox, the component overlaps other elements on the page.

**To reproduce**

1. Run the following Dojo - https://dojo.telerik.com/@gdenchev/iWAPuRiZ
2. By default the `k-listbox-actions-top` class is used. You can switch it to `k-listbox-actions-bottom` to observe the opposite behavior(part of the listbox is cut off).

The HTML is taken from one of the listbox tests.

**Expected behavior**

The ListBox should not overlap with other elements on the page(or be cut off in case it is the first element).

**Screenshots**

When the toolbar is rendered at the top:
![image](https://github.com/telerik/kendo-themes/assets/70957719/eafb115f-cf5f-448d-aa52-020b79b12cb6)

When the toolbar is rendered at the bottom:
![image](https://github.com/telerik/kendo-themes/assets/70957719/012570c4-daaa-40d0-81d7-5da05a297b6a)

**Affected package (please remove the unneeded items)**
* theme-default (checked with the ocean blue theme)

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery
* Kendo UI for Angular

The problem is present in the above two suites, however I haven't checked any of the other ones.

**Affected browsers (please remove the unneeded items)**
* All

**Build system information (please remove the unneeded items)**
* Not Applicable

**Additional context**

The element with class `k-listbox` (the container/wrapper) has a height of 200px:
![image](https://github.com/telerik/kendo-themes/assets/70957719/5716b13e-1c3f-4c38-bc4a-183f0ff3e119)

This element is meant to wrap around both the list and the action buttons. The list (`k-list-scroller`) has a height that is set to `inherit`. In this case the inherited width would be the one of the parent(200px). However, this does not seem to take the bottom/top toolbar into consideration. The toolbar also occupies some of the space and the total height exceeds the one of the `k-listbox`(200px).

PR #4766, opened 2023-10-26
Fix listbox overlapping

epetrow:
fixes: https://github.com/telerik/kendo-themes/issues/4704

### DIFF
--- a/packages/default/scss/listbox/_layout.scss
+++ b/packages/default/scss/listbox/_layout.scss
@@ -48,10 +48,13 @@
 
         .k-list-scroller {
             width: 100%;
+            min-height: 0;
             height: inherit;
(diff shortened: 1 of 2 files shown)


inikolova (approved):

inikolova:
In my opinion we should remove the following CSS rule from the test:
`.k-listbox .k-list-scroller {
   height: auto;
}`
and demonstrate how ListBox with default height acts.

Juveniel (changes requested):
see previous comment

inikolova:
@Juveniel I agree that the best option is to adjust the scroller height internally.

inikolova (approved):

JoomFX (approved):

PR #4767, opened 2023-10-27
chore: add notification templates

zhpenkov:
Adds common templates for the Notification component & update tests.


packages/html/src/notification/tests/notification-angular.tsx
```diff
@@ -15,133 +15,92 @@ const styles = `
...
-            </div>
-            <div>
-                <div className="k-notification-container">
-                    <Notification closable={true} text="Closable notification" actions={ <NotificationAction type="close" /> } />
-                </div>
-            </div>
             <div>
                 <div className="k-notification-container">
-                    <Notification themeColor="primary" text="Primary notification" />
+                    <IconNotification themeColor="info" text="Info notification" icon="info-circle" />
```
inikolova, line 26: In muy opinion this test should demonstrate only the types of notifications, but not some of the colors (as there is another test which demonstrates all colors)

inikolova (approved):

PR #4768, opened 2023-10-30
chore: add fab templates

zhpenkov:
Adds common templates for the FloatingActionButton component & update tests.


packages/html/src/fab/tests/fab-items.tsx
inikolova, line 1: I would suggest to demonstrate FAB sizes from the spec (like here -> https://github.com/telerik/kendo-themes/blob/develop/packages/html/src/button/tests/button-size.tsx)

inikolova (approved):

PR #4771, opened 2023-10-31
chore: add multiselecttree templates

zhpenkov:
Adds common templates for the MultiSelectTree component & update tests.

**Note:**
The second commit suggests removing the static adaptive mode of **MultiSelectTree** and other components ([MultiSelect](https://github.com/telerik/kendo-themes/blob/develop/packages/html/src/multiselect/multiselect.spec.tsx#L150C26-L150C26), [DropDownTree](https://github.com/telerik/kendo-themes/blob/develop/packages/html/src/dropdowntree/dropdowntree.spec.tsx#L151)) from the component's helper (spec). Those specifically set adaptive renders of the component are not accessible outside the spec and cannot be altered for the needs of a test (for example).

In my opinion, hardcoded component structures in the helpers should be migrated to **templates**, with an option for outside access and customization. **WDYT**?

**Update:**
Decided to not make any changes for now.


inikolova:
In my opinion [adaptiveHeaderSettings](https://github.com/telerik/kendo-themes/blob/multiselect-tree-templates/packages/html/src/multiselecttree/multiselecttree.spec.tsx#L58) should not be part of the MultiSelectTree props, as they are not related to the component itself. Therefore I would recommend to remove this prop and hardcode the action sheet header structure.

packages/html/src/multiselecttree/tests/multiselecttree-size.tsx
```diff
@@ -13,38 +13,20 @@ export default () =>(
...
-                    <>
-                        <Chip text="Chip Text" size="large" actions={ <ChipAction type="remove"/> } />
-                        <Chip text="Chip Text" size="large" actions={ <ChipAction type="remove"/> } />
-                    </>
-                )}
-            />
+            {MultiSelectTree.options.size.map((size) => (
+                <>
+                    <span>{size}</span>
+                    <MultiSelectTreeNormal
```
inikolova, line 19: You should handle the rounded options here as well. Check this test for reference: https://github.com/telerik/kendo-themes/blob/develop/packages/html/src/multiselect/tests/multiselect-size-rounded.tsx

packages/html/src/multiselecttree/multiselecttree.spec.tsx
```diff
@@ -55,7 +55,7 @@ export type KendoMultiSelectTreeProps = KendoMultiSelectTreeOptions & {
     dir?: 'ltr' | 'rtl';
     adaptive?: boolean;
     adaptiveSettings?: KendoActionSheetProps;
-    adaptiveHeaderSettings?: KendoActionSheetHeaderProps;
+    adaptiveTitle?: string;
```
inikolova, line 58: In my opinion adaptiveTitle is not directly related to the MulstSelectTree component, it relates to the ActionSheet itself and I would suggest to skip this prop

inikolova (approved):

PR #4785, opened 2023-11-10
chore: add dropzone templates

zhpenkov:
Adds common templates for the Dropzone component & update tests.

**Notes:**
I have made the **hint** in the component's spec configurable. (it is an option in kendo react - [ **component API** ](https://www.telerik.com/kendo-react-ui/components/upload/api/ExternalDropZoneProps/#toc-customhint))


packages/html/src/dropzone/dropzone.spec.tsx
```diff
@@ -9,14 +9,16 @@ const states = [
...
     note?: string | boolean;
 };
 
 export type KendoDropzoneState = { [K in (typeof states)[number]]?: boolean };
 
 const defaultProps = {
     icon: "upload",
-    note: "Only JPEG, PNG and SVG files are allowed."
+    hint: "Drag and drop files here to upload",
+    note: "Only JPEG, PNG and SVG files are allowed.",
```
inikolova, line 21: Could we make the `note` optional? This way in the normal template we will skip the prop with false value?

Juveniel (approved):

PR #4789, opened 2023-11-13
Unify card rendering

inikolova:
Related to https://github.com/telerik/kendo-themes-private/issues/194


tests/_output/bootstrap/card/card-action-buttons.png
epetrow, line 1: Could you please check the action buttons alignment here and in Material as they seem misaligned?

packages/html/src/card/tests/card-colors.tsx
zhpenkov, line 1: I suggest displaying the **Card** colors by iterating over the **themeColor** array in the component's spec.

packages/html/src/card/tests/card-deck.tsx
zhpenkov, line 1: For consistency, maintainability and more test coverage, should we add a separate **card-states** test, that iterates through every state from the **`states`** array in the component spec?

inikolova: At this point, cards don't have styles for different states implemented (also there isn't a design for this), so I would suggest to refrain from adding a similar test (at least for now).

tests/_output/bootstrap/card/card-action-buttons.png
inikolova: Buttons are correctly aligned, however there is not enough space to show them entirely.

epetrow (commented):
I would suggest reinstating the `k-card-vertical` class both in styles and rendering, otherwise I think it looks good.

packages/html/src/card/templates/card-normal.tsx
```diff
@@ -0,0 +1,7 @@
+import { Card } from '../../card';
+
+export const CardNormal = (props) => (
+    <Card orientation="vertical"
```
zhpenkov, line 4: We should add a `children` prop here with a default template example. 

Same goes for **templates/card-horizontal.tsx**.

epetrow (approved):

zhpenkov (approved):

JoomFX (approved):

PR #4791, opened 2023-11-14
chore: add orgchart templates

zhpenkov:
Adds common templates for the Orgchart component & update tests.


tests/_output/bootstrap/orgchart/orgchart-group.png
inikolova, line 1: We cannot have minus icon at the end - button with minus icon is always followed by an orgchart card

inikolova (approved):

PR #4796, opened 2023-11-22
chore: fix lint errors

inikolova:
Related to https://github.com/telerik/kendo-themes-private/issues/192


packages/html/src/slider/tests/slider.tsx
zhpenkov, line 1: We might iterate over the slider's `states` array from the component's spec if possible for this test.

inikolova:
> It looks like we are still supporting two different types of rendering. We should discuss and decide which way to go, and remove the uncecessary styles from the themes.

We discusses this particular case and it appears both renderings are needed - sliders in React and Vue support option for different labels/ticks positions - https://www.telerik.com/kendo-vue-ui/components/inputs/slider/ticks/.

packages/html/src/slider/templates/slider-vertical.tsx
```diff
@@ -0,0 +1,20 @@
+import { Slider, SliderTick } from "../../slider";
+
+export const SliderVertical = (props) => (
+    <Slider orientation="vertical" children={[
```
zhpenkov, line 4: We do not need to pass the **children** as an array here - `children={ <> JSX.Elements </> }` should be enough.
It applies to the other templates as well.
```suggestion
    <Slider orientation="vertical" children={
```

PR #4799, opened 2023-11-24
chore(dock-manager): remove no longer needed chrome specific styles

epetrow:
Seems like the writing-mode issue with the button in Chrome has been resolved, so I am removing the workaround.

### DIFF
--- a/packages/default/scss/dock-manager/_layout.scss
+++ b/packages/default/scss/dock-manager/_layout.scss
@@ -30,18 +30,6 @@
             height: 0;
             border-width: $kendo-dock-manager-border-width 0 0;
         }
-
-        // Needed because of a bug in Chrome - Issue 1473554
-        // A fix will be shipped likely in milestone 118 and then we can revisit this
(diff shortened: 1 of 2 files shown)


Juveniel (commented):
@epetrow can we remove [this](https://github.com/telerik/kendo-themes/pull/4799/files#diff-98717c3b7cd0c434ff88c312ca437a35b6d804d70c4350d62d87152e8705b069L24-L26) as well if it is related to the same fix ?

epetrow:
> @epetrow can we remove [this](https://github.com/telerik/kendo-themes/pull/4799/files#diff-98717c3b7cd0c434ff88c312ca437a35b6d804d70c4350d62d87152e8705b069L24-L26) as well if it is related to the same fix ?

This is unfortunatelly unrelated to the fix, currently it's needed because of  [these](https://github.com/telerik/kendo-themes/blob/00fbdd430c74d792dadbb35dc7f15107c8973d09/packages/default/scss/toolbar/_layout.scss#L27) styles in the `Toolbar`.

JoomFX (approved):

PR #4817, opened 2023-12-12
fix(common): incorrectly set current time marker width

JoomFX:
The current time marker should not have a `width` property set because it is supposed to be used as both a horizontal and a vertical line.
The components (Scheduler and Gantt) should add the required property inline (either `width` or `height`), as they currently do.


epetrow (commented):
We should add the aforementioned `width` and `height` properties in the tests, since the time marker is not visible now.

epetrow (approved):

PR #4820, opened 2023-12-15
fix(html): enable callout property

inikolova:
Cards in Timeline always have callouts, therefore we should enable this prop by default. When callout property is true, `k-card-with-callout` class is set to the card - https://github.com/telerik/kendo-themes/blob/develop/packages/html/src/card/card.spec.tsx#L69. Also, this recommendation was described in the [_Unify rendering for Timeline_ issue](https://github.com/telerik/kendo-themes-private/issues/193)

### DIFF
--- a/packages/html/src/timeline/timeline-card.spec.tsx
+++ b/packages/html/src/timeline/timeline-card.spec.tsx
@@ -23,6 +23,7 @@ export const TimelineCard = (
                 props.className
             )}
             orientation="vertical"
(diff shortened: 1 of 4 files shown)


Juveniel (commented):
Are visual changes expected here ?

inikolova:
@Juveniel no -> we need this prop enabled only because of the class rendered

Juveniel (approved):

PR #4822, opened 2023-12-18
Revert multiselecttree rendering

inikolova:
Related to https://github.com/telerik/kendo-themes/pull/4631


JoomFX (approved):

JoomFX (approved):

epetrow (commented):
I was wondering could we utilize flex instead of width and a pseudo-element e.g.

```scss
.k-multiselecttree .k-input-values:empty,
.k-multiselecttree .k-input-inner:empty {
    flex: 0;
}
```

In my understanding the problem is related to the width of these two containers, so preventing them from taking up space when they are empty should work in theory.

inikolova:
@epetrow we could use flex for the part with the **.k-input-values** width and your suggestion works as expected. However, there is another problem with the height of the Multiselecttree with small and large sizes - the height is different with and without values selected (this is an old issue and could be observed in online demos -> https://www.telerik.com/kendo-angular-ui/components/dropdowns/multiselecttree/appearance/). Hence, the rule for **.k-input-inner::before** fixes the aforementioned problem and unfortunately we could not omit it with the current component rendering

epetrow:
> @epetrow we could use flex for the part with the **.k-input-values** width and your suggestion works as expected. However, there is another problem with the height of the Multiselecttree with small and large sizes - the height is different with and without values selected (this is an old issue and could be observed in online demos -> https://www.telerik.com/kendo-angular-ui/components/dropdowns/multiselecttree/appearance/). Hence, the rule for **.k-input-inner::before** fixes the aforementioned problem and unfortunately we need it with the current component rendering

I would suggest going with the flex approach (if you also agree) and since these two seem to be separate issues maybe split them in two commits :grey_question:

inikolova:
Ok, I will split the rules in separate commits. However I would like to further test the flex approach

inikolova:
I made some additional tests and have some concerns about the :empty pseudo selector in this scenario. I would prefer to proceed with the width/max-width approach. @JoomFX what do you think?

JoomFX:
In any case, we still need the `:before` adjustment:

```
.k-multiselecttree .k-input-inner::before {
  content: "\200b";
  width: 0px;
  overflow: hidden;
  flex: none;
  display: inline-block;
  vertical-align: top;
}
```

I tested the `:empty` approach in Angular and it works.
Both approaches work.

Since a single white space will make the `:empty` approach to not work I would go the safer way with the width/min-width.
But those are cosmetic things - both approaches will do the job.

JoomFX (approved):

epetrow (approved):
Since the majority opinion is to go with the current implementation of the placeholder text fix as well as that the `:empty` approach is prone to issues mentioned above, I'll give this a thumbs up.

PR #4852, opened 2024-01-18
Add Treelist html spec

epetrow:
closes: https://github.com/telerik/kendo-themes-private/issues/190


tests/treelist/treelist-aggregates.html
```diff
@@ -257,9 +259,9 @@
                         </div>
                     </div>
                     <div class="k-grid-container">
-                        <div class="k-grid-content k-virtual-content">
```
inikolova, line 260: I think the `k-virtual-content` class is needed for virtualization and I am not sure we should remove it - could you please check?

packages/html/src/treelist/tests/treelist-aggregates.tsx
```diff
@@ -1,271 +1,249 @@
...
+                                        <GridHeaderCell rowspan={1} colspan={1} columnTitle="Name"></GridHeaderCell>
+                                        <GridHeaderCell rowspan={1} colspan={1} columnTitle="Title"></GridHeaderCell>
+                                        <GridHeaderCell rowspan={1} colspan={1} columnTitle="hireDate"></GridHeaderCell>
+                                    </TableRow>
+                                </TableThead>
+                            </GridHeaderTable>
+                        </div>
+                    </GridHeader>
+                    <GridContainer>
+                        <GridContent>
```
inikolova, line 159: I think we should keep the `k-virtual-content` class here

inikolova (approved):

PR #4899, opened 2024-02-21
chore: add element with k-sr-only class

inikolova:
According to the aria spec, we should render an element with class **k-sr-only** as a direct child of .**k-scrollview**: 

https://github.com/telerik/web-components-ux/blob/master/docs/components/carousel/aria.json#L45

This has been already implemented in jQuery and Angular.

@slavenai @dtopalov @PekoPPT


epetrow (commented):
I think this should be a fix

inikolova:
Closed in favor of https://github.com/telerik/kendo-themes/pull/4905

PR #4911, opened 2024-02-29
chore(html): spec / template corrections for rendering test adoption

zhpenkov:
This PR makes minor changes and corrections to the component's specs and templates used for reference rendering tests.

The changes to the HTML component modules are the following:

**Editor**
- split **editor-normal** template into **editor-responsive-toolbar** template and **editor-normal** template
- add **editor-resizable** template
- make Redo/Undo buttons disabled by default

**ListBox**
- add data and actions to the **listbox-normal** template

**ColorPicker**
- correct typo in **ColorPicker** popup class - ( was **k-color-picker-popup**, should be **k-colorpicker-popup**)
- change the **fillMode** of the 'select view' **ColorEditor** button group to **flat** (**k-button-group-flat**)
- make template scenarios achievable, and add additional parameters for customizing templates

**Slider**
- remove '**k-**' prefix from **k-range-slider**, **k-slider-0**, **k-slider-50**, **k-slider-100** as they are used only for applying styles in the tests and can be mistaken for actual kendo classes
- remove **RangeSlider** buttons
- add **k-first** and **k-last** classes to slider's tick


Juveniel (approved):

Juveniel:
Note: we should think of a way to addd `k-first`/`k-last`/etc classes through the spec.

Issue #4909, opened 2024-02-28
[MultiColumnComboBox] Group header is overlapped in Safari


stnikolova:
**Describe the bug**
Group headers are overlapping at the first group header instance and the rest of the group header rows are empty.

**To reproduce**
Reproduced with the MultiColumnComboBox online demo [here](https://www.telerik.com/kendo-angular-ui/components/dropdowns/multicolumncombobox/grouping/).

1. Open 'Safari' browser
1. Go to [MultiColumnComboBox](https://www.telerik.com/kendo-angular-ui/components/dropdowns/multicolumncombobox/grouping/)
1. Open the popup
1. See that the group headers overlap

**Expected behavior**
Group headers should not overlap.

**Screenshots**
![image](https://github.com/telerik/kendo-themes/assets/18744873/75e2b417-b9c4-4d07-b91c-528d800bed72)

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material
* theme-tasks

**Affected suites (please remove the unneeded items)**
* Kendo UI for Angular
* Kendo UI for React
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**
* Safari version: 17.3

**Build system information (please remove the unneeded items)**
* Not Applicable
* OS: Mac

zdravkov:
Just wanted to add that It is actually is a regression in KendoReact.

PR #4914, opened 2024-03-01
fix(table): table-list group row is misaligned in safari

Juveniel:
fixes: https://github.com/telerik/kendo-themes/issues/4909

### DIFF
--- a/packages/default/scss/table/_layout.scss
+++ b/packages/default/scss/table/_layout.scss
@@ -113,12 +113,15 @@
         list-style: none;
         outline: none;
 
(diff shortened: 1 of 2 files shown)


Juveniel:
@stnikolova, @zdravkov as there seem to be some dynamic styles applied to elements in angular/react, could you please verify if this fix successfully resolves the issue with the current components implementation?

zdravkov:
Looks ok for React

stnikolova:
Looks good in Angular as well 🚀

PR #4918, opened 2024-03-06
feat(utils): generate utility classes with css variables

JoomFX:
related to https://github.com/telerik/kendo-themes-private/issues/220

With this PR, the color utility classes (background-color, border-color, text-color and outine-color) will be generated with CSS variables as follows:

```css
.k-bg-primary {
    background-color: var(--kendo-color-primary, #ff6358);
}

.k-bg-inherit {
    background-color: var(--kendo-color-inherit, inherit);
}
```

We do not generate utility classes for all `$kendo-colors`. We do it only for the main colors in order to keep the bundle with a reasonable size.
[These are the colors](https://github.com/telerik/kendo-themes/blob/color-utility-classes/packages/utils/scss/_variables.scss#L11) we generated utility classes for.

Before this change (with `$kendo-enable-color-system` set to `false`), the Default theme CSS file is **895 KB**.
With the `$kendo-enable-color-system` set to `true`, the Default theme CSS file is **972 KB**.

**NOTE:** If `$kendo-enable-color-system` is set to `true`, some screenshots will be generated. This is expected.
The reason for these changes is that when the new colors are being used, there is a slight difference in the shade of some colors (success, error, warning, info, etc).
~~The biggest and most obvious difference is that with the old/current colors the `k-text-base` class (that we use in the Loader Container visual tests only) does not have any styling (it is an empty class), while with the new colors it takes the `base` color which is a light gray color. This shouldn't be a problem because, as far as I could see, none of the suites have "Base" as a `themeColor` in the Loader, meaning that we only have it in the visual tests.~~ This is no longer an issue with the decision to not generate utility classes for all `$kendo-colors` colors.

### DIFF
--- a/packages/fluent/scss/utils/_layout.scss
+++ b/packages/fluent/scss/utils/_layout.scss
@@ -4,7 +4,9 @@
     $kendo-spacing: $kendo-spacing,
     $kendo-font-sizes: $kendo-font-sizes,
     $kendo-border-radii: $kendo-border-radii,
-    $kendo-elevation: $kendo-elevation
+    $kendo-elevation: $kendo-elevation,
+    $kendo-colors: $kendo-colors,
+    $kendo-enable-color-system: $kendo-enable-color-system
 );
 
 @mixin kendo-utils--layout() {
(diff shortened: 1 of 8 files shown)


epetrow (approved):
Looks good to me

epetrow (approved):

Juveniel (approved):

PR #4921, opened 2024-03-11
Unify rendering PanelBar

zhpenkov:
closes: https://github.com/telerik/kendo-themes-private/issues/215


JoomFX (approved):
If the item icon from the visual tests is removed on purpose, then everything looks great! :)

NOTE: When the Spacing module is merged, some of the variable values will need to be updated.

zhpenkov:
> If the item icon from the visual tests is removed on purpose, then everything looks great! :)
> 
> NOTE: When the Spacing module is merged, some of the variable values will need to be updated.

The item icon was actually missing in the rtl test, so I added it.

Juveniel:
@zhpenkov please mark the changed styles commit accordingly e.g `feat(panelbar): update styles to match universal rendering`

packages/html/src/panelbar/panelbar-item.tsx
```diff
@@ -0,0 +1,101 @@
...
+                    ["k-panelbar-header"]: header,
+                    [`k-level-${level}`]: level || level === 0,
+                    ["k-expanded"]: expanded
+                }
+            )}
+        >
+            <span
+                className={classNames(
+                    "k-link",
+                    stateClassNames("k-link", {
```
inikolova, line 68: "k-link" is duplicated?

zhpenkov: The `stateClassNames` function accepts 2 parameters, `_componentClassName`, and `props`, that's why `k-link` is passed inside and used for a second time. There is no extra `k-link` class added to this element.

inikolova (approved):

PR #4934, opened 2024-03-20
feat(classic): enable color system

JoomFX:
BREAKING CHANGE: This commit enables the new Color System,
which introduces a new way of implementing and customizing the colors.
If the user has customized some of the old color variables like
`$kendo-color-primary`, `$kendo-body-bg`, `$kendo-base-bg`, etc, they
would need to migrate their customizations by following the Migration Guide
which can be found in the Progress Design System Docs.


zhpenkov (approved):
Looking good to me.

inikolova (approved):

PR #4954, opened 2024-03-28
chore: provide an option to set custom className

inikolova:
We should be able to set custom classNames to .k-drawer


zhpenkov (commented):
We can avoid adding another component by just adding `props.className` to the current drawer.spec.tsx here https://github.com/telerik/kendo-themes/blob/develop/packages/html/src/drawer/drawer.spec.tsx#L53
This will make the overall component configuration / templates simpler and we will also be able to apply classes directly to k-drawer element.

zhpenkov (approved):

epetrow (approved):

PR #4956, opened 2024-03-29
chore: generate missing utils

inikolova:
In this PR:

- Generate missing spacing-related utils;
- Remove examples with negative paddings, as values for padding cannot be negative;


epetrow (approved):
lgtm

JoomFX (approved):

PR #4958, opened 2024-03-29
chore: remove warnings from html package

zhpenkov:
Correct the specs/templates to eliminate the following warnings.

- Warning: Use the `defaultValue` or `value` props instead of setting children on <textarea>.
- Warning: Each child in a list should have a unique "key" prop.


JoomFX (approved):
LGTM

PR #4969, opened 2024-04-03
chore(utils): use correct css var in font weight utils

epetrow:
`k-font-*` utils were not using the correct css variables `--kendo-font-weight-*`, this PR addresses that


Juveniel:
we have not release this change yet, thus this should be a `chore`

inikolova (approved):

PR #4977, opened 2024-04-11
Update range selection styles

epetrow:
related to: https://github.com/telerik/web-components-ux/issues/647#issuecomment-1725925967

design: https://www.figma.com/file/zBhYXheWNnVirZjMuNarEg/DateRange-Picker?type=design&node-id=7%3A3&mode=dev


tests/_output/material/calendar/calendar-infinite-rtl.png
JoomFX, line 1: There seems to be an issue with the RTL styles, which is also visible in the Bootstrap screenshot, but the other themes should be checked as well.

JoomFX (changes requested):

tests/_output/material/calendar/calendar-infinite-rtl.png
epetrow: I noticed it too and it is addressed now

JoomFX (approved):

PR #4978, opened 2024-04-12
Drawer use `$kendo-drawer-content-padding-x` and `$kendo-drawer-content-padding-y`

WissamProgress:
This is based on feedback from `Ticket ID: 1648341` where they want to set the padding using SCSS variables.

Currently, it is only possible to do this using CSS:
```
.k-content-drawer {
  padding-inline: 20px;
  padding-block: 20px;
}
```

https://stackblitz.com/edit/react-siozrp-d1bfnw?file=app%2Fmain.jsx,app%2Fstyles.scss

### DIFF
--- a/packages/bootstrap/docs/customization-drawer.md
+++ b/packages/bootstrap/docs/customization-drawer.md
@@ -98,10 +98,20 @@ The following table lists the available variables for customization.
     </td>
 </tr>
 <tr>
+    <td>$kendo-drawer-content-padding-x</td>
+    <td>Null</td>
+    <td><code>null</code></td>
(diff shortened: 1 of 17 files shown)


epetrow:
Up until now we were not adding default padding to the drawer content ( since the variables were not used ), however this PR changes this behaviour and this raises a question:

Should we expose the variables and set content padding by default or keep the current behaviour and expose them with a `null` value?

I synced with @imasyuko and the suggestion is to handle this case like we did in the Tabstrip and use the same content padding values.

Juveniel (approved):

PR #4989, opened 2024-04-22
chore: do not use inherit fallback for calculated font-size

epetrow:


Juveniel (changes requested):
1) there is one additional use-case in the scheduler [here](https://github.com/telerik/kendo-themes/blob/calculated-font-size/packages/material/scss/scheduler/_layout.scss#L73).
2) The default value for `--kendo-font-size` is `.875rem`, perhaps we should use this default in calculations as well. Please check this.

inikolova (approved):

Juveniel (approved):

PR #5004, opened 2024-04-26
chore: document missed variables

inikolova:


inikolova:
In offline discussion we agreed to not document these variables for now, therefore I am closing this PR

PR #5005, opened 2024-04-26
chore(html): add map spec

zhpenkov:
Add specs and templates for the Map component.


packages/html/src/map/zoom-control.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export const ZoomControl = (
+    props: React.HTMLAttributes<HTMLDivElement>
+) => {
+    const {
+        ...other
+    } = props;
+
+    return (
+        <div
```
Juveniel, line 20: reuse the ButtonGroup component here if possible

PR #5007, opened 2024-04-26
chore(material): adjust solid button styles

JoomFX:
related to https://github.com/telerik/kendo-themes-private/issues/222


packages/material/scss/button/_variables.scss
```diff
@@ -174,19 +174,19 @@ $kendo-button-selected-shadow: null !default;
 
 /// The base background of the focused Button.
 /// @group button
-$kendo-button-focus-bg: null !default;
+$kendo-button-focus-bg: if($kendo-enable-color-system, k-color( base-hover ), null) !default;
```
inikolova, line 177: Is 'base-hover' the proper color here?

inikolova:
There is no visual distinguish for different states when button is used inside Toolbar -> https://github.com/telerik/kendo-themes/pull/5007/commits/d20b754d2331c178bd224eb97dd18741099bc46f

packages/material/scss/button/_variables.scss
JoomFX: Yes, it was confirmed by the design team that the focus bg should be the hover color, and it is already [added to the kits](https://www.figma.com/file/Bun5XXsG0UuzCMz5GyDoHK/Telerik-%26-Kendo-UI-Kit-for-Material-2.3--(UNPUBLISHED)?type=design&mode=design&t=0vw5kVDgj4XZKPvF-0).

Juveniel (approved):

PR #5011, opened 2024-04-29
chore(html): add mediaplayer spec

zhpenkov:
Add specs and templates for the MediaPlayer component.


packages/html/src/mediaplayer/mediaplayer-toolbar.tsx
```diff
@@ -0,0 +1,74 @@
...
+            <Button fillMode="flat" icon="volume-up" className="k-volume-button"></Button>
+            <div className="k-mediaplayer-volume-wrap">
+                <SliderNormal className="k-mediaplayer-volume" showButtons={false} showTicks={false} />
+            </div>
+            {showQuality &&
+                <div className="k-mediaplayer-quality-wrap">
+                    <DropdownList value={qualityValue} arrowIconName="hd" className="k-mediaplayer-quality" />
+                </div>
+            }
+            <Button fillMode="flat" icon="fullscreen" className="k-fullscreen-button"></Button>
```
Juveniel, line 64: We should fully re-use the Toolbar here. The template items( volume, quality, etc. ) should be rendered inside .k-toolbar-item. See [ref](https://demos.telerik.com/kendo-ui/mediaplayer/index) and [ref2](https://github.com/telerik/kendo-themes/blob/develop/packages/html/src/toolbar/tests/toolbar.tsx#L53).

Juveniel (approved):

PR #5014, opened 2024-04-29
chore(html):  add dock-manager spec

zhpenkov:
Add specs and templates for the DockManager component.


epetrow (approved):
Looks good overall, I am just wondering about two things:

1. Shouldn't we associate the `PaneScrollable` with the `DockManager` as it is a building block?
2. Do we want to add a "spec" for the DockNavigator?

PR #5033, opened 2024-05-10
Improve metadata generation

joneff:
The code is self-explanatory, but here is anyway: values are processed and prettified in that order

1) If the value is instance of `SassMap`, it's prettified as an object, with the values prettified recursively according to the rules hereby;
2) If the value is instance of `SassList`, it's prettified as an array, with the items prettified according to the rules hereby;
3) If the value is instance if `SassString`, it's prettified as string, with quotes removed where possible;
4) If the value is instance of `SassNumber`, it's prettified as a number, if there are not units or string otherwise;
5) If the value is instance of `SassBoolean`, it's prettified as boolean;
6) If the value is equal to `SassNull`, it's prettified as `null`;
7) In all other cases, the value is returned as string.

The prettified value is then stored as a new `prettyValue` field in `variables.json`, which is then copied to `sassdoc-data.json`.


packages/bootstrap/scss/core/color-system/_swatch.scss
```diff
@@ -179,107 +179,254 @@ $_default-colors: (
...
 /// The global default Colors map.
 /// @group color-system
-/// 
+///
 /// @subgroup {misc}
 /// [app-surface, on-app-surface, subtle,
 /// surface, surface-alt, border, border-alt]
 /// The Misc variable group.
+/// @prop {Color} app-surface - The background color of the application.
+/// @prop {Color} on-app-surface - The text color of the application..
```
epetrow, line 188: Extra dot at the end, it's present in all themes

epetrow (commented):
Looks good overall, just one question:

I noticed that we are displaying maps differently in the docs files now -- before the values were wrapped inside `<code>` while now they are inside a `ul`. Are we sure we want to remove the `code` tag?

Juveniel (approved):

epetrow (commented):
I noticed a few inconsistencies in how the data is displayed in the docs files now. It might be expected, however now we have:

Numbers being displayed as numbers - e.g. `thin: 100`

If a value has a unit it is a string - `xxs: "0.5rem"`

Also in some places there are escape characters that are being displayed in the docs now - `serif: "\"Times New Roman\",`

epetrow (approved):

PR #5045, opened 2024-05-23
Add compiled swatches to cdn

Juveniel:
resolves: https://github.com/telerik/kendo-themes-private/issues/63

// cc @dimodi


tsvetomir (approved):
# 🚂 🚋🚋🚋
Great for customers, not so great for npmjs.com, but they'll manage :)

We should encourage the use of the swatches in the docs over the CDN.

JoomFX (approved):

Issue #5046, opened 2024-05-23
[BUG] Component has less height in 8.0.0 with material theme


silviyaboteva:
**Describe the regression**
In 8.0.0, with the kendo material theme, the Dropdownlist's height get shrinked when the component has no value set. 

In the case with no value, nor placeholder set, we skip the `<span class="k-input-value-text">` element inside the `<span class="k-input-inner">` element. Here is the rendering test showcasing it: https://github.com/telerik/kendo-themes/blob/b7349376c2b1e3ab9d3a9b6ac329659c2443df9e/tests/dropdownlist/dropdownlist.html#L21-L30

However, in the test, the actual placeholder text keeps the dropdownlist height. If there is no value, the height of the component will be decreased in the Material theme.

**To reproduce**
Check the demo here: https://github.com/telerik/kendo-react/issues/2274

**Expected behavior**
The height remains the same with or without a value.

**Affected package (please remove the unneeded items)**
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for React

PR #5047, opened 2024-05-23
fix(dropdownlist): keep the height of empty dropdownlist

silviyaboteva:
Fixes https://github.com/telerik/kendo-themes/issues/5046

TODO
- [x] we should ping all teams to update the rendering accordingly (by removing the `k-input-value-text` element if there is no value set.


epetrow:
In my opinion this might not be the best way to handle the issue. I think the problem here stems from the `padding-block` value which we apply to `k-input-button`. For all themes except for Material this value matches the one applied through the sizing classes of the button. 

This causes the Dropdownlist button in Material to be shorter which becomes noticeable when `k-input-value-text` is not rendered or when there is no placeholder.

There also seems to be a difference in the rendering  -- currently in the spec of the Dropdownlist `k-input-value-text` is not rendered if there is no value, but in some suites ( Angular for [example](https://www.telerik.com/kendo-angular-ui/components/dropdowns/dropdownlist/) ) is present event if there is no placeholder or value.

So we should maybe decide:

- do we always render `k-input-value-text` and correct the component spec
- conform to the [design](https://www.figma.com/design/Bun5XXsG0UuzCMz5GyDoHK/Telerik-%26-KENDO-UI-Kit-3.0-for-Material?m=auto&node-id=593-13371) where only the inline padding of the Dropdownlist button seems to be adjusted

Another thing unrelated to the aforementioned, when updating existing or adding new tests, please do so in `packages/html/src/${component-name}/tests`

silviyaboteva:
Thank you @epetrow, I will research for the `pdding-block` solution and apply it instead.

As for the `k-input-value-text` element, I think, it's up to @telerik/kendo-front-end to decide if we have to render it in case no placeholder, nor value is provided. I also checked the Angular and Blazor suites and noticed that there is always rendered, but is this semantically correct? The KendoReact team based on the test that showcase the placeholder without this `k-input-value-text` element. And we also use this rendering in the adaptive DropdownTree and grid filter row tests.

I'll update the `packages/html/src/${component-name}/tests` as well. Thanks!

silviyaboteva:
I double checked [the spec of the DropdownList](https://github.com/telerik/kendo-themes/blob/develop/packages/html/src/dropdownlist/dropdownlist.spec.tsx#L114) and the `k-input-value-text` element is rendered only if there is a value. For reference: https://github.com/telerik/kendo-themes/blob/develop/packages/html/src/input/input-inner-span.tsx#L38. 

So, in this case the fix with the `padding-block` is required in Material theme :)

silviyaboteva:
Based on the comments from the team, I updated the value of the `button-padding-y` to `null`, so the correct padding is set `padding-block` of the `k-input-inner` element.

I also add tests with no placeholder nor value to all inputs, dropdowns and pickers, and for their fill modes.

inikolova (approved):

PR #5053, opened 2024-05-30
Detach utils package

JoomFX:
related to https://github.com/telerik/kendo-themes-private/issues/245


packages/fluent/scss/utils/index.scss
```diff
@@ -1,23 +1,480 @@
...
+    $kendo-border-radii: $kendo-border-radii,
+    $kendo-elevation: $kendo-elevation,
+    $kendo-colors: $kendo-colors,
+    $kendo-enable-color-system: $kendo-enable-color-system
 );
 
+// Accessibility
+@mixin utils--accessibility--screen-readers() {
+    @include kendo-utils--accessibility--screen-readers();
+}
```
Juveniel, line 15: why do we need to wrap each of these, instead of using them directly e.g `@include kendo-utils.mixin-name()` ? In general, this is error prone, as we need to maintain those mixins in all related packages.

epetrow (commented):
Looks good to me, one thing I would suggest is to combine the detach and add utils to components commits into one

packages/html/shared/index.html
```diff
@@ -4,6 +4,7 @@
         <title>Document</title>
         <meta charset="utf-8" />
         <link rel="stylesheet" data-role="kendo-theme" href="/packages/default/dist/all.css" />
+        <link rel="stylesheet" data-role="kendo-theme-utils" href="/packages/utils/dist/all.css" />
```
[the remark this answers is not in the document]
JoomFX: @zhpenkov yes, I also agree that we should consider this so we can catch issues with missing utility classes.
The best approach would be to "Find and Replace" all utility classes that we use in the tests (ie `.k-d-grid` -> `test-d-grid`,  `k-grid-cols-1` -> `test-grid-cols-1`, etc) and add the corresponding styles. The idea is to stop using the utility classes in the test setups.

Doing the above will take some time as we need to figure out what are the utility classes (all of them) that we use in the test setups.
If we do not have this time now, we can do it in a separate PR later on.

packages/default/scss/fab/_index.scss
```diff
@@ -38,6 +37,10 @@ $_kendo-module-meta: (
     @include module-render( "fab" ) {
         @include kendo-fab--layout();
         @include kendo-fab--theme();
+        @include kendo-utils--layout--position();
+        @include kendo-utils--typography--text-align();
+
```
epetrow, line 42: Empty line?

packages/fluent/scss/pivotgrid/index.scss
```diff
@@ -32,5 +34,9 @@ $_kendo-module-meta: (
         @include kendo-pivotgrid--theme();
         @include kendo-pivotgrid-configurator--theme();
         @include kendo-pivotgrid-legacy--theme();
+
```
epetrow, line 37: Are these empty lines left on purpose for readability?

inikolova: Nope, I will remove them

epetrow (approved):
LGTM

Juveniel:
Marking as draft, so that we do not merge this accidentally.

PR #5036, opened 2024-05-15
Migrate sass syntax from legacy to modern syntax

joneff:
I will update as I go. Currently, only the classic theme to see what kind of issues we may encounter.

Before this PR is merged, there will be other things to be worked on:

```[tasklist]
### Tasks
- [ ] https://github.com/telerik/kendo-themes/pull/5061
- [ ] https://github.com/telerik/kendo-themes/pull/5053
- [ ] https://github.com/telerik/kendo-themes/pull/5067
```

PR #5061, opened 2024-06-04
Streamline module dependencies

joneff:
As a precursor to #5036: streamline module dependencies so it's easier to work with the code base:

* ensure dependencies are named the same across the themes;
* ensure dependencies are ordered alphabetically;
* ensure the same dependencies are included across the themes.

I may have missed a "floating-label" dependency here or there, but I will tackle it in the following PR which will deal with using the "centralized" dependency tree.

There should be **NO** output / artifacts from this PR. Unless of course we fix a bug. Apart from that -- no output, no docs changed, no test changed etc.

**Update**

The following changes are expected:

* Colorpicker in nouvelle toolbar, because previously the module name was registered and used differently


packages/bootstrap/scss/forms/_index.scss
epetrow, line 1: Just a thought - do we consider a component as a dependency if we are using some of it's variables?

In some cases it seems we do, for example the forms in Fluent use variables from the Input and Window components which are listed as dependencies there.

If that is the case should we also add them for the rest of the themes since these variables are also being used?

packages/fluent/scss/coloreditor/index.scss
```diff
@@ -21,8 +21,8 @@ $_kendo-module-meta: (
 @include module.register( $_kendo-module-meta... );
 
 // Expose
-@mixin coloreditor-styles() {
-    @include module.render( "coloreditor" ) {
+@mixin color-editor-styles() {
```
epetrow, line 24: Another thing I am curious about - we are streamlining the names of the components here e.g. coloreditor -> coloreditor, colorgradient -> color-gradient, however do we also want the mixin names to also match?

Currently there is a difference between Fluent - `@mixin color-editor-styles()` and the rest of the themes - `@mixin color-editor--styles()`

epetrow:
closing in favour of [#5136](https://github.com/telerik/kendo-themes/pull/5136)

packages/bootstrap/scss/forms/_index.scss
JoomFX: Yes, if Component "A" uses Component "B" variables, then Component "B" is a dependency of Component "A".

packages/fluent/scss/coloreditor/index.scss
JoomFX: It would be great if the names are the same across all themes, however we can leave this as a separate tasks for a future improvement.

If we want the Fluent theme to import the Default theme and customize it via variables (like the other themes do), then we would need to have the same names across all themes.

PR #5064, opened 2024-06-06
fix(spreadsheet): prevent triggering the Open tool in Safari

inikolova:
On Mac in Safari browser -> Go to https://demos.telerik.com/kendo-ui/spreadsheet/index

Clicking on any tool button of the Spreadsheet, this triggers the Open tool and the browser's file selection dialog appears.

CC @dimodi


inikolova:
After further investigation we agreed that the recommended approach requires changes in the discussed button rendering, not tweaking the CSS. An "official" fix will be available when changes are scheduled and implemented 

cc @gdenchevprog

PR #5076, opened 2024-06-12
Material fluent border radius

inikolova:
According to [this issue](https://github.com/telerik/web-components-ux/issues/943#issuecomment-1845388821), there should be values for full border radius in the Material and Fluent themes as well


epetrow (approved):

Juveniel:
this should have been a fix @inikolova, @epetrow

PR #5077, opened 2024-06-14
fix(stepper): remove outline of k-step element

silviyaboteva:
This outline has to be removed for two reasons:
1. The `k-focus` class is added to `li.k-step` element by the reference rendering: https://github.com/telerik/kendo-themes/blob/develop/packages/html/src/stepper/step.tsx#L76
 2. According to the new a11y spec for keyboard navigation, the Stepper should change the focus of steps via `Tab / Shift + Tab`.

### DIFF
--- a/packages/default/scss/stepper/_layout.scss
+++ b/packages/default/scss/stepper/_layout.scss
@@ -40,7 +40,9 @@
 
 
         // Step
-        .k-step { }
+        .k-step {
+            outline: none;
+        }
 
 
         // Step link
--- a/packages/fluent/scss/stepper/_layout.scss
+++ b/packages/fluent/scss/stepper/_layout.scss
@@ -39,6 +39,8 @@
 
         // Step
         .k-step {
+            outline: none;
+
             &.k-disabled {
                 background: transparent;
             }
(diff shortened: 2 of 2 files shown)


inikolova:
LGTM. One suggestion - merge both commits into one

inikolova (approved):

silviyaboteva:
Hi @epetrow and @inikolova,

Thank you for your review! I figured out why an outline appeared in my current implementation of KendoReact Stepper keyboard navigation via Tab / Shift + Tab. I have added both `k-focus` class and `tab-index` attribute to the `li.k-step` element. In that case, the outline appears over the whole step element. 

However, [the referent rendering](https://github.com/telerik/kendo-themes/blob/develop/packages/html/src/stepper/step.tsx#L76) says the `k-focus` should be added to the `li.k-step` element, but the `tab-index` should be added to `a.k-step-link` [according to the aria spec](https://github.com/telerik/web-components-ux/blob/develop/docs/components/stepper/aria.json#L36-L51).

In this case, I will close this PR as it's not needed to remove the outline on `k-step` element.

Thanks once again!

PR #5136, opened 2024-08-05
Streamline module dependencies

epetrow:
recreation of: #5061


JoomFX (approved):
Here the `expansion-panel` and the `forms` dependencies should be renamed to `expander` and `form`:

https://github.com/telerik/kendo-themes/blob/pre-migrate-v2/packages/fluent/scss/chart-wizard/_index.scss#L9

epetrow:
> Here the `expansion-panel` and the `forms` dependencies should be renamed to `expander` and `form`:
> 
> https://github.com/telerik/kendo-themes/blob/pre-migrate-v2/packages/fluent/scss/chart-wizard/_index.scss#L9

Nice catch, I had to change them for the other themes as well

packages/bootstrap/scss/card/_index.scss
```diff
@@ -6,8 +6,8 @@
 $_kendo-module-meta: (
     name: "card",
     dependencies: (
-        "icon",
         "action-buttons",
+        "button",
         "typography",
         // "utils/flex-grid"
```
inikolova: In my opinion, we should remove these "utils"-related comments

Issue #2286, opened 2021-02-08
Import inside class throws an error


JoomFX:
**Describe the bug**
Kendo Themes cannot be imported inside a class when Dart Sass 2+ is used.
The following errors are thrown:

```
DEPRECATION WARNING: As of Dart Sass 2.0.0, !global assignments won't be able to
declare new variables. Consider adding `$imported-modules: null` at the root of the
stylesheet.

DEPRECATION WARNING: As of Dart Sass 2.0.0, !global assignments won't be able to
declare new variables. Consider adding `$data-uris: null` at the root of the
stylesheet.
```

**To reproduce**
Try to compile the following code:

```
.dark-theme {
  $primary-palette-name: yellow;

  @import "~@progress/kendo-theme-material/dist/all.scss";
}
```

**Expected behavior**
Currently, the issue can be overcome by adding the following two lines at the top of you stylesheet:

```
$imported-modules: ();
$data-uris: ();
```

But this should be fixed in the themes.
The code that needs to be looked at is located [here](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/mixins/core/_import-once.scss#L11) and [here](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/mixins/core/_data-uri.scss#L15).

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**
* All

**Build system information (please remove the unneeded items)**
* Not Applicable

ianwesterfield:
👀 Any update on prioritizing this one? Same as described 👆

joneff:
@ianwesterfield The issue is related to how variables are treated when nested inside classes.

To go into details, we convert non text files (images, fonts) to base64 encoded string and append them in an aptly named variable called `$data-uris`. Later, we retrieve the content by key.

When you are using the theme as is, everything is fine and the variable is found.

However when you nest the import things get a bit strange, for a lack of a better word. I am not sure for the exact reason, but in this nested import scenario, when we try to append said encoded content it's not appended in the same location where we try to retrieve it from.

Good news is that this all can be workaround relatively easy: we just need to reset one variable before each such import. Resetting the variable will make it globally available.

[...]

PR #5140, opened 2024-08-09
fix: importing theme in a class creates a wrong selector

inikolova:
Closes https://github.com/telerik/kendo-themes/issues/3444

### DIFF
--- a/packages/bootstrap/scss/menu/_layout.scss
+++ b/packages/bootstrap/scss/menu/_layout.scss
@@ -13,12 +13,8 @@
         padding-inline: 0;
     }
 
-    .k-menu-group,
-    .k-menu-vertical {
-
-        .k-menu-popup & {
-            padding: 0;
-        }
+    .k-menu-popup :is(.k-menu-group, .k-menu-vertical) {
+        padding: 0;
     }
 
     .k-context-menu-popup {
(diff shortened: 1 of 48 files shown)


epetrow (approved):
Looks good to me, improrting the theme inside a class no longer seems to produce invalid selectors -- at least in terms of the Slider.

JoomFX (commented):
UPDATE: Everything seems to be fine. It is just that the import inside a class needs to be performed with the proper syntax, as explained in [this issue](https://github.com/telerik/kendo-themes/issues/2286#issuecomment-1596963083):

```scss
$_imported: (); // reset the list of imported modules by the new modules system making them globally available
$_kendo-imported-modules: (); // reset the list of loaded modules for good measure
$_kendo-data-uris: (); // reset the data-uris thus making them globally available

.dark-theme {
    @import '@progress/kendo-theme-default/dist/all.scss';
}
```

~~It is not particularly related to the changes in this PR, but the compilation fails when trying to import the Default theme in an Angular application:~~

```scss
.dark-theme {
    @import "./all.scss"
}
```

~~The two error come from [here](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/colorgradient/_theme.scss#L62) and [here](https://github.com/telerik/kendo-themes/blob/develop/packages/default/scss/spreadsheet/_layout.scss#L741).
If you comment out these lines, the theme compiles without errors, but still the warnings are there. And in this case I do believe that these warnings are actually an issue, an issue related to our module system or to the changing related to the Dart Sass migration, which needs to be investigated and fixed.~~

error

JoomFX (approved):

Issue #5037, opened 2024-05-15
Provide `k-checkbox-label` when the label is placed on the left side


mbechev:
**Describe the bug**
The existing `k-checkbox-label` and `k-radiobutton-label` classes add space between the label and the input which makes them appear more user-friendly.

<img width="172" alt="image" src="https://github.com/telerik/kendo-themes/assets/33104481/d0fedea0-0301-49e7-8881-b7251c959f3e">

```html
<kendo-checkbox #checkbox1 ></kendo-checkbox>
<kendo-label class="k-checkbox-label" [for]="checkbox1" text="Checkbox Component"></kendo-label>
```

However, there is no available class if the label is placed on the left side of the input:

<img width="160" alt="image" src="https://github.com/telerik/kendo-themes/assets/33104481/9956d981-71b0-415b-a7a4-3482a42618bc">

```html
<kendo-label [for]="checkbox3" text="Checkbox Component"></kendo-label>
<kendo-checkbox #checkbox3 [checkedState]="true"></kendo-checkbox>
```

The workaround is to use an inline style, which has been used for the `k-checkbox-label`:

```html
<kendo-label 
    [for]="checkbox3" 
    text="Checkbox Component" 
    style="display: inline;margin-inline-end: 4px;">
</kendo-label>
```

https://stackblitz.com/edit/angular-q9mv5x?file=src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.module.ts

PR #5147, opened 2024-08-14
fix: add space for label before checkbox / radio button

inikolova:
Related to https://github.com/telerik/kendo-themes/issues/5037


epetrow (commented):
I wonder do we actually need to override the display of the label from `inline-flex` to `inline`? If not maybe we can merge the styles? For example:

```scss
.k-checkbox + .k-label,
.k-label + .k-checkbox,
/* etc. */ {
  margin-inline-start: $kendo-checkbox-label-margin-x;
}
```

Juveniel:
stale, need to revisit this

PR #5149, opened 2024-08-15
HTML Specs & Templates corrections  vol.1

zhpenkov:
related to: https://github.com/telerik/kendo-themes-private/issues/246

#### Items addressed in this PR

**Splitter:**
- remove splitter warnings produced from splitter spec

**PDFViewer/Pager:**
- toolbar button-group has different fillMode than its children buttons
- disabled pager applies k-disabled to the pager and k-disabled to its buttons (pdf-viewer-blank)

**Chip:**
- add chip template for all chip actions/options


epetrow (commented):
The ChipActions template needs to be exported in the [index](https://github.com/telerik/kendo-themes/blob/develop/packages/html/src/chip/index.ts) as well

inikolova:
LGTM. One suggestion, we could add a test with the ChipActions template

inikolova (approved):

Issue #5109, opened 2024-07-16
[ListView] Incorrect position of the loading indicator


tswetti:
**Describe the bug**
Currently, the loading indicator is positioned inside the content (`.k-listview-content`). This results in a scrollable indicator when the ListView is scrollable:

https://github.com/user-attachments/assets/0c4734f7-4416-4adf-a5d8-f3d21adab9fb

**To reproduce**
Angular - https://stackblitz.com/edit/angular-feqwba-wyjbjd?file=src%2Fapp%2Fapp.component.ts

**Expected behavior**
The loading indicator should not be scrollable.

PR #5150, opened 2024-08-20
fix(listview): adjust loading mask rendering

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/5109

The Loading Mask should block the whole component, including the Header, Footer and Pager.
The Grid implementation is a good example:

https://stackblitz.com/edit/angular-mshfaz?file=src%2Fapp%2Fapp.component.ts


inikolova (approved):

packages/html/src/listview/listview.spec.tsx
```diff
@@ -55,7 +55,7 @@ export const ListView = (
     const pager = <Pager className={`k-listview-pager k-listview-pager-${ pagerPosition }`} />;
 
     const loader = <>
-        <div className="k-loading-mask k-opaque" style={{ height: '100%', width: '100%' }}>
+        <div className="k-loading-mask" style={{ height: '100%', width: '100%' }}>
```
epetrow, line 58: Why are we removing the `k-opaque` class here?

epetrow (commented):
Personally I think the position of the loader image looks rather bad now, however I also agree that it should block the whole component

packages/html/src/listview/listview.spec.tsx
JoomFX: The `k-opaque` class makes the whole overlay with a solid white background. Therefore, nothing can be seen behind it.

PR #5184, opened 2024-10-02
Add Chart "No data" template styles

epetrow:
closes: https://github.com/telerik/kendo-themes-private/issues/269

Design: https://github.com/telerik/web-components-ux/issues/1054#issuecomment-2253150390


inikolova:
Chart "No data" template appearance relies on **.k-no-data** rules, which are declared in the list _layout.scss file
https://github.com/telerik/kendo-themes/blob/c2ba24b3ce69eddf6f0ad6e82d1551ac5233dc70/packages/default/scss/list/_layout.scss#L244
I would suggest to move these styles to core, as dataviz does not have dependency to list component

epetrow:
> Chart "No data" template appearance relies on **.k-no-data** rules, which are declared in the list _layout.scss file
> 
> https://github.com/telerik/kendo-themes/blob/c2ba24b3ce69eddf6f0ad6e82d1551ac5233dc70/packages/default/scss/list/_layout.scss#L244
> 
> 
> I would suggest to move these styles to core, as dataviz does not have dependency to list component

Wouldn't it be better to add it as a dependency, like it is in the case of the grid? 🤔

inikolova:
> a

I think the Chart component does not use a list at all, so in my opinion we shouldn't add it as a dependency. "No data" template styles are some kind of "common" styles, so it would be better to  move them in a common folder.

epetrow:
> I think the Chart component does not use a list at all, so in my opinion we shouldn't add it as a dependency. "No data" template styles are some kind of "common" styles, so it would be better to move them in a common folder.

I've extracted the `no-data` styles as a separate "component" and added it as dependency where needed, similar to what we did with the `time-marker`.

inikolova (approved):

tsvetomir (approved):
Looks great and easy to use. Thanks!

PR #5204, opened 2024-11-01
chore: remove custom watcher in favor of sass --watch

Juveniel:
merge after: https://github.com/telerik/kendo-themes/pull/5203


packages/bootstrap/dist.map
epetrow, line 1: I might be missing something but, why are we introducing this file? Is it only needed for the bootstrap theme?

Juveniel: we should remove this file - committed by accident

inikolova (approved):

PR #5207, opened 2024-11-04
Remove margin utils

inikolova:
As pointed in [this issue](https://github.com/telerik/kendo-themes-private/issues/245), the Typography component should suggest only font-related properties, but not space related


silviyaboteva (commented):
Should we add a breaking change in the commit message in order to display it in the changelog?

inikolova:
Adding a breaking change in the message body will bump the themes versions to 11, therefore I removed it.

epetrow (approved):

PR #5133, opened 2024-08-02
fix(grid): missing sticky rows styles

JoomFX:
fixes https://github.com/telerik/kendo-themes/issues/5121

PR #5210, opened 2024-11-04
fix(grid): fix broken sticky row styles

epetrow:
Fixes a regression introduced with https://github.com/telerik/kendo-themes/pull/5133


inikolova:
I think we should remove this row as well, as it is Angular specific -> https://github.com/telerik/kendo-themes/blob/fix-grid-sticky-rows-selection/packages/bootstrap/scss/grid/_theme.scss#L224

Juveniel:
I don't think it is a good idea to return the kendo-grid selectors. Let's discuss this further and search for an alternative.

epetrow:
> I think we should remove this row as well, as it is Angular specific -> https://github.com/telerik/kendo-themes/blob/fix-grid-sticky-rows-selection/packages/bootstrap/scss/grid/_theme.scss#L224
> 
> Also, I think it will be beneficial to add a visual test for covering sticky rows scenario

If we decide to do so due to the current differences in the rendering between the suites we will need to add two tests - one for React locked rows and one for Angular sticky rows

silviyaboteva:
> > I think we should remove this row as well, as it is Angular specific -> https://github.com/telerik/kendo-themes/blob/fix-grid-sticky-rows-selection/packages/bootstrap/scss/grid/_theme.scss#L224
> > Also, I think it will be beneficial to add a visual test for covering sticky rows scenario
> 
> If we decide to do so due to the current differences in the rendering between the suites we will need to add two tests - one for React locked rows and one for Angular sticky rows

Guys, firstly, lets revamp and unify the rendering https://github.com/telerik/kendo-themes-private/issues/283 and after that we can have a single source of truth test.

silviyaboteva (approved):
The React Grid looks as expected. Demo: https://stackblitz.com/edit/react-dolz78-4nmsyx?file=index.html

The Angular Grid also looks as expected. Demo: https://stackblitz.com/edit/angular-5c56gf-c6xfju?file=src%2Findex.html

inikolova (approved):

PR #5214, opened 2024-11-05
Incorrect border radius of the chart tooltip

inikolova:


packages/html/src/chart/tests/chart-tooltip.tsx
```diff
@@ -0,0 +1,30 @@
...
+    }
+`;
+
+
+export default () =>(
+    <>
+        <style>{style}</style>
+        <div id="test-area" className="k-d-grid k-grid-cols-1">
+            <AnimationContainer className="k-chart-tooltip-wrapper ">
+                <div className="k-popup">
```
Juveniel, line 22: you can re-use the `<Popup>` component here.

inikolova: Using the <Popup> component does not suggest the option to set className ("k-chart-tooltip-wrapper") to the AnimationContainer.

inikolova:
I checked the tooltip rendering across all suites and it differs because of some suite-specifics (the main differences are described in this issue -> https://github.com/telerik/kendo-themes-private/issues/187 ). Also, I had a discussion for the chart tooltip with @nstoychev and it appears the chart tooltip does not use the Tooltip component itself (because of the canvas rendering). Therefore my suggestion is to keep the chart tooltip rendering as is at the moment and apply the suggested css rules.

Juveniel (approved):

PR #5226, opened 2024-11-11
chore: re-schedule weekly cron job

Juveniel:
attempt to reset actor of the `Weekly` workflow


epetrow (commented):
According to a github discussion thread it seems like only disabling and re-enabling a workflow will trigger a change in the actor tied to it 🤔

PR #5227, opened 2024-11-11
Html specs corrections 5

TeyaVes:
targets: https://github.com/telerik/kendo-themes-private/issues/246


packages/html/src/datetimepicker/datetimepicker.spec.tsx
epetrow, line 1: The position of the buttons in the ActionSheet should be reversed, check the guidelines [here](https://github.com/telerik/web-components-ux/issues/408#issuecomment-1327396271)

epetrow (commented):
Are the changes in the ImageEditor expected as it does not look good in some of the themes?

TeyaVes:
> Looks okay overall, I am just wondering about two things:
> 
> 1. If we are removing the `k-tab-on-top` class, shouldn't we remove it from `spreadsheet-jquery.tsx` as well? And if it is an obsolete class should't we delete it?
> 2. Are we okay with the difference in the ImageEditor label font weight in Fluent? I know that this is the correct checkbox position, but still.

1. I've removed the k-tab-on-top class from the styles and the specs.

2. I think that we should leave the checkbox with label with the normal font weight - following the logic only the formfield labels are bold.

PR #5228, opened 2024-11-11
add StackLayout and GridLayout component modules

zhpenkov:
Add sass component modules for the GridLayout & StackLayout components.

### DIFF
--- a/packages/bootstrap/scss/index.scss
+++ b/packages/bootstrap/scss/index.scss
@@ -293,6 +293,7 @@
 
 
     // Generic content
+    @include kendo-icon--styles();
     @include kendo-messagebox--styles();
 
 


packages/bootstrap/scss/gridlayout/_index.scss
```diff
@@ -0,0 +1,19 @@
...
+@use "../core/_index.scss" as *;
+@use "../utils/_index.scss" as *;
+
+// Component
+@forward "./_variables.scss";
+@use "./_layout.scss" as *;
+@use "./_theme.scss" as *;
+
+// Expose
+@mixin kendo-grid-layout--styles() {
```
inikolova, line 11: In my opinion we should name the mixin kendo-gridlayout--styles()

packages/bootstrap/scss/index.scss
```diff
@@ -409,6 +411,7 @@
     @include kendo-splitter--styles();
     @include kendo-tile-layout--styles();
     @include kendo-grid-layout--styles();
+    @include kendo-stack-layout--styles();
```
inikolova, line 415: In my opinion we should name the mixin kendo-stacklayout--styles() (as the component is stacklayout)

epetrow (commented):
Just a few thoughts:

- I think it would be a good practice for any new component to be forwarded last in the `index.scss`, this way we will ensure that all dependencies (if any) are loaded. This is a general  suggestion and does not directly apply to this PR.
- Adding the missing icon in Bootstrap is a fix

epetrow (approved):

inikolova (approved):

PR #5232, opened 2024-11-14
Add PropertyGrid Spec

zhpenkov:
Add spec and templates for the PropertyGrid component.


epetrow (approved):
Looks good overall, I have two questions though:

1. Do we need two templates? - PropertyGridWithSplitter sounds a bit misleading to me, I would expect the PropertyGrid to contain the Splitter not the other way around. Also as I look at the component demos it seems to me that  PropertyGridWithSplitter is the actual "normal" state of the component
2. Can't we align the component in the left Splitter pane with utility classes instead?

Issue #4601, opened 2023-07-07
Grid content is not aligned correctly when using k-text-right/left/center classes


mihaela-lukanova:
**Describe the regression**
When the "k-text-left", "k-text-right", or "k-text-center" class is added to the Grid column header or cell, the respective styles are not applied, and the column content is not aligned as expected. If the class is added with "!important", the styles are applied correctly.

**To reproduce**
 A Dojo sample for reproduction: https://dojo.telerik.com/uwAKeYel

**Expected behavior**
The Grid content should be aligned as expected when any of these classes are added to the "headerAttributes" or "attributes" column properties.

**Screenshots**

![image](https://github.com/telerik/kendo-themes/assets/79098346/0b2e106f-d941-4993-9ce3-591f965752e0)

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material
* theme-classic
* theme-fluent

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery

**Affected browsers (please remove the unneeded items)**
* All

**Build system information (please remove the unneeded items)**
* Not Applicable

**Additional context**
The last working version is 2022.1.301

inikolova:
I would like to break the issue down into two parts:

- The Grid header one -  Currently header cells include an additional `<span class="k-cell-inner">` element, which uses Flexbox for styling. Aligning the header cells to the right requires an alternative approach to adjust the Flexbox layout and achieve the desired alignment (for example using the **k-justify-content-end** utility class - https://dojo.telerik.com/BnuvkmKR

- The Grid body part - In the pointed v2022.1.301 version **.k-text-right** has _**!important**_ flag. After introducing a separate kendo-theme-utils package, we provide the option to use a class with or without **!important** flag (https://github.com/telerik/kendo-themes/blob/develop/packages/utils/scss/_mixins.scss#L9C118-L10C13). Hence to get the same result like in the initial example you should use **!k-text-right** class (https://dojo.telerik.com/kjAIBgMv)

Given this, I don't consider the points mentioned above as regressions, so I am closing the issue. However feel free to open it again if needed.

PR #5237, opened 2024-11-19
fix(grid): remove cells text alignment inheritance

inikolova:
Part of https://github.com/telerik/kendo-themes/issues/4601

Note: removing the _text-align: inherit_ rule will address the problem for grid content cells only. Currently header cells include an additional `<span class="k-cell-inner">` element, which used Flexbox for styling. Aligning the header cells to the right requires an alternative approach to adjust the Flexbox layout and achieve the desired alignment. (Updated example from the issue - https://dojo.telerik.com/XTRvnfcl

### DIFF
--- a/packages/core/scss/components/grid/_layout.scss
+++ b/packages/core/scss/components/grid/_layout.scss
@@ -93,7 +93,6 @@
             border-color: inherit;
             outline: 0;
             font-weight: inherit;
-            text-align: inherit;
             position: static;
             overflow: hidden;
             text-overflow: ellipsis;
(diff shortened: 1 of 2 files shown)


TeyaVes:
@inikolova , For the header cell we can add _k-justify-content-end_ utility class and the text inside will be aligned to the right without the need of custom CSS classes and rules.

inikolova:
> SS classes and rules.

Yep, I will post multiple suggestions in the referenced github issue, once this PR is merged :)

Issue #5242, opened 2024-11-25
HTML select value in the adaptive Pager has wrong vertical alignment in Safari


dimodi:
**Describe the bug**

The `select` element that shows on narrow screens in adaptive mode uses similar styles as the DropDownList, including flexbox. It appears that Safari doesn't apply them as expected, which results in wrong vertical alignment of the `select` value.

**To reproduce**

https://dojo.telerik.com/AvAcHMvG

**Expected behavior**

The number in the `select` element should be vertically centered. Currently it is above it's expected centered position.

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material
* theme-fluent
* theme-classic

**Affected suites (please remove the unneeded items)**
* All

**Affected browsers (please remove the unneeded items)**
* Safari (including the latest one as of writing 18.1.1)

PR #5243, opened 2024-11-25
fix(pager): select value in adaptive pager has wrong vertical alignment in Safari

dimodi:
Fixes https://github.com/telerik/kendo-themes/issues/5242

Another option may be to avoid flexbox for the `select` element, but this may need more lines of code.

### DIFF
--- a/packages/default/scss/input/_layout.scss
+++ b/packages/default/scss/input/_layout.scss
@@ -91,6 +91,7 @@
         padding-block: $kendo-input-padding-y;
         padding-inline: $kendo-input-padding-x;
         appearance: auto;
+        align-items: center;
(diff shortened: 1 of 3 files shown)


inikolova:
LGTM :) However I would suggest to cut short the commit message. Also, the "pager" should start with small letter

dimodi:
OK, I fixed the letter casing and shortened the commit message a bit.

TeyaVes:
@dimodi , in the description of the issue you can add "fixes: {link to the targeted issue}" - it this way once the PR is closed, the issue will be automatically closed as well.

inikolova (approved):

Issue #5251, opened 2024-12-03
[Gantt] Resizing indicators cursor styling is misleading


dtopalov:
**Is your feature request related to a problem? Please describe.**
The Gantt Task resizing and completion drag-editing indicators are misleading. When hovering the draggable element, cursor becomes a left or right arrow (depending on whether the user hovers the start or end of the task), and the cursor for the "completion" triangle below the task becomes a right arrow.

https://demos.telerik.com/kendo-ui/gantt/index

**Describe the solution you would like to see implemented**
All cursors should be "resize-x (or -ew)" arrows pointing both left and right, because the user can resize in both horizontal directions for all actions (resizing the task from both ends, dragging the completion indicator to increase/decrease the task completion percentage).

**Additional context**
- Simply replacing `k-resize-e/w` with `k-resize-x` on the respective element does not work as the elements are no longer visible on hover.
- The completion triangle (`k-task-draghandle`) needs special treatment as the generic class brings the cursor styling (not a dedicated cursor-specific class like `k-resize-XXX`):
    ![Image](https://github.com/user-attachments/assets/88c01af6-3c92-4213-823b-0110829661d2)

> [!WARNING]
> If replacing the cursor indicator with a more suitable one is considered a breaking change (rather than a fix) the respective updates can be enforced through the rendering tests in Q2, and we can support both the left/right arrow and the bi-directional one until then.

PR #5252, opened 2024-12-03
fix(gantt): tasks resizing indicators should point both left and right

TeyaVes:
closes: https://github.com/telerik/kendo-themes/issues/5251

The resize handle- k-resize-e and k-resize-w classes are used for positioning the resizing indicators (k-resize-e is with left:0, k-resize-w is with right:0). This is why, we cannot update the classes to k-resize-ew or if we update them, we'll need to add a new class used for positioning. Thus, in order to keep the same rendering, I've set the `cursor: ew-resize` to the` .k-resize-handle` selector.

> [!NOTE]
> Some operating systems such as Windows use the same bidirectional horizontal arrow to represent all three cursors- e-resize, w-resize and ew-resize. The same though, is not the case in macOS where the arrows are single-headed for the e-resize and w-resize cursors.


inikolova (approved):

epetrow (approved):
LGTM

Issue #5247, opened 2024-11-28
$kendo-input-shadow variable is unused


epetrow:
**Describe the bug**
Currently if you try to customize the base box-shadow of the Input through the `$kendo-input-shadow` variable, there will be no effect as the variable is never used.

**To reproduce**

1. Go to https://stackblitz.com/edit/stackblitz-starters-zsbnfm?file=styles.scss.
1. Observe that even though `$kendo-input-shadow` has been customized there is no visual effect

**Expected behavior**
You should be able to customize the base input box-shadow through the given variable

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Additional context**
Reported in support ticket 1671836

PR #5261, opened 2024-12-06
fix(input): allow base input box-shadow customization

epetrow:
closes: [5247](https://github.com/telerik/kendo-themes/issues/5247)


TeyaVes (commented):
LGTM , I am just wondering whether we need to add a declaration and variable in the Fluent theme as well.

TeyaVes (approved):

epetrow:
> LGTM , I am just wondering whether we need to add a declaration and variable in the Fluent theme as well.

I don't think we have the same concept when it comes to shadows in our Fluent inputs, but I might be mistaken.

Issue #5256, opened 2024-12-04
Link annotations are not clickable


Stamo-Gochev:
**Describe the bug**

Link annotations are not clickable. The bug is caused by a mixture of https://github.com/telerik/kendo-pdfviewer-common/issues/93 for the JS part and missing styles.

Here is a list of styles that are required for the fix (they can be rewritten):

```css
.k-annotation-layer section {
      position: absolute;
      pointer-events: auto;
}

:is(.k-link-annotation) > a {
    position: absolute;
    top: 0;
    left: 0;
    width: 100%;
    height: 100%;
}

```

See [the reference](https://github.com/mozilla/pdf.js/blob/11ce57ac2236d06e49b078e1759d271432f3799a/web/annotation_layer_builder.css#L115-L140)

There is also a problem with this style:

```css
.k-pdf-viewer-pages .k-page .k-annotation-editor-layer {
    inset: 0;
}
```
which prevents the link HTML elements to be hovered and clicked. However, setting something like `inset: unset` does not work (the links are clickable, but the canvas section is mismatched on interaction.

**To reproduce**

See https://github.com/telerik/kendo-pdfviewer-common/issues/93

1. Go to https://demos.telerik.com/blazor-ui/pdfviewer/overview
2. Load a file with a hyperlink
3. Try to click on the link to open it. 

> [!NOTE]
> The hyperlink also cannot be clicked if a highlight annotation is done over it and the user returns in text selection mode. This means that hyperlinks do not work both in text selection mode with or without annotations.

**Expected behavior**
The link is clickable. You can test the expected behavior on https://mozilla.github.io/pdf.js/web/viewer.html

**Actual behavior**
The link cannot be opened.

**Screenshots**

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material
* theme-tasks

**Affected suites (please remove the unneeded items)**
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**
* All

inikolova:
After further investigation, it appears that the issue with the un-clickable HTML link elements is not related to the _**inset**_ property, as initially suspected. Instead, the root cause stems from the **_pointed-events_** property.

PR #5263, opened 2024-12-06
fix(pdf-viewer): link annotations are not clickable

inikolova:
Fixes: https://github.com/telerik/kendo-themes/issues/5256

Notes:
As discussed with @Stamo-Gochev, both _**k-link-annotation**_ and _**k-highlight-editor-disabled**_ would be added to the component (currently they don't present)

### DIFF
--- a/packages/default/scss/pdf-viewer/_layout.scss
+++ b/packages/default/scss/pdf-viewer/_layout.scss
@@ -149,6 +149,7 @@
 
                 section {
                     position: absolute;
+                    pointer-events: auto;
                 }
 
                 .k-annotation-text-content {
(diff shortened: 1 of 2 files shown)


packages/default/scss/pdf-viewer/_layout.scss
```diff
@@ -149,6 +149,7 @@
 
                 section {
                     position: absolute;
+                    pointer-events: auto;
```
epetrow, line 152: Do we need to explicitly add `pointer-events: auto`? The element will behave as it would if the pointer-events property were not specified. 🤔

epetrow (approved):

TeyaVes (approved):

PR #5266, opened 2024-12-09
Adaptive Toolbar

TeyaVes:
targets: https://github.com/telerik/kendo-themes-private/issues/299

#  Toolbar Rendering Details

## Scrolling Toolbar
- Add class `k-toolbar-scrollable` if scrolling is enabled
- Add class `k-toolbar-scrollable-overlay ` only if we want to display the gradient overlay - this is in the case when the Toolbar is scrollable, there are enough items for scrolling and the scrollButtons are hidden
- When scrolling is enabled, wrap the toolbar items in `k-toolbar-items` wrapper

<details>
  <summary>Details</summary>

```
<div class="k-toolbar">
    <div class="k-toolbar-items">
        <Button />
        <MenuButton />
        ...
    </div>
</div>
```

</details>

- The scrollButtons should be outside the container separated by a Toolbar separator with class `k-toolbar-button-separator`
- The scrollButtons should have `k-toolbar-prev` and `k-toolbar-next `classes
- Add `k-toolbar-scroll-items` class to `k-toolbar-items` element if we want to enable the native browser scroll. Otherwise, don’t add the class and handle the scrolling in the component
- Add classes `k-toolbar-scrolling-start`, `k-toolbar-scrolling-end `- if we are at the beginning of the scrolling (at the very left) or at the end. This takes care of the gradient overlay when the toolbar is scrollable but the buttons are not visible.
- Enable/disable the scrollButtons if the scrolling toolbar is at the very start or very end (similarly to the current scrollable Tabstrip behaviour)

## Overflow section
- The overflow section is opened inside a Popup component and the toolbar items are wrapped inside an `k-toolbar-items-list`
- fillMode and size classes are added to `k-toolbar-items-list` - `k-toolbar-items-list-solid, -flat, -outline; k-toolbar-items-list-sm, -md, -lg` depending on the Toolbar component options
    - The fillMode is responsible for the color of the separator inside the overflow section; the size - for the paddings and gaps between the Toolbar items inside the overflow section
- The fillMode and size component options inside the k-toolbar-items-list are propagated to the components inside

## Global changes:
- Add class `k-toolbar-separator` to the separators
- Add toolbar separator next to the scroll and overflow buttons in all scenarios with class `k-toolbar-button-separator`

<details>
  <summary>Details</summary>

```
<div class="k-toolbar">
    <Button class="k-toolbar-prev">
    <Button class="k-toolbar-next">
    <ToolbarSeparator class="k-toolbar-button-separator">
    <div class="k-toolbar-items">
        <Button />
        <MenuButton />
        ...
    </div>
</div>


<div class="k-toolbar">
    <Button class="k-toolbar-prev">
    <ToolbarSeparator class="k-toolbar-button-separator">
    <div class="k-toolbar-items">
        <Button />
        <MenuButton />
        ...
    </div>
    <ToolbarSeparator class="k-toolbar-button-separator">
    <Button class="k-toolbar-next">
</div>


<div class="k-toolbar">
    <Button />
    <MenuButton />
    ...
    <ToolbarSeparator class="k-toolbar-button-separator">
    <Button class="k-toolbar-overflow-button">
</div>
```

</details>


packages/default/scss/toolbar/_layout.scss
```diff
@@ -39,6 +39,44 @@
...
+            &::after {
+                content: '';
+                height: 100%;
+                aspect-ratio: 1;
+                position: absolute;
+                inset-inline-end: 0;
+                z-index: 3;
+            }
+
+            &.k-toolbar-scrollable-start::before {
```
epetrow, line 64: Can't these styles be combined like so:

```scss
&.k-toolbar-scrollable-start::before,
&.k-toolbar-scrollable-end::after {
  display: none;
}
```

> NB: Also applies to the Fluent theme

packages/default/scss/toolbar/_layout.scss
```diff
@@ -148,11 +223,39 @@
...
+        }
+
+        .k-toolbar-items-list-#{$size} {
+            padding-block: $_padding-y;
+            padding-inline: $_padding-x;
+            gap: $_spacing;
         }
     }
+
+    .k-toolbar-sm .k-toolbar-button-separator {
```
epetrow, line 247: I think these styles can be improved. You could include the calculated size of the button-separator in the `$kendo-toolba-sizes` instead of having separate rules for each button-separator in each size. 

> NB: Also applies to the Fluent theme

packages/default/scss/toolbar/_layout.scss
```diff
@@ -39,6 +39,44 @@
...
+            &.k-toolbar-scrollable-start::before {
+                display: none;
+            }
+
+            &.k-toolbar-scrollable-end::after {
+                display: none;
+            }
+        }
+
+        &.k-toolbar-scrollable-buttons {
```
epetrow, line 73: Maybe I am misunderstanding but doesn't this 

> Add class k-toolbar-scrollable if scrolling is enabled
> Add class k-toolbar-scrollable-buttons if scrollable buttons are enabled

suggest that `k-toolbar-scrollable-buttons` should be scoped to `k-toolbar-scrollable` as well

> NB: Also applies to the Fluent theme

epetrow (changes requested):
Two additional things that I am wondering:

1. Is overflow `true` by default since it does look like it from the tests?
2. Should we split the `toolbar-section` tests into two since we are showing sizing and fill mode in one?

TeyaVes:
> Two additional things that I am wondering:
> 
> 1. Is overflow `true` by default since it does look like it from the tests?
> 2. Should we split the `toolbar-section` tests into two since we are showing sizing and fill mode in one?

1. By default, the toolbar is with width: 100% and wraps if the content is wider. The other options are resizable with overflow menu or overflow section and scrollable with buttons or no buttons.
2. I've split the toolbar-section tests in two.

packages/html/src/toolbar/tests/toolbar-scrollable.tsx
```diff
@@ -0,0 +1,40 @@
...
+            </section>
+
+            <span className="k-colspan-all k-col-span-full">Toolbar Buttons End</span>
+            <section className="k-colspan-all k-col-span-full">
+                <ToolbarScrollable scrollButtons="end" />
+            </section>
+
+            <span className="k-colspan-all k-col-span-full">Toolbar No Buttons Scroll Start</span>
+            <section className="k-colspan-all k-col-span-full">
+                <ToolbarScrollable scrollButtons="hidden" scrollingPosition="start"/>
```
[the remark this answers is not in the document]
inikolova: In my opinion this behaviour is correct -> scrollingPosition="start" indicates that we are at the beginning, with all items from the start visible. Items towards the end are hidden, and a gradient overlay appears at the end, suggesting that there is more content to scroll through.

packages/bootstrap/scss/toolbar/_variables.scss
```diff
@@ -43,6 +43,19 @@ $kendo-toolbar-md-spacing: k-spacing(2) !default;
 /// @group toolbar
 $kendo-toolbar-lg-spacing: k-spacing(2.5) !default;
 
+/// The height of the Toolbar button separator.
+/// @group toolbar
+$kendo-toolbar-separator-height: null !default;
```
epetrow, line 48: I think we should either set this variable to `$kendo-button-md-calc-size` and reuse it for `$kendo-toolbar-md-separator-height` or remove it since currently it is not used anywhere.

> NB: Applies to all themes

packages/default/scss/toolbar/_theme.scss
```diff
@@ -108,6 +155,40 @@
...
+                &::before {
+                    background: linear-gradient(270deg, $kendo-toolbar-outline-scroll-overlay);
+                }
+
+                &::after {
+                    background: linear-gradient(90deg, $kendo-toolbar-outline-scroll-overlay);
+                }
+            }
+
+            &.k-toolbar-flat {
```
epetrow, line 180: @The linear-gradient variables below should be `$kendo-toolbar-flat-scroll-overlay`

epetrow (changes requested):

packages/bootstrap/scss/toolbar/_variables.scss
TeyaVes: I'm using it in the layout:

```
// Overflow separator
.k-toolbar-button-separator {
    height: $kendo-toolbar-separator-height;
    margin-inline-start: auto;
}
```

TeyaVes:
The visual changes in the grid-angular tests are expected as previosly, the toolbar inside the grid didn't have a sizing class and no padding were applied, which is not correct.

epetrow (approved):

packages/html/src/toolbar/templates/toolbar-scrollable-buttons-hidden.tsx
```diff
@@ -0,0 +1,28 @@
+import { Toolbar } from "../toolbar.spec";
+import { Button } from '../../button';
+import { ButtonGroup } from '../../button-group';
+import { MenuButton } from '../../menu-button';
+import { SplitButton } from '../../split-button';
+
+export const ToolbarScrollableButtonsHidden = ({scrollingPosition = "start", ...props}: any) => (
+    <Toolbar scrollable scrollButtonsbuttons="hidden" scrollingPosition={scrollingPosition} {...props} children={[
```
inikolova, line 8: There is a typo in the prop -> it should be scrollButtons

inikolova (approved):

PR #5272, opened 2024-12-13
feat(dataviz): enable css variable support

epetrow:
closes: [#295](https://github.com/telerik/kendo-themes-private/issues/295)

### DIFF
--- a/packages/fluent/scss/dataviz/_layout.scss
+++ b/packages/fluent/scss/dataviz/_layout.scss
@@ -15,26 +15,6 @@
         }
     }
 
-    // Exported variables
-    // stylelint-disable selector-class-pattern
-    .k-var--chart-font {
-        font-size: var( --kendo-chart-font-size, #{$kendo-chart-font-size} );
-    }
-
-    .k-var--chart-title-font {
-        font-size: var( --kendo-chart-title-font-size, #{$kendo-chart-title-font-size} );
-    }
-
-    .k-var--chart-pane-title-font {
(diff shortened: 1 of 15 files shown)


packages/fluent/scss/dataviz/_layout.scss
```diff
@@ -475,22 +455,52 @@
...
+            } @else if string.index($name, "font-size") {
+                font-size: var(--kendo-chart-#{$name});
+            } @else if string.index($name, "font-weight") {
+                font-weight: var(--kendo-chart-#{$name});
+            } @else if string.index($name, "opacity") {
+                opacity: var(--kendo-chart-#{$name});
             } @else if $type == "number" {
                 // margin-top can store positive & negative values
-                margin-top: $value;
+                margin-top: var(--kendo-chart-#{$name});
```
TeyaVes, line 501: Is the margin-top used at all? I don't see any margins exported in the table, so I think we can delete this as well (we can still double check with the Dataviz team)

packages/fluent/scss/dataviz/_layout.scss
```diff
@@ -475,22 +444,111 @@
...
+        chart-series-28: $kendo-series-28,
+        chart-series-29: $kendo-series-29,
+        chart-series-30: $kendo-series-30,
+
+        chart-gauge-pointer: $kendo-series-f,
+        chart-gauge-track: if($kendo-enable-color-system, k-color( base-emphasis ), k-get-theme-color-var( neutral-30 ))
+
+    );
+
+    .k-chart,
```
inikolova, line 547: I think we should add ._k-sparkline_ class here as well

inikolova (approved):

PR #5293, opened 2025-01-10
Setup SCSS unit tests

zhpenkov:
related to: https://github.com/telerik/kendo-themes-private/issues/248

The unit tests are located in the `/units` folder, which contains subfolders for each theme's tests at `/units/{theme}`. Additionally, there is a `utility.js` file that holds the main logic for testing the **components** and **modules**.

**Run unit tests by using the predefined scripts**:
 - `npm run test:units` - run all unit tests
 - `npm run test:{theme} --prefix units` - run theme unit tests
 
 **Alternatively, directly use the jest CLI and run any specific test:**
 - `jest units` run all unit tests
 - `jest units/{theme}` run theme unit tests
 - `jest units/{theme}/{component}.test.js` run component unit test
 
 Use `npx jest {command}` if you do not have jest globally installed.

**Prerequisites**:
- `npm run sass`
- `npm run docs`

### DIFF
--- a/packages/fluent/docs/customization-action-buttons.md
+++ b/packages/fluent/docs/customization-action-buttons.md
@@ -87,6 +87,16 @@ The following table lists the available variables for customization.
     <td colspan="4" class="theme-variables-description-container"><div><b>Description</b><div class="theme-variables-description">The border color of the action buttons container.</div></div>
     </td>
 </tr>
+<tr>
+    <td>$kendo-actions-button-spacing</td>
+    <td>String</td>
+    <td><code>k-spacing(2)</code></td>
+    <td><code>var(--kendo-spacing-2, 0.5rem)</code></td>
+</tr>
+<tr>
+    <td colspan="4" class="theme-variables-description-container"><div><b>Description</b><div class="theme-variables-description">Spacing between the action buttons.</div></div>
+    </td>
+</tr>
 </tbody>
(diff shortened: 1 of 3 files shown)


.github/workflows/_test-units.yml
```diff
@@ -0,0 +1,57 @@
+name: Test Units
+
+on:
+  workflow_call:
+    inputs:
+      node-version:
+        description: 'Version of node to use'
+        required: false
+        type: number
+        default: 20
```
epetrow, line 10: Do we specifically target v20 or would we rather do `20.x` as in most of the other workflows?

.github/workflows/_test-units.yml
```diff
@@ -0,0 +1,57 @@
...
+          key: root-node_modules-${{ steps.setup-node.outputs.node-version }}-${{ hashFiles('package-lock.json') }}
+
+      - name: Install
+        if: steps.cache-root-node_modules.outputs.cache-hit != 'true'
+        run: |
+          npm ci --no-audit --no-fund
+
+      - name: Build test assets
+        run: |
+          npm run sass --prefix packages/core
```
epetrow, line 50: Wouldn't this also would if we call `npm run docs` just once after the themes are compiled?

epetrow (approved):

PR #5305, opened 2025-01-22
fix(dataviz): chart legend font-size is incorrect

epetrow:


TeyaVes (approved):

inikolova:
A quick question, justo to make sure -> in Bootstrap **$kendo-chart-label-font-size** should be **_small_**, right?
https://github.com/telerik/kendo-themes/blob/develop/packages/bootstrap/scss/dataviz/_variables.scss#L94

epetrow:
> A quick question, justo to make sure -> in Bootstrap **$kendo-chart-label-font-size** should be **_small_**, right? https://github.com/telerik/kendo-themes/blob/develop/packages/bootstrap/scss/dataviz/_variables.scss#L94

Yes, it should remain small since it matches the previous font-size.

inikolova (approved):

PR #5333, opened 2025-02-06
feat(html): add AI Powered Editor Inline Prompt spec

TeyaVes:


packages/html/src/prompt/prompt.spec.tsx
epetrow, line 1: Why are we removing the `KendoPrompProps`? Are they no longer needed?

TeyaVes: It seems that we are not using positionMode prop in the Prompt and the KendoPromptProps don't seem to be needed.

epetrow (approved):

PR #5337, opened 2025-02-10
Cell row span

inikolova:


epetrow (changes requested):
Some general observations from my end:

- The naming `.k-table-td-row-span, $kendo-grid-cell-row-span` seem rather misleading from an API point of view. As a user I would expect this to be responsible for how many rows a cell would span while we are applying box-shadow.
- With the current implementation there are double borders, are we okay with that?
- Also I don't know if this would be considered as part of the issue, but I think we should think about how these cells behave when a row is hovered e.g.
image

inikolova:
@epetrow thank you for the review :) Up to the points: 

_- The naming .k-table-td-row-span, $kendo-grid-cell-row-span seem rather misleading from an API point of view...._
Me and @dtopalov chose to use .**k-table-td-row-span** class name for consistency with **k-table-td** (**k-table-td-row-span** is always set to **k-table-td** element). Regarding the variable name, I agree with you and changed it to **_$kendo-grid-cell-row-span-shadow_** - do you think it's better this way?
_- With the current implementation there are double borders, are we okay with that?_
Yep, this outcome has been synced with @dmanova;
_- Also I don't know if this would be considered as part of the issue, but I think....._ 
The hover behaviour has been synced with @dmanova and @dtopalov

packages/core/scss/components/grid/_theme.scss
```diff
@@ -91,6 +91,10 @@
             background-color: $kendo-grid-hover-bg;
         }
 
+        .k-table-td-row-span {
+            box-shadow: $kendo-grid-cell-row-span-shadow;
```
Juveniel, line 95: Is this a generic enhancement for the Table ? If so, the respective styles should be part of the Table module.

epetrow (approved):

packages/core/scss/components/grid/_theme.scss
inikolova: This is rather enhancement for the Grid component, so in my opinion these styles should be part of the Grid module.

inikolova: We further discussed this with @TeyaVes and @zhpenkov and agreed to move these styles to the Table component

TeyaVes (approved):

PR #5343, opened 2025-02-17
Drop legacy Colour System support

epetrow:
closes: [#324](https://github.com/telerik/kendo-themes-private/issues/324)


TeyaVes (commented):
I would suggest we figure out a consistent way of using the k-color function- either with spaces or no spaces, because I noticed that we are using it in different ways:
- k-color(primary)
- k-color( primary )
- k-color( primary)
- k-color(primary )

gulpfile.js
```diff
@@ -83,75 +89,35 @@ function swatchJsonTransformer( json ) {
...
+        sassContent.push(colorsMap.map( (color) => `\t${color.name}: ${color.value},`).join( '\n' ));
+        sassContent.push(`\t) !default,`);
+    }
 
-            if ( colorsMap.length ) {
-                sassContent.push(`\n$kendo-colors: (`);
-                sassContent.push(colorsMap.map( (color) => `\t${color.name}: ${color.value},`).join( '\n' ));
-                sassContent.push(`);\n`);
-            }
+    // Universal variables are also included here as they are part of the a11y swatch
```
TeyaVes, line 110: I'm not sure if we need the comment about the universal variables here as it's kind of misleading

inikolova (approved):

PR #5347, opened 2025-02-20
Fix drawer icon color

TeyaVes:
Add support for the scenario when the icons inside the drawer are with a different theme color by default, but on hover, focus, selected states, they should have the text color as otherwise they might be not visible


packages/html/src/drawer/tests/drawer-item-states.tsx
```diff
@@ -9,17 +10,44 @@ const styles = `
...
+            <section>
+                <DrawerNormal items={
+                    <>
+                        {[ 'normal', ...DrawerItem.states ].map((state) => (
+                            <>
+                                <li className={`k-drawer-item k-${state}`}>
+                                    <Icon icon="star" themeColor="primary" />
+                                    <span className="k-item-text">{`Item state: ${state}`}</span>
+                                </li>
+                            </>
```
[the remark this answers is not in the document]
epetrow: I also agree that this could be a property in the DrawerItem spec

packages/html/src/drawer/tests/drawer-item-states.tsx
```diff
@@ -9,17 +10,44 @@ const styles = `
...
-                    {[ 'normal', ...DrawerItem.states ].map((state) => (
-                        <DrawerItem text={`Item state: ${state}`} {...{ [state]: true }} icon="star-outline" />
-                    ))}
-                    <DrawerItem text="Item state: selected + hover" selected hover icon="star-outline" />
-                    <DrawerItem text="Item state: selected + focus" selected focus icon="star-outline" />
-                </>
-            }>
-            </DrawerNormal>
+        <div id="test-area" className="k-d-grid k-gap-6">
+            <span>Drawer with icons</span>
```
epetrow, line 14: Maybe we could change this to: Drawer with outline icons ?

TeyaVes: Does it matter? The idea here is that in the second case the theme color of the icons is changed, while here they are normal- it doesn't really matter if its and outline or solid icon

packages/html/src/drawer/tests/drawer-item-states.tsx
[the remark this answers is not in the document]
TeyaVes: okay, I'll add it as a property

TeyaVes:
> This fixes this concrete scenario, but do the actual components have the option to change the **themeColor** of the **Icon** of the **Drawer Item**? Is this achievable from the API of the component in the first place.

This can be achieved with custom rendering and template and we have this case in the page templates so I think it would be best if we fix it in the themes. Plus, we have similar styles in the fluent theme because there the drawer icons are different color than the text by default.

packages/html/src/drawer/tests/drawer-item-states.tsx
epetrow: I think it matters, but I'll leave it for you to decide

TeyaVes: In the second example, I've replaced the outline icon with solid so that the change in the color is more noticeable. I'll replace all icons for consistency.

inikolova:
To me this scenario looks totally valid and I don't consider it as a custom one. In my opinion this is a good enhancement and we should include it to the themes.

inikolova (approved):

PR #5349, opened 2025-02-20
chore(html): correct minor spec discrepancies

zhpenkov:
Adding some unintentionally overlooked details in the component's specs.

These corrections do **NOT** result in any rendering changes.


epetrow (approved):
LGTM, maybe adding an optional floating label to the spec should be a `feat`

TeyaVes (approved):

PR #5355, opened 2025-02-28
feat: update to material v3

inikolova:
Design -> https://www.figma.com/design/2hhKcJiRYUTBZuAb72Sbk3/COPY-Telerik-%26-Kendo-UI-Kit-3.2-for-Material-(UNPUBLISHED)-(Copy)?node-id=13580-107079&m=dev


packages/classic/scss/drawer/_variables.scss
```diff
@@ -105,11 +117,16 @@ $kendo-drawer-selected-text: $kendo-selected-text !default;
 
 /// The background color of the selected and hovered Drawer item.
 /// @group drawer
+// $kendo-drawer-selected-hover-bg: $kendo-selected-hover-bg !default;
```
TeyaVes, line 120: remove this

packages/material/scss/slider/_variables.scss
```diff
@@ -37,10 +37,10 @@ $kendo-slider-button-focus-shadow: null !default;
 
 /// The thickness of the Slider track.
 /// @group slider
-$kendo-slider-track-thickness: 2px !default;
+$kendo-slider-track-thickness: 4px !default;
```
epetrow, line 40: We could also replace this with a value from the spacing module, as done with the other variables below

packages/material/scss/tabstrip/_variables.scss
```diff
@@ -146,7 +146,7 @@ $kendo-tabstrip-item-selected-font-weight: null !default;
 
 /// The shadow of the focused TabStrip items.
 /// @group tabstrip
-$kendo-tabstrip-item-focus-shadow: $kendo-list-item-focus-shadow !default;
+$kendo-tabstrip-item-focus-shadow: inset 0 0 0 2px color-mix(in srgb, currentColor 20%, transparent) !default; // $kendo-list-item-focus-shadow
```
epetrow, line 149: Does the comment mean that the value should be replaced with the variable or is it just for reference. If it is for reference, maybe it can be removed?

packages/core/scss/components/input/_layout.scss
```diff
@@ -518,11 +518,7 @@
                 box-sizing: content-box;
             }
         }
-        .k-dropdown-operator.k-picker-#{$size} {
```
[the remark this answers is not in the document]
TeyaVes: We don't need it anymore - the dropdown operator is a button and it'll take the button padding, with the new material guidelines overriding the button padding with the input one causes layout misalignment - in the rest of the themes, everything looks as expected without this additional styles.

epetrow (approved):
LGTM

PR #5361, opened 2025-03-04
ColumnMenu enhancements

epetrow:
closes: [#344](https://github.com/telerik/kendo-themes-private/issues/344)

enabler for: [#336](https://github.com/telerik/kendo-themes-private/issues/336), [#337](https://github.com/telerik/kendo-themes-private/issues/337) and [#340](https://github.com/telerik/kendo-themes-private/issues/340)

### DIFF
--- a/packages/core/scss/components/grid/_layout.scss
+++ b/packages/core/scss/components/grid/_layout.scss
@@ -1266,6 +1266,27 @@
                 align-items: center;
             }
         }
+
+        .k-columnmenu-indicators {
+            padding-inline: k-spacing(1);
+            display: flex;
+            align-items: center;
+            justify-content: center;
+        }
+
+        // TODO: Check if k-columnmenu-item should not be flex by default
+        .k-columnmenu-item:has(> .k-columnmenu-item-actions, > .k-columnmenu-indicators) {
+            display: flex;
+            align-items: center;
+        }
+
+        .k-columnmenu-drag-handle {
+            margin-inline-end: k-spacing(1.5);
+            display: inline-flex;
+            align-items: center;
+            justify-content: center;
+            cursor: move;
+        }
     }
     .k-column-menu-tabbed {
         border-radius: $kendo-tabstrip-item-border-radius $kendo-tabstrip-item-border-radius 0 0;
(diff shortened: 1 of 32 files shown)


packages/core/scss/components/grid/_layout.scss
```diff
@@ -1266,6 +1266,13 @@
                 align-items: center;
             }
         }
+
+        .k-columnmenu-indicators {
+            padding-inline: k-spacing(1);
```
inikolova, line 1271: Maybe we could define a variable here?

packages/core/scss/components/grid/_layout.scss
```diff
@@ -1277,7 +1277,15 @@
         .k-columnmenu-item:has(> .k-columnmenu-item-actions) {
             display: flex;
             align-items: center;
-            justify-content: space-between;
+            justify-content: center;
+        }
+
+        .k-columnmenu-drag-handle {
+            margin-inline-end: k-spacing(1.5);
```
inikolova, line 1284: Maybe we could use a variable here?

packages/core/scss/components/grid/_layout.scss
epetrow: I don't want to add ColumnMenu variables at this point as the styles are spread out as it is. I think it would be better if we do this when we are consolidating the styles.

packages/core/scss/components/grid/_layout.scss
epetrow: I have the same opinion here, maybe it is not such a good idea to add variables at this moment.

packages/html/src/column-menu/column-menu-expander.spec.tsx
```diff
@@ -10,7 +10,9 @@ const options = {};
 
 export type KendoColumnMenuExpanderProps = {
     itemText?: string;
-    itemIcon?: string;
+    itemStartIcon?: string;
+    itemEndIcon?: string;
+    itemSortIndex?: number;
```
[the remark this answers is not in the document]
epetrow: I went with this to be compliant with the existing "API" because of the existing itemText prop, but we could change the names as you suggested.

inikolova (approved):

PR #5362, opened 2025-03-04
ActionSheet Component Improvements

zhpenkov:
closes https://github.com/telerik/kendo-themes-private/issues/342 
enabler for: https://github.com/telerik/kendo-themes-private/issues/336, https://github.com/telerik/kendo-themes-private/issues/337, https://github.com/telerik/kendo-themes-private/issues/340

## Style changes
Added Header and Footer border for the non-adaptive **ActionSheet**
Added Footer border for the adaptive **ActionSheet** 
Added scrollable behavior of the **ActionSheet** content.

Added styles for `justify` **ActionButtons** alignment
Corrected styling for vertical **ActionButtons** orientation - buttons are now always stretched.

## Rendering changes
Removed `!k-overflow-hidden` class from `k-actionsheet-content`.
Removed `<div class="k-actionsheet-subtitle k-text-center"></div>` element when no content inside.

Added test (_action-sheet-actions_) with **ActionSheet** actions at the start and the end of the header, rendering:

```html
<div class="k-actionsheet-titlebar-group k-hbox">
  <div class="k-actionsheet-actions">
    ...
  </div>
  <div class="k-actionsheet-title">...</div>
  <div class="k-actionsheet-actions">
    ...
  </div>
</div>
```

Updated Specs and added (_action-sheet-action-buttons, action-sheet-adaptive_) tests.


packages/core/scss/components/action-sheet/_layout.scss
```diff
@@ -124,6 +126,10 @@
     // Actionsheet footer
     .k-actionsheet-footer {
         flex: none;
+        border-width: 0;
```
inikolova, line 129: In my opinion we don't need border-width: 0;

packages/core/scss/components/action-sheet/_layout.scss
```diff
@@ -124,6 +126,10 @@
...
+        border-top-width: $kendo-actionsheet-footer-border-width;
+        border-style: solid;
+        border-color: transparent;
```
inikolova, line 132: The border-color should be defined in the _theme.scss file

TeyaVes: exactly, I don't think that we need transparent border here

packages/html/src/action-sheet/actionsheet-header.tsx
```diff
@@ -7,8 +7,10 @@ export const ACTIONSHEETHEADER_CLASSNAME = `k-actionsheet-titlebar`;
 export type KendoActionSheetHeaderProps = {
     title?: string;
     subTitle?: string;
-    actions?: string[];
+    actionsStart?: string[];
```
TeyaVes, line 10: what if you have 1 action only? I don't this this would work. Also, I am not sure if actionsStart/end is the best name here- maybe actionsTitlebar and actionsFooter?

packages/fluent/scss/action-sheet/_layout.scss
```diff
@@ -144,6 +146,10 @@
     // Actionsheet footer
     .k-actionsheet-footer {
         flex: none;
+        border-width: 0;
+        border-top-width: $kendo-actionsheet-footer-border-width;
+        border-style: solid;
+        border-color: transparent;
```
inikolova, line 152: Remove the transparent bg from the Fluent theme as well

TeyaVes (approved):

inikolova (approved):

PR #5372, opened 2025-03-17
Adaptive & Responsive Pager

zhpenkov:
closes: https://github.com/telerik/kendo-themes-private/issues/335

The `adaptive` prop has been removed from the Pager spec as well as the **adaptive** templates.

This prop was previously used to control the old adaptive pager behavior for showing and hiding certain elements. However, the new adaptive pager behavior now depends on the width of the pager and the available space. As a result, the previous method of implementing adaptiveness is no longer applicable.

Added test ( _pager-adaptive_ ) for all options along with their respective adaptive order for hiding elements.
Added test for opened page size dropdownlist in an actionsheet.


packages/html/src/pager/pager.spec.tsx
```diff
@@ -36,8 +37,9 @@ export type KendoPagerState = { [K in (typeof states)[number]]?: boolean };
 
 const defaultOptions = {
     size: Size.medium,
-    adaptive: false,
-    showPagerSizeInfo: true,
+    pagerSizeInfo: true,
```
TeyaVes, line 40: do we need the "Info" here? Wouldn't it be clearer if we leave these as "pagerSize, pageTitle, maxPages"?

tests/_output/bootstrap/pager/pager-adaptive.png
TeyaVes, line 1: I don't really get the 0-6 numbers here.. what do they stand for? Screen size/breakpoints or smth else?

TeyaVes (commented):
I see that we have large pager, but I don't see a small one.. Or the last pager in the pager-adaptive test is not with a sizing option lg, but smth else?

inikolova (approved):

epetrow (approved):

PR #5392, opened 2025-03-25
Add Actionsheet View

zhpenkov:
The `k-actionsheet-view` wraps the actionsheet header, content, and footer.

_Reference rendering:_
```html
<div class="k-actionsheet">
  <div class="k-actionsheet-view">
    <div class="k-actionsheet-titlebar"></div>
    <div class="k-actionsheet-content"></div>
    <div class="k-actionsheet-footer"></div>
  </div>
</div>

```


epetrow (commented):
Two things from me:

1. I think you should split this into two commits, since currently we have `chore(html)` but there are changes to the `scss` files there as well
2. Is the `template` prop needed?

epetrow (approved):

PR #5394, opened 2025-03-26
Dialog, Window and Action Buttons - update styles for M3

inikolova:
Design https://www.figma.com/design/2hhKcJiRYUTBZuAb72Sbk3/%5BMaterial-3%5D-Telerik-%26-Kendo-UI-Kit-for-Material?node-id=2371-63234&m=dev

In this PR:

- Window
- Dialog
- Action Buttons


packages/material/scss/window/_theme.scss
```diff
@@ -3,4 +3,9 @@
 
 @mixin kendo-window--theme() {
     @include kendo-window--theme-base();
+
+    .k-window-minimized {
```
TeyaVes, line 7: I think these rules should be added to the layout.scss file

packages/html/src/dialog/tests/dialog.tsx
```diff
@@ -50,7 +50,7 @@ export default () =>(
             </section>
 
             <section>
-                <DialogNormal title="Scrolling" actionButtonsAlign="end" style={{ height: "170px" }} >
+                <DialogNormal title="Scrolling" actionButtonsAlign="end" style={{ height: "220px" }} >
```
TeyaVes, line 53: I believe we should show scrollable content here and in Bootstrap this height removes the scrollable content. If we are increasing the height here, maybe we should add more content inside the dialog as well?

TeyaVes (commented):
in the designs, the action buttons in the color picker have differents horizontal and vertical margins, we need to double check that

epetrow (commented):
Just a couple of questions that I have:

1. Currently the way padding is applied to the ActionButtons seems a bit inconsistent to me. At first I thought that if the scenario is adaptive, we do not change the padding and it is desktop we apply it. However this does not seem to be the case. Is this intended or am I missing something?
2. Do we need to change the height of the Dialog in the scrollable test? Now the overflow is not visible in most of the themes.

inikolova:
> 1. . Is this intended or am I missing something?
We set padding to the action buttons indeed (in the current case it is k-spacing(6) set to elements with class **k-actions**). However the ActionSheet itself sets padding to **.k-adaptive-actionsheet k-actionsheet-footer** element, which is the same k-actions element - <div class="k-actions k-actions-stretched k-actions-horizontal k-actionsheet-footer">...... So the short answer here is yes, it is intended :)
>2. Do we need to change the height of the Dialog in the scrollable test? ... 
Yep, otherwise the Dialog content is not visible at all. I also checked the other themes and there is overflow

inikolova:
> in the designs, the action buttons in the color picker have differents horizontal and vertical margins, we need to double check that

I checked this with @imasyuko and confirm this is intended -> Material 3 provides guidelines for actions buttons inside popup (https://www.figma.com/design/DA406nWiiUc8BmMKb5EjUZ/Material-3-Design-Kit--Community-?node-id=51954-18122&t=CjOALw6aeipSGeTy-0) and inside Dialogs (https://www.figma.com/design/DA406nWiiUc8BmMKb5EjUZ/Material-3-Design-Kit--Community-?node-id=50723-10916&t=riBfTSOOp5owqKpU-0) -> and the paddings are different

TeyaVes (approved):

TeyaVes:
> Just a couple of questions that I have:
> 
> 1. Currently the way padding is applied to the ActionButtons seems a bit inconsistent to me. At first I thought that if the scenario is adaptive, we do not change the padding and it is desktop we apply it. However this does not seem to be the case. Is this intended or am I missing something?
> 2. Do we need to change the height of the Dialog in the scrollable test? Now the overflow is not visible in most of the themes.

@epetrow , on your second question, at first I thought so as well- that the overflow is not visible in most of the themes. This, however, is not the case, it's just that the cut is immediately after a sentence and it looks like there isn't any more content, but there in fact is- it's just not visible in the images where the scrollbar is not visible.

Issue #4981, opened 2024-04-16
Built-in support for add new item functionality in select-based components


yanisslav:
**Is your feature request related to a problem? Please describe.**
At present, we do not have a definitive rendering structure for the MultiSelect popup when the 'Allow custom values' feature is enabled. Although this feature is currently available in both [Kendo React](https://www.telerik.com/kendo-react-ui/components/dropdowns/multiselect/custom-values/#toc-primitive-values) and [Kendo Angular](https://www.telerik.com/kendo-angular-ui/components/dropdowns/multiselect/custom-values/#toc-primitive-values) suites, their implementations differ from each other, and the popup renders in a slightly different manner.

**Describe the solution you would like to see implemented**
This feature seems to have already been discussed, and there is a design - https://github.com/telerik/web-components-ux/issues/607#issuecomment-1463653343

yanisslav:
Rendering is also missing for other select-like components, such as the Combobox.

radkostanev:
Updating the scope of this, as we should ideally add support for this feature in all select-like components, not just one particular we have demand for.

inikolova:
Based on a research and an offline discussion with @dmanova and @dtopalov, it appears the _allowCustom_ functionality is available only for select elements with inputs - MultiSelect, MultiColumnComboBox, ComboBox and AutoComplete. Therefore I removed the labels for DropDownList, DropDownGrid, DropDownTree.

PR #5398, opened 2025-03-27
Rendering and styles for allowCustom values in select-based components

inikolova:
Design: https://www.figma.com/design/gWUIux95BHxemqWwOw9ZwB/Default-Theme---Internal-Usage?node-id=19786-1144574&t=QYICYdHPYzltgBc8-0

Closes:  https://github.com/telerik/kendo-themes/issues/4981

The suggested rendering for allowCustom values in this PR differs the current implementation in React and Angular. However it is derived from in-depth research and offline discussions with @dtopalov, @dmanova and @telerik/kendo-front-end. Details about the implemented behaviour could be found in[ this comment](https://github.com/telerik/web-components-ux/issues/607#issuecomment-2753991325)


packages/core/scss/components/list/_layout.scss
```diff
@@ -228,7 +234,8 @@
                 line-height: $_header-line-height;
             }
 
-            .k-list-item {
+            .k-list-item,
+            .k-list-custom-value {
```
epetrow, line 238: I don't think this line is needed

packages/fluent/scss/list/_layout.scss
```diff
@@ -227,7 +232,8 @@
                 --INTERNAL--kendo-list-header-line-height: var( --kendo-list-#{$size}-header-line-height, #{$_header-line-height} );
             }
 
-            .k-list-item {
+            .k-list-item,
+            .k-list-custom-value {
```
epetrow, line 236: Same here, I don't think we need to add `k-list-custom-value`

epetrow (approved):
Looks good, I am just wondering if we should change the default custom value. It makes sense for the scenario in the design, but in my opinion it does not fit the current tests.

PR #5400, opened 2025-03-28
Indicators - update styles for M3

inikolova:
Design: https://www.figma.com/design/2hhKcJiRYUTBZuAb72Sbk3/%5BMaterial-3%5D-Telerik-%26-Kendo-UI-Kit-for-Material?node-id=3668-154625&m=dev

In this PR:

- Notification
- Badge
- Loader
- Skeleton


tests/_output/material/loader/loader-container-colors.png
epetrow, line 1: If the light variant was hard to see before, now it has become barely visible. We are okay with that?

inikolova: > If the light variant was hard to see before, now it has become barely visible. We are okay with that?

The light color is intended to be used over dark surface, so in my opinion we are OK in this particular case. However, maybe we should consider changing the .k-loader-container-panel background in general (using a variable seems a good option to me)

TeyaVes (approved):

PR #5401, opened 2025-03-31
Add No-Data Test for Screen Readers

zhpenkov:
Related to: https://github.com/telerik/kendo-themes-private/issues/343

1. Ensure that the class for the "No Data" element is set consistently to `k-no-data`.
2. Add an example for `k-no-data` element always visible with `k-sr-only`.


Juveniel (approved):

TeyaVes (approved):
Looks great! I am just wondering why the Avatar image is generated

PR #5428, opened 2025-04-14
Make Non-Responsive Pager Wrap By Default

zhpenkov:
Add wrapping behavior for the non-responsive pager.

This will NOT affect the Responsive pager (old and new implementation), as there is no case where the items would overflow the container and wrap.


inikolova (approved):

epetrow (approved):
Looks good, I am not a fan of the commit message tho 😅

TeyaVes (approved):

PR #5431, opened 2025-04-15
Floating labels

TeyaVes:


packages/core/scss/components/floating-label/_variables.scss
```diff
@@ -12,10 +12,20 @@ $kendo-floating-label-focus-scale: null !default;
...
+
 $kendo-floating-label-transition: null !default;
 
 $kendo-floating-label-bg: null !default;
 $kendo-floating-label-text: null !default;
 
 $kendo-floating-label-focus-bg: null !default;
 $kendo-floating-label-focus-text: null !default;
+
+$kendo-floating-label-border-bg: k-color( surface-alt ) !default;
```
inikolova, line 31: I think in the core variable files we should keep only _null_ values

TeyaVes: yes, agree, I'll update this now

packages/html/src/floating-label/tests/floating-label-position.tsx
```diff
@@ -0,0 +1,175 @@
...
+const styles = `
+    .k-floating-label-container {
+        width: 100%;
+    }
+`;
+
+export default () =>(
+    <>
+        <style>{styles}</style>
+        <div id="test-area" className="k-d-grid k-grid-cols-5">
```
inikolova, line 13: The class here should be k-grid-cols-4, otherwise the test is misaligned

inikolova (approved):

epetrow (approved):
LGTM

inikolova (approved):

PR #5432, opened 2025-04-16
Correct Grid Adaptive Multi Checkboxes Filter

zhpenkov:
- Unscopes the `.k-multicheck-wrap` class from `k-popup` as this element is also used inside the actionsheet for adaptive scenarios. Those styles are required because they provide reset styles for the `li` elements
- Adds an adaptive test for the grid filter with multi checkboxes.


epetrow (commented):
Maybe we would want to delete the "legacy" adaptive tests such as `grid-adaptive-filter-multi`?

Also I wonder if this line should be moved to the Popup styles 🤔 

```css
    .k-popup .k-multicheck-wrap {
        max-height: 300px;
        overflow: auto;
    }
```

epetrow (approved):

inikolova (approved):

TeyaVes (approved):

PR #5433, opened 2025-04-16
Correct Column List Max Height

zhpenkov:
ColumnList applies `max-height: 200px` when used in ActionSheet, resulting in unexpected behavior:
image

ColumnList max-height styles should apply only in popups.

### DIFF
--- a/packages/core/scss/components/grid/_layout.scss
+++ b/packages/core/scss/components/grid/_layout.scss
@@ -1306,9 +1306,6 @@
         padding: 0;
         margin: 0;
         list-style: none;
-        max-height: 200px;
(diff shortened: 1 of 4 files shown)


packages/core/scss/components/grid/_layout.scss
```diff
@@ -1306,6 +1306,9 @@
         padding: 0;
         margin: 0;
         list-style: none;
+    }
+
+    .k-popup .k-column-list {
```
Juveniel, line 1311: avoid adding `k-popup` here

epetrow (approved):

inikolova (approved):

Issue #5399, opened 2025-03-27
[Grid] No distinction between locked and scrollable cells in Grid footer


dtopalov:
**Describe the bug**
When both locked columns and footers are enabled in the Grid, there is no distinction between the scrollable and locked areas similar to the ones available in the headers and content areas.

**To reproduce**
https://stackblitz.com/edit/angular-spy1hhp3?file=src%2Fapp%2Fapp.component.ts

**Expected behavior**
Same visual distinction in footer as in content and header, e.g.:

https://stackblitz.com/edit/angular-spy1hhp3-jnkd11dx?file=src%2Fapp%2Fapp.component.ts

**Screenshots**

![Image](https://github.com/user-attachments/assets/2544272b-2997-42b9-ae9e-32f5389c5293)

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material
* theme-tasks

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**
* All
* Not Applicable
* Firefox version: [for example, 72, lts]
* Chrome version: [for example, 80]
* Edge version: [for example, 80]
* Safari version: [for example, 13]
* IE version: [for example, 11]

**Build system information (please remove the unneeded items)**
* Not Applicable
* OS: [for example, Windows, Linux, Mac]
* Node version: [for example, 10.15.3]
* NPM version: [for example, 6.4.1]

**Additional context**
Add any other context about the bug.

PR #5435, opened 2025-04-17
fix(grid): no distinction between locked and scrollable cells in footer

inikolova:
Closes https://github.com/telerik/kendo-themes/issues/5399


epetrow (commented):
There are no changes for the Bootstrap theme.

image

epetrow (approved):

PR #5446, opened 2025-05-08
Material v3 fixes

TeyaVes:
- Gauge's track is with a wrong color


packages/core/scss/components/dataviz/_variables.scss
```diff
@@ -135,3 +135,7 @@ $kendo-chart-overlay-bg: null !default;
 $kendo-chart-no-data-font-size: null !default;
 $kendo-chart-no-data-font-weight: null !default;
 $kendo-chart-no-data-padding-y: null !default;
+
+// Gauge
+$kendo-gauge-pointer: null !default;
```
inikolova, line 140: As these variables are used for colouring, maybe we could add **_-bg_** suffix to the variables names?

inikolova (approved):

PR #5448, opened 2025-05-09
fix(color-gradient): revert slider and hsv rectangle fixed height

TeyaVes:
fix for: https://www.telerik.com/kendo-angular-ui-develop/components/inputs/colorgradient


epetrow (commented):
- I noticed that there are differences in the values of `$kendo-color-gradient-canvas-rectangle-height` and `$kendo-color-gradient-slider-vertical-size`, for example

before:
image
now:
image

I would expect that this is needed? Some of the values seem a bit odd and I was wondering if there is a design for this

- Also most likely this change is required now, but did we check if it would not break the Angular implementation

```scss
.k-colorgradient-canvas {
  .k-hsv-gradient {
     aspect-ratio: 1;
     height: 100%;
  }
}
```

TeyaVes:
> * I noticed that there are differences in the values of `$kendo-color-gradient-canvas-rectangle-height` and `$kendo-color-gradient-slider-vertical-size`, for example
> 
> before: image now: image
> 
> I would expect that this is needed? Some of the values seem a bit odd and I was wondering if there is a design for this
> 
> * Also most likely this change is required now, but did we check if it would not break the Angular implementation
> 
> ```css-scss
> .k-colorgradient-canvas {
>   .k-hsv-gradient {
>      aspect-ratio: 1;
>      height: 100%;
>   }
> }
> ```

@epetrow , these changes are expected and yes, the values are a bit odd, because the gradient needs to be a square and with these widths, we are ensuring that it is a square. The suggested styles are as it was previously implemented, but this caused a bug in Angular, which can be observed here: https://www.telerik.com/kendo-angular-ui-develop/components/inputs/colorgradient .

epetrow (approved):

inikolova (approved):

Issue #5447, opened 2025-05-09
[ButtonGroup] Impossible to tell if focused button is selected visually


austinw-fineart:
**Describe the bug**
The ButtonGroup component gives no visual indication of whether a focused button is selected or not using the Material theme.

**To reproduce**

1. Go to https://www.telerik.com/kendo-angular-ui/components/buttons/buttongroup
2. Change theme to Material
3. Click button group
4. Click button group again

**Expected behavior**
Button selected state should clearly be visible even while button is focused

**Screenshots**
Is this button selected? You tell me.
![Image](https://github.com/user-attachments/assets/cf379f49-b1b4-4630-a334-72b052ea0a30)

**Affected package (please remove the unneeded items)**
* theme-material

**Affected suites (please remove the unneeded items)**
* Kendo UI for jQuery
* Kendo UI for Angular
* Kendo UI for React
* Kendo UI for Vue
* Telerik UI for Blazor

**Affected browsers (please remove the unneeded items)**
* All

**Build system information (please remove the unneeded items)**
* Not Applicable

**Additional context**
Add any other context about the bug.

inikolova:
According to the [Material 3 guidelines](https://www.figma.com/community/file/1035203688168086460), the buttons in focus and pressed states share the same background variables (primary + on-primary, 12%). In our Kendo theme inspired by Material 3, these complex colors are converted into solid colors (using steps from the accent color palette), so these two states share the same background color. The focus indication successfully differentiates the focused button from the selected or active one, but it does not distinguish the button in focus state from a selected focused button (which is more complex and does not have its own styles not only in the Kendo theme, but in [Material](https://m3.material.io/) too.

austinw-fineart:
It's true that the initial M3 spec did not specify a selected state or color. That's since been addressed by M3 Expressive in the now updated page on [buttons](https://m3.material.io/components/buttons/specs). That said, ButtonGroup's Material counterpart should've been based off of [segmented buttons](https://m3.material.io/components/segmented-buttons/overview) instead (deprecated by M3 Expressive). Its selected state changes the button's color directly and thus its state colors as well.

[...]

PR #5471, opened 2025-05-28
fix(button): focus and selected states lack clear visual differentiation

inikolova:
Related to https://github.com/telerik/kendo-themes/issues/5447

Based on offline discussions between the Front-end and Design teams, we agreed to introduce a clearer distinction between the focus and selected states of the Button component. This decision is intended to enhance accessibility and improve the visual clarity, even though it deviates from the [Material 3 guidelines](https://www.figma.com/community/file/1035203688168086460).

**Note:** According to the M3 guidelines, the focus and pressed states of buttons utilize the same background colors. With the changes in this PR, the focus state of the button will now use the background of the normal state, without applying an additional background color.


tests/_output/material/button/button-solid.png
epetrow, line 1: This might be expected, but the hover state of the light variant is barely visible now

epetrow (commented):
Also, please rebase this branch, so the Bootstrap related images are removed.

tests/_output/material/button/button-solid.png
inikolova: I guess you mean the _focus_ state? And yes, this is synced with @imasyuko -> the light button should be used over dark surfaces

tests/_output/material/button/button-rtl.png
TeyaVes, line 1: We can also add a scenario for Selected + Focused in the following test.. In this way, we'll also check the selected + focused state of the primary button

PR #5479, opened 2025-06-06
Responsive forms

TeyaVes:
targets: https://github.com/telerik/kendo-themes-private/issues/386
POC: https://github.com/telerik/kendo-themes-private/issues/296
Suggested rendering: https://stackblitz.com/edit/stackblitz-starters-v5tuzw?description=HTML/CSS/JS%20Starter&file=script.js,styles.css,index.html&terminalHeight=10&title=Static%20Starter


packages/html/src/form/form.spec.tsx
```diff
@@ -31,6 +31,28 @@ const defaultOptions = {
     tag: 'form'
 };
 
+function getGridClasses(prefix: string, value?: number | Array<Record<string, number>>): string[] {
```
zhpenkov: We might extract this function, as it is used repeatedly in three files.

epetrow (commented):
I see that the bundle size has increased quite a lot with the introduction of the utilities. With how the responsive utils are currently implemented, I don't think there are a lot of options, but never the less it is something to consider.

PR #5499, opened 2025-06-26
Update Prompt Components

epetrow:
closes: https://github.com/telerik/kendo-themes-private/issues/388
closes: https://github.com/telerik/kendo-themes-private/issues/371


packages/html/src/prompt/prompt-send-button.tsx
```diff
@@ -0,0 +1,50 @@
...
+    } = props;
+
+    return (
+        <Button
+            {...other}
+            className={classNames(
+                props.className,
+                PROMPT_SEND_BUTTON_CLASS,
+                {
+                    ['k-retrieving']: generating,
```
inikolova: Maybe we should change the class name here to _k-generating_?

packages/bootstrap/scss/prompt/_variables.scss
```diff
@@ -67,6 +67,13 @@ $kendo-prompt-suggestion-border: k-color(border) !default;
 /// @group prompt
 $kendo-prompt-suggestion-shadow: k-elevation(1) !default;
 
+// Inline prompt
+/// The vertical padding of the inline AIPrompt content.
```
inikolova: Do we need the AI before Prompt?

packages/html/src/prompt/prompt-stop-fab.tsx
```diff
@@ -0,0 +1,54 @@
...
+    return (
+        <FloatingActionButton
+            {...other}
+            className={classNames(
+                props.className,
+                PROMPT_STOP_FAB_CLASS,
+                {
+                    ['k-generating']: generating,
+                },
+                'k-z-10'
```
inikolova: I would suggest to set the z-index value via css, instead of using utility class

packages/html/src/prompt/prompt-stop-fab.tsx
```diff
@@ -0,0 +1,54 @@
...
+            )}
+            style={{ bottom: "16px", right: "16px"}}
```
inikolova, line 39: Could we set these values inside the scss files?

PR #5513, opened 2025-07-14
Rtl styles

TeyaVes:
targets: https://github.com/telerik/kendo-themes-private/issues/358

Things to be done:

- [x] Convert directional properties to logical properties in all themes
- [x] Make sure that the rtl styles in the tests look okay
- [x] Remove the RTL styles where possible
- [ ] Try creating a mixin that handles the transition and gradient in RTL
- [ ] [Optional] add linting rule and make sure that we are using directional properties only when needed- will do this in a separate PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,10 +139,17 @@ jobs:
       - name: Done
         run: echo "Done!"
 
-  ci-a11y:
-    name: Status check > A11y
+  a11y:
+    name: A11y
     needs: [ compile-themes, render-test-pages ]
     uses: ./.github/workflows/_test-contrast.yml
     with:
       max-chunks: 4
 
+  ci-a11y:
+    name: Status check > A11y
+    runs-on: ubuntu-latest
+    needs: [ a11y ]
+    steps:
+      - name: Done
+        run: echo "Done!"


tests/_output/material/scheduler/scheduler-month-view.png
inikolova, line 1: Compared to the LTR styles, maybe we should double check this change -> https://github.com/telerik/kendo-themes/blob/develop/tests/_output/material/scheduler/scheduler-month-view.png

packages/core/scss/components/slider/_layout.scss
```diff
@@ -176,7 +176,7 @@
 
         .k-tick {
             text-align: end;
-            margin-inline-start: 2px;
+            margin-left: 2px;
```
inikolova: Is this change intended?

packages/fluent/scss/slider/_layout.scss
```diff
@@ -8,23 +8,12 @@
...
-        width: min-content;
-        height: min-content;
         gap: $kendo-slider-spacing;
 
         .k-label {
             font-size: .875em;
             pointer-events: all;
         }
-
-        &[disabled],
```
inikolova, line 20: Is this change intended?

packages/core/scss/components/slider/_layout.scss
TeyaVes: yes, it's expected- in the first commit, I have replaced the margin-left to margin-inline-start, but then I have reverted it, because it was breaking the styles.

packages/fluent/scss/slider/_layout.scss
TeyaVes: yes, this style is no longer needed as we are having this in the core folder

Issue #5427, opened 2025-04-14
[Gantt][Tooltip] Gantt tooltip renders behind the toolbar


dtopalov:
**Describe the bug**
Due to the React Gantt rendering specifics, the Toolbar element receives a `k-gantt-header` class which in turn sets `z-index` to 6. This happens to exceed the z-index value of the tooltip popup container (at least in Angular):

https://stackblitz.com/edit/angular-frwbmnku-sb2smzrr?file=src%2Fapp%2Fapp.component.ts

**To reproduce**

Open the stakblitz demo and hover the first task.

**Expected behavior**
Provide some Gantt specific styling for popups containers for the built-in tooltips (e.g. .k-animation-container.k-gantt-tooltip) with high enough z-index.

**Screenshots**

![Image](https://github.com/user-attachments/assets/5f5d3194-496b-4138-80d2-c0c58b1447bb)

Issue #5397, opened 2025-03-26
k-dock-manager-unpinned-container does not always cover the content under it


ntacheva:
**Describe the bug**

 The `<div class="k-dock-manager-unpinned-container">` has z-index of 20. As a result, it does not cover the content of the pane underneath it if it has higher z-index.

**To reproduce**

1. Run this sample: https://blazorrepl.telerik.com/GpOHcpPz30gmOMIR22
1. Unpin the "Properties" tab
1. Select the "Properties" tab so it pops over -> the  "Properties" pane does not cover the content of the "Sequence" pane because the Treeview's animation contaier inside it has z-index of 100.

**Expected behavior**

The unpinned pane is always rendered on top and covers the content underneath it.

**Screenshots**

![Image](https://github.com/user-attachments/assets/e273fe6c-328a-48ec-8e94-33efb100269c)

**Affected package (please remove the unneeded items)**
* theme-default
* theme-bootstrap
* theme-material
* theme-tasks

**Affected suites (please remove the unneeded items)**

All suites that have a DockManager component.

**Affected browsers (please remove the unneeded items)**
* All

**Build system information (please remove the unneeded items)**
* Not Applicable

Issue #5025, opened 2024-05-08
Badge component z-index 9999


mbechev:
`.k-badge-edge` class has a **z-index** property set to 9999. Does it need such a big **z-index**?

```css
.k-badge-edge {
    position: absolute;
    z-index: 9999;
}
```

**Background of the discussion**
We tried using the badge inside a Grid cell, but opening a popup appears behind the badge.
<img width="231" alt="image" src="https://github.com/telerik/katana/assets/33104481/098d0894-6f16-4b4c-a210-39f10b8ba00d">
https://stackblitz.com/edit/angular-ramb33?file=src%2Fapp%2Fapp.component.ts,src%2Fapp%2Fapp.module.ts

PR #5514, opened 2025-07-15
Revise z index

inikolova:
Closes https://github.com/telerik/kendo-themes-private/issues/366

This PR addresses the following issues as well:

https://github.com/telerik/kendo-themes/issues/5427
https://github.com/telerik/kendo-themes/issues/5397
https://github.com/telerik/kendo-themes/issues/5025

Based on an audit of z-index usage across all themes and components, I would suggest the following organisation into semantic groups with clearly defined and meaningful ranges:
**1. Base / background (-1-0)**
    For elements that should appear behind normal content (pseudo-elements, placeholders, backgrounds);
**2. Content / basic (1-99)**
    For normal content elements and their internal layering - like grid headers, calendar header, form inputs etc.
**3. Navigation (100-999)**
    For elements that should appear above normal content (for example Sticky Appbar);
**4. Floating (1000-9999)**
    For elements that float above the main interface (Notification, Badge, Dock manager unpinned containers, Map attribution, Fixed Appbar);
**5. Overlay (10000-11999)**
    For modal dialogs and overlays, Drawer overlays, Fullscreen modes.
**6. Popups (12000 - 19999)**
    For Tooltip, Popover, Context menu, Chat tooltip.
**7. User feedback (20000-29999)**
    For drag&drop elements elements, loading states, and temporary user feedback.
**8 Special Cases (30000+)**
    For emergency overrides and special cases (!!!use sparingly).

 Here are some further suggestions based on the research:
    1. Structural Elements (for example k-header) which are direct children of a component should have z-index 1;
    2. **k-resize-handle** -> its z-index has been increased to 10, as it should always appear above its surrounding elements.; this value is now consistent across all components;
    3. **Interactive states** - when component state needs a higher priority, increase its z-index with 1;  
    4. **.k-edit-cell** should have the highest priority than all other cells - normal, sticky, intersection of sticky row and cell, therefore is z-index is increased to 5.


tests/_output/classic/spreadsheet/spreadsheet-jquery.png
zhpenkov, line 1: Is this change expected?
Or is this test legacy and no longer relevant?

inikolova: This test is legacy, therefore I removed it

TeyaVes:
Maybe we can add a list with the componets that use z-index and order them by z-index

PR #5523, opened 2025-07-28
chore(html): remove prompt suffix separator

zhpenkov:


epetrow (approved):
LGTM, only thing that is weird is why the preview components differences are generated only for the fluent theme.

Issue #5354, opened 2025-02-28
Slider Tooltip Callout Position bug


TeyaVes:
**Describe the bug**
When calculating the position of the slider tooltip callout position, we are using the following styles:

```
 // Slider tooltip
    .k-slider-tooltip {
        .k-callout-n,
        .k-callout-s {
            margin-inline-start: - list.slash( $kendo-tooltip-callout-size, 2 );
        }

        .k-callout-w,
        .k-callout-e {
            margin-top: - list.slash( $kendo-tooltip-callout-size, 2 );
        }
    }
```

However, when compiled, the mentioned styles produce invalid CSS:

```
.k-slider-tooltip .k-callout-n, .k-slider-tooltip .k-callout-s {
    margin-inline-start: -6px / 2;
}
```

On the other hand, it seems that the callout is positioned correctly despite the invalid CSS, meaning that we need to check if the styles are needed at all.

PR #5567, opened 2025-08-28
chore(slider): remove reduntant styles

TeyaVes:
closes: https://github.com/telerik/kendo-themes/issues/5354

The styles for the slider tooltip callout come from the Tooltip component, and the ones coming from the slider are redundant (and in fact compile to invalid CSS)


inikolova (approved):

epetrow (commented):
I wonder if this should be a fix since the issue we are closing has been categorised as a bug. Also as far as I am aware, milestones should be set on the issue that the pr is addressing.

zhpenkov (approved):

Issue #3818, opened 2022-08-11
Move the ::before element inside the k-list-item-text to a different place


gdenchevprog:
**Is your feature request related to a problem? Please describe.**

With the new ListBox rendering, there's extra space above each item when the content of the item is surrounded in \<p> tags. 

Run the following Dojo example to observe the described behavior:

https://dojo.telerik.com/ifjeaITx

Screenshot:
![image](https://user-images.githubusercontent.com/70957719/184092799-b2e6e3b0-7d21-4887-a472-8df823b8bc5e.png)

It seems like the extra space comes from the ::before element inside the k-list-item-text span.

![image](https://user-images.githubusercontent.com/70957719/184093025-272eeb9c-273f-4587-a4d8-0f4b473183b4.png)

**Describe the solution you would like to see implemented**

The proposal is to see if it's possible to move the ::before element to a different place or remove it altogether.

gdenchevprog:
Mentioned again in the following forum thread:
https://www.telerik.com/forums/listbox-item-template#5627782

martintabakov:
Also reported in  ticket #1613362

inikolova:
The mentioned **:before** element was added in order to address a problem with wrong height of empty list items: 
Issue: https://github.com/telerik/kendo-themes/issues/3388
Fix: https://github.com/telerik/kendo-themes/commit/ebd716e8ca576719b6d9a574676e0d1d4e7c6c3c

With the current list implementation there are not many options to handle all scenarios - default item, empty item, item with template, empty item with template with CSS only. After a deep investigation, we believe the "correct" direction is to use the new [**_lh_** ](https://www.w3.org/TR/css-values-4/#lh) unit, however it is not supported in all major browsers yet. Therefore we will postpone the fix for the current issue until the **_lh_** unit has full support.

epetrow:
Fixed with https://github.com/telerik/kendo-themes/pull/5897

PR #5569, opened 2025-08-29
List items height

inikolova:
Closes https://github.com/telerik/kendo-themes/issues/3818


TeyaVes (approved):

packages/html/src/list/tests/list.tsx
zhpenkov, line 1: Just a suggestion.
We can add this item either first or last, as this edge case seems somewhat unusual between the states.

Juveniel:
we should check if this works correctly in virtualized scenarios. We have actually tried this fix before, but hit some limitations/scenarios where the `lh` height failed. Please, review some actual component demos with virtualization/advanced scenarios to verify that this works correctly.

packages/html/src/list/tests/list.tsx
inikolova: I already tried that, but since we are targeting an empty item, it is not visible at all - whether it is the first or the last one. That's why I placed it between states, where a background is present.

PR #5570, opened 2025-08-29
fix(fluent): empty cells should not be interactive

epetrow:
In the Fluent theme empty calendar cells can be hovered, focused, etc which is not the case in the other themes.

### DIFF
--- a/packages/fluent/scss/calendar/_layout.scss
+++ b/packages/fluent/scss/calendar/_layout.scss
@@ -109,6 +109,10 @@
         display: none;
     }
 
+    .k-calendar-td.k-empty {
+        pointer-events: none;
+    }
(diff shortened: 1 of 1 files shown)


packages/fluent/scss/calendar/_layout.scss
```diff
@@ -109,6 +109,11 @@
         display: none;
     }
 
+    .k-calendar-td.k-empty {
+        pointer-events: none;
+        user-select: none;
```
inikolova: I think we don't need **user-select: none** here, as it is inherited from _.k-calendar_

inikolova (approved):

zhpenkov (approved):

Issue #5026, opened 2024-05-08
Filtering icon of DropDowns doesn't have padding


tswetti:
**Describe the bug**
Initially reported here - https://github.com/telerik/kendo-angular/issues/4294

The filtering icon of the DropDownTree and MultiSelectTree does not have padding in Kendo Angular due to the icon wrapper:
![image](https://github.com/telerik/kendo-themes/assets/65012268/3662bf39-b8ed-4f45-b183-066ae730673e)
The DropDownList has padding as expected.

The cause of the issue could be either:
- The CSS selector that applies the padding doesn't take into consideration `.k-icon-wrapper-host`
- Both DropDowns lack the `.k-input-icon` class that is present in the DropDownList

**To reproduce**
https://www.telerik.com/kendo-angular-ui/components/dropdowns/multiselecttree/filtering/#toc-built-in-filtering

**Expected behavior**
The filtering icon to have padding:
![image](https://github.com/telerik/kendo-themes/assets/65012268/a5656664-0161-4a82-bd65-5d1784f8e913)

**Affected suites**
* Kendo UI for Angular

PR #5573, opened 2025-09-01
fix(input): adornments icon padding

zhpenkov:
closes: https://github.com/telerik/kendo-themes/issues/5026

The bug is caused because the Angular rendering includes an extra element wrapping the icon:
`k-input-prefix > .k-icon-wrapper-host .k-icon`

cc: @svetq


TeyaVes (approved):
Approving this, although having so many selectors does not seem ideal.. however, currently I cannot think of a better approach either

PR #5587, opened 2025-09-10
Fix Sass Deprecation Warnings

zhpenkov:
fixes: https://github.com/telerik/kendo-themes/issues/5621

## Overview
With `1.91.0` Sass version, using the underscore prefix (`_`) to mark variables as private (e.g., `$_default-{name}`) now generates warnings when these maps are overridden in theme files.

**Screenshot:**
image

## Root Cause
Dart Sass has changed how it handles private variable overrides, making the previous pattern of using `$_default-{map}` problematic.

## Implemented Solution
Updated default maps to remove the underscore prefix, making them public variables:
- `$_default-{map}` → `$default-{map}`
- This eliminates the override warnings
- Maintains the same functionality for theme customization


epetrow (commented):
Running `sass:dist` now takes about 2-3 minutes, while on `develop` it takes ~20 seconds. I don't know if there is anything we can do about it, but I think we should investigate.

epetrow (commented):
We also have a bunch of private variables in the utils package, what do we do about those?

PR #5607, opened 2025-09-25
Chat component enhancements v2

inikolova:
…the text

**In this PR:**
_**New features:**_
	- Dynamic message border radius (test **chat-messages-border-radius**);
	- Chat files new layout options - scroll and wrapping (test **chat-files-layout**)
	- Chat events - markers that serve to inform participants about important occurrence related to the conversation or chat UI (test **chat**);
	- User status (test **chat-user-status**). **_Note_**: by default k-chat-user-status does not have any appearance-related styles. It's an empty container with position, size and border-radius defined.
	- Empty chat templates - (test **chat-empty**). By default k-message-list-content-empty does not have any appearance-related styles. It's an empty container with layout and sizing.

**_Changes / enhancements_**
	- Update message rendering (previously incorrect) for scenarios with skeleton and message-reference;
	- Chat header use themeColor _**inherit**_ now (previously it was light);
	- Files in textarea -> previously they were wrapping, now scrolling behaviour is applied;
	- Files size -> now all files have min-width: 96px and max-width: 160px (previously there was fixed with 240px);
	- Files background and border colors (previously they were transparent);
	- Deleted file background and border (previously it was error color and it matches the colors of normal attachments);
	- Full width messages -> message bubbles should be as wide as their content is;

Design issue -> https://github.com/telerik/web-components-ux/issues/1403#issuecomment-3348819904

Design files -> https://www.figma.com/design/cTPYz7xGzVWYQieZPBkCto/Conversational-UI?node-id=8357-73849&m=dev


packages/html/src/chat/chat-files-wrapper.tsx
```diff
@@ -1,17 +1,43 @@
...
+                }
+            )}
+        >
+            {props.children}
+        </ul>
+    )
+};
+
+// ChatFileWrapper.states = states;
+// ChatFileWrapper.options = options;
```
zhpenkov: _nitpick_
Remove the comments or add the appropriate states and options.

packages/html/src/chat/chat-message.tsx
```diff
@@ -87,7 +89,16 @@ export const ChatMessage = (
...
+                content={
+                    <>
+                        {additionalContent}
+                        {text &&
+                            <span className="k-chat-bubble-text">
+                                {text}
+                            </span>
+                        }
+                    </>
+                }
```
zhpenkov, line 101: This seems quite complex. Why not have a single `content` instead of `text`, `additionalContent` properties and pass the JSX elements directly into it, depending on the use-case scenario?

packages/html/src/chat/chat.spec.tsx
```diff
@@ -59,8 +61,10 @@ export const Chat: KendoComponent<KendoChatProps & React.HTMLAttributes<HTMLDivE
             )} dir={dir}>
             {header}
             {pinned}
-            <div className="k-message-list k-avatars">
-                <div className="k-message-list-content">
+            <div className="k-message-list">
+                <div className={classNames("k-message-list-content",
+                        empty && "k-message-list-content-empty"
+                )}>
```
zhpenkov, line 67: To avoid adding an extra `empty` prop, we can make the message list content apply `k-message-list-content-empty` if there are no children.

tests/_output/bootstrap/chat/chat-empty.png
zhpenkov, line 1: Might remove notes from visual regression tests.

tests/_output/bootstrap/chat/chat-files.png
zhpenkov, line 1: The error state in files differs a lot. Is that expected?

tests/_output/bootstrap/chat/chat-user-status.png
zhpenkov, line 1: Might remove notes.

tests/_output/bootstrap/chat/chat-files.png
inikolova: Yep, this change is intended. Here is the design file -> https://www.figma.com/design/cTPYz7xGzVWYQieZPBkCto/Conversational-UI?node-id=8845-28621&m=dev

tests/_output/bootstrap/chat/chat-user-status.png
inikolova: I prefer to keep them here, to avoid confusion.

inikolova: After offline discussion, we agreed to remove them

packages/html/src/chat/chat-message.tsx
inikolova: After offline discussion, we agreed to rename the _additionalContent_ prop to _content_ (it is used for skeleton, loading and message reference) and keep the _text_ prop as well.

packages/html/src/chat/chat.spec.tsx
inikolova: In offline discussion, we agreed to proceed with the initial approach

tests/_output/default/chat/chat-files-layout.png
TeyaVes, line 1: by design, the Download all button should be aligned on the left with the attachments, but it doesn't seem so in the actual implementation because of the inline-padding added to the k-chat-download-button-wrapper

inikolova: Removed!

tests/_output/bootstrap/chat/chat-files.png
TeyaVes: I guess this has been discussed, but for me personally, the change is a bit weird- there is no distinction between a deleted file and a non-deleted one. I think that at least it should look like a disabled one with opacity..

imasyuko: Hey guys, I know the change is significant, but the previous styling for deleted attachments was based on the decision that attachments are positioned outside of the message bubbles. Later, another decision was made indicating that all attachments — regardless of whether they are accompanied by a message — are part of a message bubble, which requires different styling options for deleted attachments. 
The colors from the error swatch do not look appropriate within a message bubble that uses primary color variables across any theme. It is also not necessary to emphasize deleted attachments; using different icons, messaging, and remove actions is sufficient.

tests/_output/bootstrap/suggestion/suggestion-group.png
epetrow, line 1: Do we need a separate test for this if we are showing the behaviour in `chat-suggested-actions`?

packages/html/src/suggestion/suggestion-group.spec.tsx
```diff
@@ -12,10 +12,12 @@ const options = {};
 export type KendoSuggestionGroupOptions = {};
 
 export type KendoSuggestionGroupProps = KendoSuggestionGroupOptions & {
+    scrollButtons?: boolean;
     scrollable?: boolean;
 };
 
 const defaultOptions = {
+    scrollButtons: false,
     scrollable: true
```
zhpenkov: _nitpick_

Is there a case where:

**scrollButtons: true
scrollable: false**

If there is no such case, we might remove scrollButtons and change the scrollable prop to:

**scrollable: "normal" | "buttons"**

tests/_output/material/chat/chat-files.png
zhpenkov, line 1: Do we have a test with chat files and long messages? (e.g, the message in length is larger than the files below)
Does the message expand the bubble or overflows?

zhpenkov (approved):

packages/html/src/suggestion/suggestion-group.spec.tsx
inikolova: Actually when scrollButtons are true, scrollable is always false. This approach is being used because of the jQuery suite, where no changes in the Chat component are planned for this release.
A template for this scenario is added:
https://github.com/telerik/kendo-themes/pull/5607/commits/76e3dbdacaf552615f841052098cb3b2c068b484#diff-d94868a02a471f7a6517578d808fb12cefcaa9140d1483449ac051b2232bf62eR4

tests/_output/material/chat/chat-files.png
inikolova: No don't have a specific test for message with long length and a file. However, we have multiple tests for different message variants - chat-messages, chat-full-width-messages, chat-expandable-messages etc. where the message length is different. Having or not attachment does not change the behaviour of a message

tests/_output/bootstrap/suggestion/suggestion-group.png
inikolova: This test was added in the previous release, as the suggestions were extracted as a separate component. So, in my opinion we should keep this test

PR #5651, opened 2025-10-30
[html] remove k-item from k-tabstrip-item element

zhpenkov:
This PR adds support for `.k-tabstrip-item` as the primary styling class for TabStrip items, while maintaining backward compatibility with the existing `.k-item` class.

### Changes
- Added `.k-tabstrip-item` selectors alongside existing `.k-item` selectors.
- Both classes are currently supported.

### Migration Path
Currently, both `.k-item` and `.k-tabstrip-item` classes are supported. Since all suites currently use both classes, we will be removing `.k-item` from the reference rendering in the HTML package.

### DIFF
--- a/tests/_preview/preview-components.html
+++ b/tests/_preview/preview-components.html
@@ -187,7 +187,7 @@
                                 <div class="k-tabstrip k-tabstrip-top k-tabstrip-md">
                                     <div class="k-tabstrip-items-wrapper k-hstack">
                                         <ul class="k-tabstrip-items k-reset k-tabstrip-items-start">
-                                            <li class="k-item k-tabstrip-item k-first">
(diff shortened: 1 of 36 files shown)


epetrow (approved):

packages/core/scss/components/tabstrip/_layout.scss
```diff
@@ -421,28 +437,42 @@
                 inset-inline-start: 0;
                 pointer-events: none;
             }
+            .k-tabstrip-item.k-active::after,
             .k-item.k-active::after {
                 display: block;
             }
 
+            .k-tabstrip-item,
```
TeyaVes: shouldn't we add .k-tabstrip-item > .k-link here :?

TeyaVes (approved):

PR #5649, opened 2025-10-29
Infra Cleanup

zhpenkov:
Infrastructure cleanup focused on file naming conventions and code organization.

### Main Entry Points
- **Renamed `index.import.scss` → `import.scss`** across all packages (core, themes, utils)

### Function Files
- **Renamed `*.import.scss` → `*.scss`** for all function files in `packages/core/scss/functions/`:
  - `_color.import.scss` → `_color.scss`
  - `_lang.import.scss` → `_lang.scss`
  - `_list.import.scss` → `_list.scss`
  - `_math.import.scss` → `_math.scss`
  - `_map.import.scss` → `_map.scss`
  - `_meta.import.scss` → `_meta.scss`
  - `_string.import.scss` → `_string.scss`

### Removed Unused Imports
- **Removed unused function imports** from component variable files

### Simplified Function Implementations
- **Removed wrapper functions** that only called Sass built-in functions:
  - `k-color-alpha` (use `color.alpha()` directly)
  - `k-list-*` functions (use `list.*` directly)
  - `k-map-*` functions (use `map.*` directly)
  - `k-math-*` functions (use `math.*` directly)
  - `k-string-*` functions (use `string.*` directly)
  - `k-meta-*` functions (use `meta.*` directly)
- **Kept only custom utility functions** that add value beyond Sass built-ins
- **Updated function calls** to use native Sass modules directly where appropriate

### Updated Import Statements
- **Updated all import statements** to reference new file paths

## Impact

- ✅ No functional changes - purely organizational
- ✅ Cleaner, more consistent file naming
- ✅ Reduced code duplication

PR #5657, opened 2025-11-03
Add Missing Component Dependencies

zhpenkov:
Separated this commit from https://github.com/telerik/kendo-themes/pull/5649, as the component order in the dist needs to be investigated further and retained.


Juveniel:
@zhpenkov is this still relevant ? If not, please close it.

PR #5673, opened 2025-11-11
Focus streamline

inikolova:
Design -> https://www.figma.com/design/Oiuw9rwVWppmEK76zTYTqw/Focus-Indication?node-id=4096-241678&m=dev


Juveniel:
Postponing this for the time being. We can revisit again if needed at a later strage.

PR #5679, opened 2025-11-14
feat(list): add rendering for accessible list

inikolova:
In this PR:

- add spec, templates and tests for accessible grouped list (i.e. each group from the list is in separate `<ul>` element);

- remove **list-angular.spec** and the respective templates / tests, as this rendering is not used any more. Currently the Angular suite uses the accessible list rendering

Closes https://github.com/telerik/kendo-themes-private/issues/432
Closes https://github.com/telerik/kendo-themes-private/issues/431

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3244,9 +3244,9 @@
       }
     },
     "node_modules/@istanbuljs/load-nyc-config/node_modules/js-yaml": {
-      "version": "3.14.1",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-3.14.1.tgz",
-      "integrity": "sha512-okMH7OXXJ7YrN9Ok3/SXrnu4iX9yOk+25nqX4imS2npuvTYDmo/QEZoqwZkYaIDk3jVvBOTOIEgEhaLOynBS9g==",
+      "version": "3.14.2",
+      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-3.14.2.tgz",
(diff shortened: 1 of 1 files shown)


packages/html/src/dropdownlist/templates/dropdownlist-grouping-accessible.tsx
```diff
@@ -0,0 +1,24 @@
...
+        popup={
+            <ListAccessible>
+                <ListGroup label={"Group 1"} root>
+                    <ListItem>List group 1 item 1</ListItem>
+                    <ListItem>List group 1 item 2</ListItem>
+                    <ListItem>List group 1 item 3</ListItem>
+                </ListGroup>
+                <ListGroup>
+                    <ListItem group>Group 2</ListItem>
+                    <ListItem>List group 1 item 1</ListItem>
```
TeyaVes, line 16: I am a bit confused here. Shouldn't it be List group 2 item 1?

packages/html/src/dropdownlist/tests/dropdownlist-grouping.tsx
```diff
@@ -25,15 +26,151 @@ export default () => (
...
+                    popup={
+                        <List>
+                            <ListGroup label={"Group 1"}>
+                                <ListItem>List group 1 item 1</ListItem>
+                                <ListItem>List group 1 item 2</ListItem>
+                                <ListItem>List group 1 item 3</ListItem>
+                            </ListGroup>
+                            <ListGroup label={"Group 2"}>
+                                <ListItem>List group 1 item 1</ListItem>
+                                <ListItem>List group 1 item 2</ListItem>
```
TeyaVes, line 41: same as above

PR #5690, opened 2025-12-03
feat(html): add segmentedControl and smartToolbox components; enhance list rendering

inikolova:
~~The design of the new component indicates that changes to the current list rendering will be required to achieve the desired outcome. The following two branches outline two possible approaches for implementing these modifications.~~

~~_- Without adding an additional wrapper element for the text + icon:_
https://github.com/telerik/kendo-themes/commits/list-rendering-update/~~

~~_- - With adding an additional wrapper element for the text + icon:_
https://github.com/telerik/kendo-themes/tree/list-additional-wrapper~~

This PR suggests rendering and styling for the following new components:

**- Segmented Control**
Design -> https://www.figma.com/design/iPaMpXnn7gf6nU4esvImtX/Segmented-Control?node-id=0-1&p=f&m=dev
The Segmented Control is composed of multiple buttons arranged side by side, allowing users to select a single option. It includes a thumb indicator that:
  -- Clearly highlights the selected segment;
  -- Has fluid width, depending on the selected button content;
  -- Animates smoothly when the selection changes.

**- Smart Toolbox (name TBD)**
Design -> https://www.figma.com/design/57JL8WfKZV2nsRDZWwX0U7/Prompt-Controlled-DataGrid---1398-?node-id=1952-9565&m=dev
The Smart Toolbox is an enhanced input component with refined styling and an associated popup. Key characteristics:
  -- Improved visual styling;
  -- A popup with offset positioning set through the component API.

**- List Component Updates**
Design -> https://www.figma.com/design/0bezQDg0DS4eo8f7B6bXlv/List?node-id=8-180&p=f&m=dev
This PR also introduces updated rendering and styling for the List component, including:
  -- Support for icons, actions, and descriptions; 
  -- Improved rendering for grouped lists with accessibility optimizations. 
Related issue -> https://github.com/telerik/kendo-themes-private/issues/431


packages/bootstrap/scss/list/_variables.scss
```diff
@@ -388,21 +410,33 @@ $kendo-list-item-selected-hover-text: k-color(on-primary) !default;
...
 /// @group list
-$kendo-list-group-item-text: null !default;
+$kendo-list-group-item-text: k-color(subtle) !default;
 /// The border color of the List group items..
 /// @group list
 $kendo-list-group-item-border: inherit !default;
 /// The base shadow of the List group items.
 /// @group list
-$kendo-list-group-item-shadow: null !default;
+$kendo-list-group-item-shadow: inset 0 1px 0 k-color(border) !default;
```
zhpenkov, line 352: Would be good to use elevation system.

packages/core/scss/components/prompt-box/_layout.scss
```diff
@@ -0,0 +1,58 @@
...
+    .k-prompt-box {
+        border-radius: $kendo-prompt-box-border-radius;
+        flex-flow: column nowrap;
+
+        .k-input-inner {
+            padding: $kendo-prompt-box-input-inner-padding;
+        }
+    }
+
+    .k-prompt-box-header {
```
zhpenkov: Should we add the footer compartment as well ?

packages/core/scss/components/segmented-control/_layout.scss
```diff
@@ -0,0 +1,67 @@
...
+        display: flex;
+        flex: 1;
+        align-items: center;
+        justify-content: flex-start;
+        border-width: $kendo-segmented-control-button-border-width;
+        border-radius: $kendo-segmented-control-button-border-radius;
+        cursor: pointer;
+        font-size: inherit;
+        line-height: inherit;
+        z-index: 1;
```
zhpenkov: We should use z-index system. I think "handle" ?

packages/core/scss/components/smart-box/_layout.scss
```diff
@@ -0,0 +1,33 @@
...
+        padding-block: $kendo-smart-toolbox-popup-padding-block;
+        border-radius: $kendo-smart-toolbox-popup-border-radius;
+        gap: $kendo-smart-toolbox-popup-spacing;
+
+        > .k-list {
+            border-radius: 0;
+        }
+
+        > .k-segmented-control {
+            margin-inline: k-spacing(2);
```
zhpenkov: Are those overrides necessary ?

tests/_output/bootstrap/list/list-grouping.png
zhpenkov, line 1: Should not we remove the old rendering ?

zhpenkov, line 1: Also we might add a dotted border for those list tests in order to better visualize how the paddings change.

tests/_output/bootstrap/list/list.png
zhpenkov, line 1: We can split this test

Maybe we can have a standalone list-actions
Or list-rtl

tests/_output/bootstrap/list/list-sizes.png
zhpenkov, line 1: here the first group and the second group are different
also when they do not have an icon should not they be less indented ?

tests/_output/bootstrap/segmented-control/segmented-control.png
zhpenkov, line 1: In the 100% width scenario should not the inner buttons be 50/50 and text centered in the middle

tests/_output/bootstrap/prompt-box/prompt-box-affixes.png
zhpenkov, line 1: This is quite complex.

Also i do not think we should be showing the chat behavour (affix-start "plus" moves to the affix-end) here.
as well as all other "attach files", "send button".

The PromptBox is a standalone component with standalone options for affix-start/affix-end.
we should just display a generic scenario with affix-start/end and header.

packages/html/src/list/list-group.tsx
```diff
@@ -4,6 +4,7 @@ export type KendoListGroupProps = {
     root?: boolean;
     virtualization?: boolean;
     label?: string;
+    groupIconName?: string;
```
zhpenkov, line 7: we do not pass this to the component

packages/html/src/list/list-group-item.tsx
```diff
@@ -1,9 +1,19 @@
...
 
 const className = `k-list-group-item`;
 
-export const ListGroupItem = (props: React.HTMLAttributes<HTMLLIElement>) => (
-    <li className={classNames(className, props.className)}>
-        <span className="k-list-item-text">{props.children}</span>
-    </li>
-);
+export type ListGroupItemProps = {
+    groupIconName?: string;
```
zhpenkov, line 7: we can rename to just `icon?: string;`

packages/html/src/prompt-box/prompt-box.spec.tsx
```diff
@@ -0,0 +1,117 @@
...
+                    [`k-input-${fillMode}`]: fillMode,
+                    'k-hover': hover,
+                    'k-focus': focus,
+                    'k-disabled': disabled
+                }
+            )}
+        >
+            {header && <div className={`${PROMPT_BOX_CLASSNAME}-header`}>{header}</div>}
+            <div className={`${PROMPT_BOX_CLASSNAME}-content`}>
+                {startAffix && <div className={`${PROMPT_BOX_CLASSNAME}-start-affix`}>{startAffix}</div>}
```
zhpenkov: might use `addon` as a naming if we want to differ from prefix/suffix

packages/html/src/segmented-control/segmented-control-button.spec.tsx
```diff
@@ -0,0 +1,94 @@
...
+export type KendoSegmentedControlButtonOptions = {
+  variant?: (typeof SEGMENTEDCONTROLBUTTON_VARIANTS)[number] | null;
+};
+
+export type KendoSegmentedControlButtonProps = KendoSegmentedControlButtonOptions & {
+  icon?: string;
+  text?: string;
+  iconClassName?: string;
+  showArrow?: boolean;
+  arrowIconName?: string;
```
zhpenkov: Do not need those

packages/html/src/segmented-control/segmented-control-button.spec.tsx
```diff
@@ -0,0 +1,94 @@
...
+    States.selected,
+    States.hover,
+    States.focus,
+    States.disabled,
+] as const;
+
+const options = {};
+
+export type KendoSegmentedControlButtonOptions = {
+  variant?: (typeof SEGMENTEDCONTROLBUTTON_VARIANTS)[number] | null;
```
zhpenkov: Do we use that

packages/html/src/segmented-control/tests/segmented-control-states.tsx
```diff
@@ -0,0 +1,32 @@
...
+    }
+`;
+
+export default () => (
+    <>
+        <style>{styles}</style>
+        <div id="test-area" className="k-d-grid">
+
+            <SegmentedControl>
+                {(SegmentedControlButton.states || []).map((state) => (
```
zhpenkov: We can simplify to `SegmentedControlButton.states.map`

packages/html/src/smart-toolbox/templates/smart-toolbox-opened-empty.tsx
zhpenkov, line 1: Speaking of the templates are those the default smart-toolbox template variants?

I think this component should also be displayed here as a standalone component.
we need this scenario only for the grid integration with search / semantic search.

I think in the smart-toolbox tests/templates we should only add generic scenarios.

Those complex scenarios with segmented control, etc.. I think should be displayed in their own context, for example in a grid test.

packages/bootstrap/scss/list/_variables.scss
inikolova: The idea here is to use a shadow as a horizontal divider line, rather than adding extra elements or using a border. Since borders are included in an element’s height calculation, using a shadow avoids affecting layout and spacing, which is important for the virtualization feature

packages/core/scss/components/prompt-box/_layout.scss
inikolova: The footer section has not been implemented at this point, as there are no current requirements or scenarios that call for it.

packages/core/scss/components/smart-box/_layout.scss
inikolova: Yep; I checked with @lkarakoleva

tests/_output/bootstrap/list/list-grouping.png
inikolova: As discussed offline, a separate commit was created to remove the old rendering. I’m keeping it separate for now as a backup, to ensure all teams can properly adopt the changes.

tests/_output/bootstrap/list/list-sizes.png
inikolova: The groups are actually the same. The difference you’re referring to is the sticky header. I’ve updated the example description to make this clearer.

packages/bootstrap/scss/list/_variables.scss
epetrow: If that is the case then is `$kendo-list-group-item-shadow` an appropriate name? In my opinion it is not, as it is misleading.

tests/_output/bootstrap/segmented-control/segmented-control.png
inikolova: The buttons are actually split 50/50. Please ignore the white selection, as it uses a fixed width of 100px in all tests; in the real implementation, its width will be dynamically calculated by the component. Regarding text centering, after syncing with the designers, it should be centered in all scenarios, so I’ve updated the styles accordingly.

tests/_output/bootstrap/list/list-grouping.png
inikolova: I am not sure if I understand the idea here - could you please elaborate?

tests/_output/bootstrap/prompt-box/prompt-box-affixes.png
inikolova: I understand the concern. The prompt box is indeed a standalone component and doesn’t have any direct relationship to the chat. The tests use generic items—such as buttons and icons—as affixes. Some of them may resemble elements used in the chat (which is understandable), but that isn’t the intended implication.

The goal of the tests is to demonstrate that the component supports a wide range of affixes, and in my view, they help illustrate clearly that any type of affix can be added.

packages/html/src/segmented-control/segmented-control-button.spec.tsx
inikolova: Nope, removed!

packages/core/scss/components/list/_layout.scss
```diff
@@ -138,6 +146,45 @@
...
+        align-items: center;
+        gap: $kendo-list-item-gap;
+    }
+
+    .k-list-item:has(.k-list-item-icon) .k-list-item-text::before,
+    .k-list-item:has(.k-list-item-actions) .k-list-item-text::before {
+        content: none;
+    }
+
+    .k-list-item-text:has(+ .k-list-item-icon) {
```
epetrow: Maybe remove this?

packages/core/scss/components/prompt-box/_theme.scss
```diff
@@ -0,0 +1,4 @@
+@use "../../mixins/index.scss" as *;
```
epetrow: Is this import needed?

packages/core/scss/components/smart-toolbox/_layout.scss
```diff
@@ -0,0 +1,33 @@
...
+            margin-block-start: k-spacing(2);
+        }
+    }
+
+    .k-smart-toolbox-no-data {
+        flex-direction: column;
+        padding-inline: k-spacing(2);
+        padding-block: k-spacing(4);
+        gap: $kendo-smart-toolbox-popup-spacing;
+        font-size: var(--kendo-font-size-sm);
```
epetrow: Missing fallback value

packages/core/scss/components/smart-toolbox/_layout.scss
```diff
@@ -0,0 +1,33 @@
+@use "./variables.scss" as *;
+@use "../../spacing/index.scss" as *;
+
+@mixin kendo-smart-toolbox--layout-base() {
+
+    .k-smart-toolbox {}
```
epetrow: Could we consider removing this?

packages/core/scss/components/smart-toolbox/_theme.scss
```diff
@@ -0,0 +1,4 @@
+@use "../../mixins/index.scss" as *;
```
epetrow: Is this import needed?

packages/fluent/scss/list/_variables.scss
```diff
@@ -184,7 +195,7 @@ $kendo-list-item-group-label-lg-padding-y: var( --kendo-list-item-group-label-lg
 
 /// The border width of the List group items.
 /// @group list
-$kendo-list-group-item-border-width: var( --kendo-list-group-item-border-width, 1px 0 0 ) !default;
+$kendo-list-group-item-border-width: 0 !default;
```
epetrow: We are removing a previously existing css variable, is that intended?

packages/fluent/scss/list/_variables.scss
```diff
@@ -326,6 +333,9 @@ $kendo-list-item-bg: var( --kendo-list-item-bg, #{k-color(surface-alt)} ) !defau
 /// The text color of the List items.
 /// @group list
 $kendo-list-item-text: var( --kendo-list-item-text, #{k-color(on-app-surface)} ) !default;
+/// The border-radius of the List items.
+/// @group list
+$kendo-list-item-border-radius: var( --kendo-list-item-text, #{k-border-radius(md)} ) !default;
```
epetrow: The name of the css variable refers to a colour which would produce invalid css.

packages/fluent/scss/list/_variables.scss
```diff
@@ -370,13 +380,13 @@ $kendo-list-item-disabled-text: var( --kendo-list-item-disabled-text, #{color-mi
 $kendo-list-group-item-bg: var( --kendo-list-group-item-bg, #{k-color(surface-alt)} ) !default;
 /// The text color of the List group items.
 /// @group list
-$kendo-list-group-item-text: var( --kendo-list-group-item-text, #{k-color(on-app-surface)} ) !default;
+$kendo-list-group-item-text: var( --kendo-list-group-item-text, #{k-color(subtle)} ) !default;
 /// The border color of the List group items.
 /// @group list
-$kendo-list-group-item-border: var( --kendo-list-group-item-border, inherit ) !default;
+$kendo-list-group-item-border: var( --kendo-list-group-item-border, var( --kendo-list-group-item-border, inherit ) ) !default;
```
epetrow: I am confused here 😕

packages/fluent/scss/smart-toolbox/_variables.scss
```diff
@@ -0,0 +1,24 @@
...
+
+/// The inline padding of the popup of the Smart toolbox.
+/// @group smart-toolbox
+$kendo-smart-toolbox-popup-padding-inline: k-spacing(1) !default;
+/// The block padding of the popup of the Smart toolbox.
+/// @group smart-toolbox
+$kendo-smart-toolbox-popup-padding-block: k-spacing(1) !default;
+/// The border radius of the popup of the Smart toolbox.
+/// @group smart-toolbox
+$kendo-smart-toolbox-popup-border-radius: var( --kendo-slider-spacing, #{k-border-radius(md)}) !default;
```
epetrow: Do we actually need to couple this to `--kendo-slider-spacing`?

packages/html/src/autocomplete/templates/autocomplete-grouping-modern.tsx
epetrow, line 1: Are we removing this because we are unifying how grouping works now?

packages/html/src/list/list-group-item.tsx
```diff
@@ -1,9 +1,19 @@
...
+};
+
+export const ListGroupItem = (props: ListGroupItemProps & React.HTMLAttributes<HTMLLIElement>) => {
+    const { groupIconName, ...other } = props;
+
+    return (
+        <li {...other} className={classNames(className, props.className)}>
+            {groupIconName && <span className="k-list-item-icon-wrapper"><Icon icon={groupIconName} className="k-list-item-icon" /></span>}
+            <span className="k-list-item-text">{props.children}</span>
```
epetrow, line 16: Shouldn't we explicitly set text here as the name of the span element suggest? Since currently anything can be a child of `k-list-item-text` - a button, a grid

tests/_output/bootstrap/list/list-grouping.png
zhpenkov: We might add a style - `outline: 1px dotted;` for the section elements in this test.
This will help us easily catch if a padding has changed visually.
Right now the background of the list blends with the background of the page.

packages/html/src/prompt-box/prompt-box.spec.tsx
inikolova: As agreed in our offline discussion, the naming should be _**affix**_ , and the corresponding class name should be **_k-prompt-box-affix_**, regardless of whether it’s positioned at the start or the end.

tests/_output/bootstrap/prompt-box/prompt-box-affixes.png
zhpenkov: Are those pairs the same with the exception that the second has a header?
if this is the case we should make their elements the same in order to avoid confusion.

Screenshot 2026-01-08 at 11 27 03

packages/bootstrap/scss/list/_variables.scss
inikolova: This variable is already present in the themes, so I chose to use it instead of deleting it. We also have other variables across the theme that use hardcoded values to simulate borders—for example, $kendo-breadcrumb-focus-shadow.

packages/html/src/list/list.spec.tsx
```diff
@@ -1,10 +1,13 @@
...
+import { ListItem } from './list-item.spec';
+import { ListGroup } from './list-group';
+import { ListGroupItem } from './list-group-item';
+import { ListHeader } from './list-header';
 import { NoData } from '../nodata';
 
 import { KendoComponent } from '../_types/component';
 import { LIST_FOLDER_NAME, LIST_MODULE_NAME } from './constants';
-export const LIST_CLASSNAME = `k-list`;
+export const List_CLASSNAME = `k-list`;
```
epetrow: Why are we changing this?

tests/_output/bootstrap/prompt-box/prompt-box-affixes.png
inikolova: As you mentioned, the affixes should display generic items, and we can include any content we want inside them. In the second example, I chose to use an icon (rather than a button) to clearly distinguish between the header, where the button is displayed, and the start affix. In my view, it’s already clear that any content can be rendered within affixes, but I’ll update the example to make this even clearer and more convenient

packages/html/src/segmented-control/segmented-control-button.spec.tsx
inikolova: Nope :) removed, thank you :)

packages/core/scss/components/segmented-control/_layout.scss
inikolova: Actually we don't need z-index here at all, therefore I removed it

packages/html/src/prompt-box/prompt-box.spec.tsx
```diff
@@ -0,0 +1,117 @@
...
+export const PromptBox = (
+    props: KendoPromptBoxProps & KendoPromptBoxState & React.HTMLAttributes<HTMLDivElement>
+) => {
+    const {
+        fillMode = defaultOptions.fillMode,
+        lineMode = defaultOptions.lineMode,
+        startAffix,
+        endAffix,
+        header,
+        isExpanded = false,
```
epetrow: Why not this to defaultOptions as well?

packages/html/src/prompt-box/prompt-box.spec.tsx
```diff
@@ -0,0 +1,117 @@
...
+        children,
+        value,
+        placeholder,
+        hover,
+        focus,
+        disabled,
+        ...other
+    } = props;
+
+    let lineModeClassName = '';
```
epetrow: Instead of all this, why not just add `${PROMPT_BOX_CLASSNAME}-${lineMode}`

tests/_output/bootstrap/list/list-grouping.png
inikolova: Done

packages/html/src/segmented-control/segmented-control-button.spec.tsx
```diff
@@ -0,0 +1,85 @@
+import { Icon } from '../icon';
+import { classNames, stateClassNames, States} from '../misc';
+
+import { KendoComponent } from '../_types/component';
+import { SEGMENTEDCONTROLBUTTON_FOLDER_NAME, SEGMENTEDCONTROLBUTTON_MODULE_NAME } from './constants';
+export const SEGMENTEDCONTROLBUTTON_CLASSNAME = `k-segmented-control-button`;
```
epetrow: Why not `SEGMENTED_CONTROL_BUTTON`?

packages/html/src/segmented-control/segmented-control-button.spec.tsx
```diff
@@ -0,0 +1,85 @@
...
+            )}
+
+            {text
+                ?
+                <>
+                    {text && <span className="k-segmented-control-button-text">{text}</span>}
+                    {props.children}
+                </>
+                : props.children && <span className="k-segmented-control-button-text">{props.children}</span>
+            }
```
epetrow: This logic seems weird to me, we are checking `text` twice here, also there is inconsistency in the behaviour:

- When `text` exists -> it gets wrapped in the span, but the `children` are rendered unwrapped
- When `text` does not exist -> `children` get wrapped inside `k-segmented-control-button-text`

packages/html/src/segmented-control/segmented-control.spec.tsx
```diff
@@ -0,0 +1,63 @@
+import { classNames, optionClassNames, Size } from '../misc';
+
+import { KendoComponent } from '../_types/component';
+import { SEGMENTED_CONTROL_FOLDER_NAME, SEGMENTED_CONTROL_MODULE_NAME } from './constants';
+export const SEGMENTEDCONTROL_CLASSNAME = `k-segmented-control`;
```
epetrow: If we have `SEGMENTED_CONTROL_FOLDER_NAME` and `SEGMENTED_CONTROL_MODULE_NAME`, shouldn't this be `SEGMENTED_CONTROL_CLASSNAME`?

tests/_output/bootstrap/prompt-box/prompt-box.png
epetrow, line 1: What state are we showcasing in the examples without text? If it is normal as well, why do we need to show it twice.

tests/_output/bootstrap/segmented-control/segmented-control-sizes.png
epetrow, line 1: The search button in the Large size does not look good here

tests/_output/bootstrap/segmented-control/segmented-control.png
epetrow, line 1: Are we certain that the text and icon in the full width example should be aligned like that? I am referring to the Search button

packages/html/src/segmented-control/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const SEGMENTED_CONTROL_MODULE_NAME = 'k-segmented-control-button';
+export const SEGMENTED_CONTROL_FOLDER_NAME = SEGMENTED_CONTROL_MODULE_NAME;
+
```
TeyaVes, line 3: I'm a bit confused here- the SEGMENTED_CONTROL_MODULE_NAME is 'k-segmented-control-button', while the SEGMENTEDCONTROLBUTTON_MODULE_NAME is: segmentedcontrolbutton . I think both should be changed to - segmented-control and segmented-control-button

tests/_output/bootstrap/segmented-control/segmented-control.png
inikolova: If I am right, you are referring the "white" selection -> keep in mind that in all tests its width is hardcoded to 100/150px - in a real scenario this width will be dynamically calculated by the component and should cover the whole _k-segmented-control-button_.

tests/_output/bootstrap/segmented-control/segmented-control-sizes.png
inikolova: Could you elaborate a bit more? If you are referring the "white" selection, in all tests its width is hardcoded to 100/150px - in a real scenario this width will be dynamically calculated by the component and should cover the whole _k-segmented-control-button_. If this is not the case, please specify what does not look ok :)

tests/_output/bootstrap/prompt-box/prompt-box.png
inikolova: I followed the pattern from any other input component tests, where an empty input (without text or placeholder) is demonstrated. For example textbox -> https://github.com/telerik/kendo-themes/blob/develop/tests/_output/default/textbox/textbox-solid.png

packages/html/src/segmented-control/segmented-control.spec.tsx
inikolova: You're right - I changed it

packages/html/src/segmented-control/segmented-control-button.spec.tsx
inikolova: Updated! Thank you

packages/html/src/segmented-control/constants.ts
inikolova: Agree! Done :)

packages/bootstrap/scss/list/_variables.scss
```diff
@@ -7,6 +7,13 @@
 /// @group list
 $kendo-list-font-family: var( --kendo-font-family, inherit ) !default;
 
+/// The inline padding of the List container.
```
TeyaVes: let's change it to horizontal and vertical padding instead of inline and block for consistency

packages/bootstrap/scss/list/_variables.scss
```diff
@@ -154,6 +161,10 @@ $kendo-list-md-item-line-height: null !default;
 /// @group list
 $kendo-list-lg-item-line-height: null !default;
 
+/// The gap of the list item components.
```
TeyaVes: List item with capital L

packages/bootstrap/scss/list/_variables.scss
```diff
@@ -388,21 +406,33 @@ $kendo-list-item-selected-hover-text: k-color(on-primary) !default;
...
 /// The base shadow of the List group items.
 /// @group list
-$kendo-list-group-item-shadow: null !default;
+$kendo-list-group-item-shadow: inset 0 1px 0 k-color(border) !default;
 
 /// The color of the 'Option Label' text.
 /// @group list
 $kendo-list-option-label-text: k-color(subtle) !default;
 
+/// The font size of the list item description.
```
TeyaVes: List items with capital L here and below as well

packages/html/src/segmented-control/segmented-control-button.spec.tsx
inikolova: You are absolutely right -> there was an error in the logic, however I modified it and now the text is always wrapped inside k-segmented-control-button-text

packages/core/scss/components/prompt-box/_theme.scss
inikolova: Nope. Removed

packages/core/scss/components/smart-toolbox/_layout.scss
inikolova: done

packages/core/scss/components/smart-toolbox/_theme.scss
inikolova: nope. removed :)

packages/fluent/scss/list/_variables.scss
inikolova: Nope, it's a mistake. I revert it

packages/bootstrap/scss/segmented-control/_variables.scss
```diff
@@ -0,0 +1,165 @@
...
+/// The horizontal padding of the medium Segmented Control button.
+/// @group segmented-control
+$kendo-segmented-control-md-button-padding-x: k-spacing(3) !default;
+/// The horizontal padding of the large Segmented Control button.
+/// @group segmented-control
+$kendo-segmented-control-lg-button-padding-x: k-spacing(3) !default;
+
+/// The vertical padding of the small Segmented Control button.
+/// @group segmented-control
+$kendo-segmented-control-sm-button-padding-y: calc( #{k-spacing(.5)} + 1px ) !default;
```
TeyaVes, line 80: sometimes we are using k-spacing(1.5) / 2, simetimes we are using k-spacing(0.5) + 1px. Let's decide on one approach and use it. I personally vote for the + 1px approach

packages/fluent/scss/list/_variables.scss
inikolova: It appears $kendo-list-item-border-radius is defined twice in this file; I removed the wrong one

packages/fluent/scss/list/_variables.scss
inikolova: fixed

packages/fluent/scss/smart-toolbox/_variables.scss
inikolova: nope, it's a mistake. Fixed. Also added css variables in this file

packages/bootstrap/scss/list/_variables.scss
inikolova: done; in all themes

packages/bootstrap/scss/segmented-control/_variables.scss
inikolova: ok, I will update the calculations for concistency

packages/html/src/prompt-box/prompt-box.spec.tsx
inikolova: The current logic is intentional because:

- 'auto' mode shouldn't output a class name itself - it conditionally becomes multiline based on isExpanded;

- The class names use the full suffix -singleline and -multiline (not just -single/-multi)

packages/html/src/prompt-box/prompt-box.spec.tsx
inikolova: by mistake. I fixed it

packages/html/src/list/list.spec.tsx
inikolova: reverted

packages/html/src/autocomplete/templates/autocomplete-grouping-modern.tsx
inikolova: Yep; we should have only one type of grouping - no more _classic grouping_, _modern grouping_, _angular grouping_ etc.

packages/html/src/list/list-group-item.tsx
inikolova: Actually this part hasn't been changed - https://github.com/telerik/kendo-themes/blob/9672e2cde2ee4888a3040f732ac0dee2d6d3b5d6/packages/html/src/list/list-group-item.tsx#L7

units/specificity-specs.ts
```diff
@@ -369,7 +369,7 @@ export const allComponents: SpecComponent[] = [
   ImageEditor,
   Input,
   List,
-  ListModern,
+  // ListModern,
```
TeyaVes: I suppose this should be deleted afterwards, not commented

packages/html/src/list/list-group.tsx
inikolova: groupIconName is needed on ListGroup - it's a pass-through prop that gets extracted and forwarded to ListGroupItem by the parent List component in list.spec.tsx

packages/html/src/list/list-group-item.tsx
inikolova: we have _iconName_ in the list-item.spec, so I prefer to keep the prop name _groupIconName_ in order to avoid confusions

units/specificity-specs.ts
inikolova: Yes, I will delete it, once all teams adopt the new rendering

packages/html/src/prompt-box/prompt-box.spec.tsx
epetrow: Okay what do you think about this then? 💭 

```js

const isMultiline = lineMode === 'multi' || (lineMode === 'auto' && isExpanded);

{
   [`${PROMPT_BOX_CLASSNAME}-singleline`]: lineMode === 'single',
   [`${PROMPT_BOX_CLASSNAME}-multiline`]: isMultiline,
   [`k-input-${fillMode}`]: fillMode,
   'k-hover': hover,
   'k-focus': focus,
   'k-disabled': disabled
}
```

tests/_output/bootstrap/segmented-control/segmented-control.png
epetrow: image
 Here I am referring to how the text sits in the "white" selection, I think regardless of the width as it is at the moment it will always sit at the end, which looks a bit weird. Am I wrong?

packages/html/src/prompt-box/prompt-box.spec.tsx
inikolova: I like this idea and added it to the spec :) Thank you :)

tests/_output/bootstrap/segmented-control/segmented-control.png
inikolova: The image below demonstrates the hover state of the actual button -> and the text is centered. Keep in mind that the "white" selection as an additional HTML element, which is positioned dynamically above the "selected" button. It also has a fluid width, which should be dynamically calculated depending on the actual button content 
Screenshot 2026-01-08 at 17 56 11

packages/html/src/smart-toolbox/templates/smart-toolbox-opened-empty.tsx
inikolova: done :)

packages/html/src/list/list.spec.tsx
zhpenkov, line 1: We can refactor the `list.spec.tsx` to remove that extra children logic and compose the component by its subcomponents.

inikolova:
@telerik/kendo-a11y-core, can you please review the proposed changes?

packages/core/scss/components/smart-box/_layout.scss
```diff
@@ -0,0 +1,55 @@
...
+            margin-block-start: k-spacing(2);
+        }
+
+        .k-list-item-text {
+            white-space: nowrap;
+        }
+    }
+
+    .k-smart-box-no-data {
```
epetrow, line 76: I wonder could we use the existing `k-no-data` styles here or does the smart-box variant have to look different?

packages/default/lib/swatches/default-ocean-blue-a11y.json
```diff
@@ -727,16 +727,6 @@
...
-                "kendo-chat-file-bg": {
-                    "name": "The background color of the Chat file",
-                    "type": "color",
-                    "value": "transparent"
-                },
-                "kendo-chat-file-border": {
-                    "name": "The border color of the Chat file",
-                    "type": "color",
-                    "value": "#424242"
-                },
```
epetrow, line 739: If we are removing the chat specific file variables because we are introducing a file-box "component" do we need to add it's variables here or is that not needed anymore?

packages/html/src/file-box/file-box-wrapper.tsx
```diff
@@ -0,0 +1,45 @@
+import { classNames } from "../misc";
+
+export const FILEBOXWRAPPER_CLASSNAME = `k-file-box-wrapper`;
```
epetrow: `FILE_BOX_MODULE_NAME`, `FILE_BOX_FOLDER_NAME` vs `FILEBOXWRAPPER_CLASSNAME`? Shouldn't we follow some kind of consisntency?

packages/html/src/file-box/file-box.tsx
```diff
@@ -0,0 +1,75 @@
+import { classNames, States } from '../misc';
+import { Icon } from '../icon';
+import { IconButton } from '../button';
+import { MenuButton } from '../menu-button';
+
+export const FILEBOXFILE_CLASSNAME = 'k-file-box';
```
epetrow: Some point here, the naming is not consistent

packages/html/src/file-box/tests/file-box-scroll-gradient.tsx
```diff
@@ -0,0 +1,43 @@
...
+            <span>Scrollable start</span>
+            <span>scrollable end</span>
+            <span>Scrollable both</span>
+
+            <section>
+                <FileBoxWrapper scrollingPosition="start">
+                    <FileBox />
+                    <FileBox />
+                    <FileBox />
+                </FileBoxWrapper>
```
epetrow: Looking at this, wouldn't it make more sense for the wrapper/container to be the actual FileBox and the individual items inside the box to be File or FileItem (I would prefer `k-file` but I know we can't do that)

tests/_output/bootstrap/autocomplete/autocomplete-custom-value.png
epetrow, line 1: Isn't weird not to use elevation for the sticky header here?

tests/_output/bootstrap/smart-box/smart-box-size-rounded.png
epetrow, line 1: The prefix and suffix look really weird in medium and large sizes 😕

tests/_output/classic/action-sheet/action-sheet-list.png
epetrow, line 1: Is the change here expected?

packages/default/lib/swatches/default-ocean-blue-a11y.json
inikolova: It is not needed any more, as the design was changed

packages/core/scss/components/smart-box/_layout.scss
inikolova: Yes, it has very different look

packages/html/src/file-box/tests/file-box-scroll-gradient.tsx
inikolova: I was thinking on the names for a long time, but in my opinion the box is the actual file - it consists of icons, text, buttons. Therefore I prefer to keep the naming as is

packages/core/scss/components/segmented-control/_variables.scss
```diff
@@ -0,0 +1,75 @@
...
+$kendo-segmented-control-lg-button-padding-y: null !default;
+
+$kendo-segmented-control-button-border-width: null !default;
+$kendo-segmented-control-button-border-radius: null !default;
+
+$kendo-segmented-control-sm-button-gap: null !default;
+$kendo-segmented-control-md-button-gap: null !default;
+$kendo-segmented-control-lg-button-gap: null !default;
+
+// $kendo-segmented-control-sizes: null !default;
```
epetrow: Should we remove this?

packages/html/src/smart-box/smart-box-send-button.tsx
```diff
@@ -0,0 +1,52 @@
...
+            {...other}
+            rounded="full"
+            size="small"
+            icon={processing ? "stop-sm" : "arrow-up-outline"}
+            active={processing}
+            className={classNames(
+                props.className,
+                SMART_BOX_SEND_BUTTON_CLASSNAME,
+                {
+                    ["k-processing"]: processing,
```
zhpenkov, line 39: is this like `k-generating` ?
Can't we reuse it here ?

inikolova: It was discussed in the working group that this component does not generate anything and we agreed that the state should be named _Processing_. Therefore I used this className

packages/html/src/segmented-control/segmented-control-button.spec.tsx
```diff
@@ -0,0 +1,80 @@
...
+        >
+            {icon && (
+                <Icon
+                    className={classNames(iconClassName, 'k-segmented-control-button-icon')}
+                    icon= {icon}
+                />
+            )}
+
+            {(text || props.children) && (
+                <span className="k-segmented-control-button-text">{text || props.children}</span>
```
epetrow, line 68: Are we intentionally wrapping everything in `k-segmented-control-button-text` even if it is not actually text?

packages/html/src/smart-box/smart-box.spec.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+export const SMART_BOX_CLASSNAME = `k-smart-box`;
+
+const states = [
+    States.hover,
+    States.focus,
+    States.disabled,
+];
+
+const options = {
```
epetrow, line 21: It's redundant to define these options here and then set size, rounded and fillMode to `Input.defaultOptions`. If we are extending the Input, lets add only options, states, etc. that are only specific to SmartBox

packages/html/src/segmented-control/segmented-control-button.spec.tsx
inikolova: This component follows the implementation of the Button component, however I update the spec to get text only

packages/bootstrap/scss/smart-box/_variables.scss
```diff
@@ -0,0 +1,71 @@
+@use "sass:map";
+@use "../core/_index.scss" as *;
+
+// Smart box
+
+/// The default size of the SmartBox.
+/// @group smart-box
+/// @role default
+$kendo-smart-box-default-size: "md" !default;
+/// The default size of the SmartBox.
```
TeyaVes: the default roundness

packages/bootstrap/scss/smart-box/_variables.scss
```diff
@@ -0,0 +1,71 @@
...
+/// The horizontal padding of the medium Smart box.
+/// @group smart-box
+$kendo-smart-box-md-padding-x: k-spacing(1) !default;
+/// The horizontal padding of the large Smart box.
+/// @group smart-box
+$kendo-smart-box-lg-padding-x: calc( #{k-spacing(2)} + 1px) !default;
+
+/// The horizontal margin of the prefix icon in the small Smart box.
+/// @group smart-box
+$kendo-smart-box-sm-prefix-margin-x: k-spacing(0) !default;
```
TeyaVes: I would rather suggest using 0 instead of k-spacing(0) for consistency with the smart-box-popup-padding-inline

packages/fluent/scss/list/_variables.scss
```diff
@@ -137,6 +144,10 @@ $kendo-list-md-item-line-height: var( --kendo-list-md-item-line-height, var(--ke
 /// @group list
 $kendo-list-lg-item-line-height: var( --kendo-list-lg-item-line-height, var(--kendo-line-height, normal) ) !default;
 
+/// The gap of the List item components.
+/// @group list
```
TeyaVes, line 148: add CSS var here for consistency

packages/fluent/scss/list/_variables.scss
```diff
@@ -321,9 +328,21 @@ $kendo-list-option-label-text: var( --kendo-list-option-label-text, #{k-color(su
 /// @group list
 $kendo-list-option-label-disabled-text: var( --kendo-list-option-label-disabled-text, #{color-mix(in srgb, k-color(on-app-surface) 46%, transparent)} ) !default;
 
+/// The font size of the List item description.
+/// @group list
+$kendo-list-item-description-font-size: var( --kendo-list-item-description-font-size, var( --kendo-font-size-sm, inherit ) ) !default;
+/// The line of the list item description.
```
TeyaVes: the line HEIGHT

packages/fluent/scss/smart-box/_variables.scss
```diff
@@ -0,0 +1,70 @@
+@use "sass:map";
+@use "../core/_index.scss" as *;
+
+// Smart box
+
+/// The default size of the SmartBox.
+/// @group smart-box
+/// @role default
+$kendo-smart-box-default-size: "md" !default;
+/// The default size of the SmartBox.
```
TeyaVes: roundness here and in the other variables files as well

packages/html/src/smart-box/smart-box-send-button.tsx
zhpenkov: In the feature, we might stick with one naming for all generating/processing/sending animations and reuse it.
Especially if they are visually the same.

PR #5698, opened 2025-12-15
Chat v3

inikolova:
Design https://www.figma.com/design/D29kA4UkYk4xUkCzD1AzHM/KendoUI-Chat-Design?node-id=324-5946&m=dev

Related to https://github.com/telerik/kendo-themes-private/issues/445

**- Prompt Box**
Design -> https://www.figma.com/design/gxBPvHG9PZfx8phbWWny8z/PromptBox?node-id=0-1&p=f&m=dev
The PromptBox is a composite component that consists of:
  -- A text input (textbox or textarea, depending on the line mode);
  -- A header section that allows custom content.
This component is intended for internal usege within the Chat component (for now) and supports three line modes:
  -- Single – a single-line text input;
  -- Multi – a multiline textarea;
  -- Auto – starts as a single-line textarea and automatically increases its height as the user types.

This PR improves Chat rendering by introducing new features and refining existing UI behavior.

### Rendering updates

- **Header**: Replaced the AppBar component with the Toolbar component;
- **Pinned message**: The pinned message is now positioned as sticky with a transparent background (inside the _k-message-list_ element), allowing the chat content to scroll beneath it;
- **Message metadata**: Message time now appears on the same line as the status message, positioned below the message bubble;
- **Menu/Download buttons** inside attachments are changes to small size;
- **Download All icon** is changed to "download-light" (previously download);
- The **pin** icon is changed to **pin-outline**;
- With this release, we replaced the chat textarea with the new PromptBox component. As a result, the previous chat textarea files are now part of the PromptBox itself. With regard to this change, we extracted a new FileBox component that replaces the former chat-files implementation. This required updating the following CSS classes:
- k-chat-file → k-file-box
- k-chat-file-wrapper → k-file-box-wrapper

### New features

- **Message send failure handling**: Added rendering for failed message states, including a resend button;
- **Scroll-to-bottom button**: Added a floating action button (FAB) using the light theme color;
- **Scroll buttons**: Introduced scroll buttons with gradient styling over the suggestions;
- **Horizontal line between suggestions and messages** visually indicates where the messages are cut. This divider should appear when the chat massages increase and scrollbar appears;
- Gradient when scrolling suggestions and files inside promptBox;

### Visual refinements
- **Avatar alignment**: Avatars are now aligned with their corresponding message bubbles;
- **Sequential message styling**: Updated border radius for sequential messages;
- **Chat file backgrounds**: Chat file attachments now use a non-transparent background.


packages/html/src/chat/chat.spec.tsx
```diff
@@ -49,6 +52,8 @@ export const Chat: KendoComponent<KendoChatProps & React.HTMLAttributes<HTMLDivE
         tools = defaultOptions.tools,
         value,
         empty,
+        scrollToBottom,
+        ifScrollbarAppears,
```
TeyaVes: wouldn't it be better to name this srollbar? ifScrollbarAppears seems too complicated to me

packages/html/src/chat/tests/chat-message-failed.tsx
```diff
@@ -0,0 +1,55 @@
...
+export default () =>(
+    <>
+        <style>{styles}</style>
+        <div id="test-area" className="k-d-grid k-grid-cols-2">
+            <span>Messages</span>
+            <span>Full width messages</span>
+            <section>
+                <ChatNormal>
+                    <ChatMessageGroup type="sender">
+                        <ChatMessage resendIcon="arrow-rotate-cw" failedIcon='warning-triangle' messageFailed="Failed to send" text="Failed message with Author and Avatar." />
```
TeyaVes: is the failedIcon always warning-triangle and the resendIcon arrow-rotate-cw? If yes, why don't we add them to the spec?

packages/html/src/chat/tests/chat.tsx
```diff
@@ -1,9 +1,11 @@
 import { ChatHeader, ChatMessage, ChatNormal, ChatTimestamp, ChatMessageGroup, ChatEvent } from '..';
+import { ToolbarItem } from '../../toolbar';
 
 
 const styles = `
     .k-chat {
         min-height: 370px;
+        max-height: 370px;
```
TeyaVes, line 8: isn't it better to use height? If the max and min-height are 370px, then the height should be 370px?

packages/bootstrap/scss/chat/_variables.scss
```diff
@@ -41,6 +41,10 @@ $kendo-chat-message-max-width: min(75%, 460px) !default;
 /// @group chat
 $kendo-chat-message-gap: k-spacing(2) !default;
 
+/// The border radius of sequential Chat messages.
+/// @group chat
+$kendo-sequential-messages-border-radius: 2px !default;
```
TeyaVes: I would suggest using border-radius-sm here

packages/classic/scss/chat/_variables.scss
```diff
@@ -39,6 +39,10 @@ $kendo-chat-message-max-width: min(75%, 460px) !default;
 /// @group chat
 $kendo-chat-message-gap: k-spacing(2) !default;
 
+/// The border radius of sequential Chat messages.
+/// @group chat
+$kendo-sequential-messages-border-radius: 2px !default;
```
TeyaVes: border-radius-sm

packages/core/scss/components/suggestion/_layout.scss
```diff
@@ -60,6 +66,22 @@
...
+    .k-suggestion-group-scrollable-end::before,
+    .k-suggestion-group-scrollable-start::after,
+    .k-suggestion-scrollwrap .k-suggestion-group-scrollable-end::before,
+    .k-suggestion-scrollwrap .k-suggestion-group-scrollable-start::after {
+        content: '';
+        aspect-ratio: .3;
+        position: absolute;
+        top: 0;
+        z-index: k-z-index("base", 3);
+        height: calc( var( --kendo-font-size, .875rem ) * #{$kendo-suggestion-line-height} + #{$kendo-suggestion-padding-y} * 2 + #{$kendo-suggestion-padding-y} + #{$kendo-suggestion-border-width} * 2);
```
TeyaVes: why are we adding the kendo-suggestion-padding-y 3 times here?

packages/core/scss/components/suggestion/_theme.scss
```diff
@@ -79,4 +79,14 @@
         }
     }
 
+    .k-suggestion-group-scrollable-end::before,
+    .k-suggestion-scrollwrap .k-suggestion-group-scrollable-end::before {
```
TeyaVes, line 57: do we need to add an RTL rule here?

tests/_output/bootstrap/chat/chat-expandable-messages.png
TeyaVes, line 1: the toggle icon looks a bit weird to me, do we have a design for it?

packages/core/scss/components/chat/_layout.scss
```diff
@@ -435,10 +469,10 @@
 
     .k-message-box-wrapper {
         padding-inline: $kendo-chat-padding-x;
-        padding-block: calc( #{$kendo-chat-padding-y} / 2 ) $kendo-chat-padding-y;
+        padding-block: $kendo-chat-message-box-wrapper-spacing $kendo-chat-padding-y;
```
epetrow: I wonder should we set padding through the spacing variable? Would it be an expected outcome to change a variable related to the spacing of items inside a wrapper and this to also modify its padding?

packages/core/scss/components/chat/_theme.scss
```diff
@@ -62,6 +62,19 @@
...
     }
 
+    .k-message-failed .k-chat-bubble,
+    .k-message-failed .k-chat-bubble.k-selected {
+        @include fill(
+            $kendo-chat-failed-bubble-text,
+            $kendo-chat-failed-bubble-bg,
+            $kendo-chat-failed-bubble-border
+        );
+    }
```
epetrow: The normal and selected states are the same? 🤔

packages/core/scss/components/suggestion/_layout.scss
epetrow: Would it be expected for the height of the pseudo-element to change if someone customizes the typography?

packages/core/scss/components/suggestion/_layout.scss
```diff
@@ -60,6 +66,22 @@
...
+    }
+
+    .k-suggestion-group-scrollable-start::after,
+    .k-suggestion-scrollwrap .k-suggestion-group-scrollable-start::after {
+        right: 0;
```
epetrow: Could we use `inset-block-end` instead?

packages/default/scss/chat/_variables.scss
```diff
@@ -39,6 +39,10 @@ $kendo-chat-message-max-width: min(75%, 460px) !default;
 /// @group chat
 $kendo-chat-message-gap: k-spacing(2) !default;
 
+/// The border radius of sequential Chat messages.
+/// @group chat
+$kendo-sequential-messages-border-radius: 2px !default;
```
epetrow: Can we use a value from the radii module?

packages/fluent/scss/chat/_variables.scss
```diff
@@ -40,6 +40,10 @@ $kendo-chat-message-max-width: min(75%, 460px) !default;
 /// @group chat
 $kendo-chat-message-gap: k-spacing(2) !default;
 
+/// The border radius of sequential Chat messages.
+/// @group chat
+$kendo-sequential-messages-border-radius: 2px !default;
```
epetrow: Is it possible to use a value from the radii module here as well?

packages/html/src/chat/chat-message.tsx
```diff
@@ -67,6 +70,9 @@ export const ChatMessage = (
         expanded,
         removed,
         files,
+        messageFailed,
+        failedIcon,
+        resendIcon,
```
epetrow: Having these three separate props seems redundant to me. When a message fails, you'd typically want to show all the failure-related UI elements together.

A cleaner approach would be to use a single `failed` prop (or something similar) that could be:

1. `boolean` - Shows the failed state with default icons and text
2. `string` - Shows the failed state with custom message text (and default icons)

What do you think?

packages/html/src/chat/tests/chat-message-failed.tsx
epetrow: In my opinion there even is no need for three separate props for the failed state

tests/_output/bootstrap/bubble/bubble.png
epetrow, line 1: I can' seem to find this example in the design, but with this change the expand/collapse icon and the text are not aligned and it look a bit strange.

tests/_output/bootstrap/chat/chat-bubble-states.png
epetrow, line 1: Is the spacing between the message bubbles in the case for sender and receiver supposed to be different?

tests/_output/bootstrap/chat/chat-upload.png
epetrow, line 1: Maybe we can tweak the `min-height` of the first chat so that the example with the uploaded single file is clearly visible

packages/html/src/chat/tests/chat.tsx
inikolova: There is _min-height: 600px_, which comes from the themes

packages/core/scss/components/suggestion/_layout.scss
inikolova: Maybe you mean inset-inline-end: 0; :) yes, we could and I added it

packages/default/scss/chat/_variables.scss
inikolova: done

packages/core/scss/components/suggestion/_layout.scss
inikolova: _.k-suggestion-group.k-suggestion-group-scrollable_ has **padding-block: calc( #{$kendo-suggestion-padding-y} / 2 );** In the calculation for the height of :before/ :after elements I added $kendo-suggestion-padding-y instead of (calc( #{$kendo-suggestion-padding-y} / 2 ) * 2)

packages/core/scss/components/chat/_theme.scss
inikolova: After sync with @lkarakoleva we added different hover and selected states

packages/core/scss/components/suggestion/_layout.scss
inikolova: I added **_$kendo-suggestion-font-size_** variables across the themes and used them for this calculation

tests/_output/bootstrap/chat/chat-bubble-states.png
inikolova: I am not sure which spacing you refer -> I double checked and it appears the spacing between bubbles is the same. Could you provide more details?

packages/html/src/chat/chat-message.tsx
inikolova: done

tests/_output/bootstrap/bubble/bubble.png
inikolova: Fixed

tests/_output/bootstrap/chat/chat-expandable-messages.png
inikolova: Fixed

packages/core/scss/components/chat/_layout.scss
inikolova: After sync with @lkarakoleva we reverted the previous padding-block-start value

tests/_output/bootstrap/chat/chat-bubble-states.png
epetrow: I am referring to the screenshot below, maybe it is purely visual, but the spacing seems different between sender and receiver for the focused and selected states.
image

inikolova: I double checked the margin and confirm that it is the same. It's maybe a visual illusion because of the shadow colors Screenshot 2026-01-22 at 8 55 40
Screenshot 2026-01-22 at 8 55 56

inikolova:
@telerik/kendo-a11y-core, can you please review the proposed changes?

packages/core/scss/components/chat/_theme.scss
```diff
@@ -62,6 +62,29 @@
...
 
+    .k-message-failed .k-chat-bubble {
+        @include fill(
+            $kendo-chat-failed-bubble-text,
+            $kendo-chat-failed-bubble-bg,
+            $kendo-chat-failed-bubble-border
+        );
+
+        &:hover,
+        &.k-state-hover {
```
epetrow: This is a legacy class name that we no longer use, change to `k-hover`

packages/html/src/chat/chat-file.tsx
epetrow, line 1: This looks really weird to me, if the idea is to just wrap/extend `FileBox` then  there is no need to duplicate the static properties. Also I would question the need for the `k-chat-file-deleted` class if we already have `k-file-deleted`

packages/html/src/chat/chat-files-wrapper.tsx
epetrow, line 1: I have the same issue with this file as well, plus the layout prop is a bit confusing.

First we have:
```tsx
const layout = props.layout ?? defaultOptions.layout;  // Defaults to 'vertical'
```

But then we say:

```tsx
<FileBoxWrapper
    {...other}
    layout="horizontal"  //  Always hardcodes "horizontal" ?
    className={classNames(
        ...
        {
            [`k-chat-files-horizontal`]: layout === "horizontal",
            [`k-chat-files-wrap`]: layout === "wrap"
        }
    )}
>
```

packages/html/src/chat/chat-file.tsx
inikolova: There is no actual need. I left the classes in place to avoid them being removed and avoid breaking changes in the Chat

packages/html/src/chat/chat-files-wrapper.tsx
inikolova: Yep - I set this layout="horizontal", because of:
- The default layout of the FileBoxWrapper is horizontal and it does not set a className;
- The default layout of the files inside a chart message is vertical and I wanted to avoid adding k-files-vertical className;

Do you think this is not a good approach?

packages/bootstrap/scss/file-box/_variables.scss
```diff
@@ -0,0 +1,73 @@
...
+/// The text color of the File Box.
+/// @group file-box
+$kendo-file-box-text: k-color(base-on-subtle) !default;
+/// The border color of the File Box.
+/// @group file-box
+$kendo-file-box-border: $kendo-file-box-bg !default;
+
+/// The gradient used for the File Box wrapper scrolling indicator.
+/// @group file-box
+$kendo-file-box-wrapper-scroll-gradient: rgba(255, 255, 255, 0.00) 0%, #FFFFFF 100% !default;
```
TeyaVes: one of the values is in rgba, the other one in hex, let's keep them the same?

packages/bootstrap/scss/suggestion/_variables.scss
```diff
@@ -39,16 +43,22 @@ $kendo-suggestion-focus-shadow-color: color-mix(in srgb, k-color(on-app-surface)
 /// @role default
 $kendo-suggestion-default-theme-color: "base" !default;
 
+/// The gradient used for scroll indicators in the Suggestion group.
+/// @group suggestion
+$kendo-suggestion-scroll-gradient: rgba(255, 255, 255, 0.00) 0%, #FFFFFF 100% !default;
```
TeyaVes: same here

packages/core/scss/components/bubble/_layout.scss
```diff
@@ -53,9 +53,8 @@
         .k-bubble-expandable-indicator {
             padding-inline: $kendo-bubble-expandable-icon-padding-x;
             padding-block: $kendo-bubble-expandable-icon-padding-y;
-            margin-block-start: calc( #{$kendo-bubble-padding-y} * -1);
             margin-inline-end: calc( #{$kendo-bubble-padding-x} * -1);
-            margin-block-end: calc( ( #{$kendo-bubble-padding-x} / 2 ) * -1);
+            margin-block: calc( (#{$kendo-bubble-expandable-icon-padding-y} - 2px) * -1);
```
TeyaVes: Why are we subtracting 2px here?

packages/fluent/scss/chat/_variables.scss
```diff
@@ -6,10 +6,10 @@
...
 /// @group chat
-$kendo-chat-padding-x: var( --kendo-chat-padding-x, #{k-spacing(4)} ) !default;
+$kendo-chat-padding-x: var( --kendo-chat-padding-x, #{k-spacing(5)} ) !default;
 /// The vertical padding of the Chat.
 /// @group chat
-$kendo-chat-padding-y: k-spacing(4) !default;
+$kendo-chat-padding-y: var( --kendo-chat-padding-y, #{k-spacing(5)} ) !default;
 /// The minimum width of the Chat.
 /// @group chat
 $kendo-chat-width: 280px !default;
```
TeyaVes: For some of the variables, we have introduced CSS vars, but for others we haven't. It seems a bit inconsistent to me.

packages/core/scss/components/bubble/_layout.scss
inikolova: This is the bubble border width. I changed it with a variable

packages/fluent/scss/prompt-box/_variables.scss
```diff
@@ -0,0 +1,34 @@
...
+/// @group prompt-box
+$kendo-prompt-box-border-radius: var( --kendo-prompt-box-border-radius, k-border-radius(xxxl) ) !default;
+
+/// The box shadow of the PromptBox.
+/// @group prompt-box
+$kendo-prompt-box-box-shadow: var( --kendo-prompt-box-box-shadow, null ) !default;
+
+/// The padding of the PromptBox header.
+/// @group prompt-box
+$kendo-prompt-box-header-padding: var( --kendo-prompt-box-header-padding, k-spacing(2) ) !default;
```
TeyaVes: shall we add interpolation to the fallback values everywhere?

packages/fluent/scss/chat/_variables.scss
inikolova: This is an omission from previous releases, however I will add the missing css vars

packages/fluent/scss/prompt-box/_variables.scss
inikolova: I think it is not necessary. however I added.

PR #5717, opened 2026-01-09
chore(html): add component demos

kvelikov:


packages/html/src/icon/icon.spec.tsx
```diff
@@ -58,11 +60,13 @@ const defaultOptions = {
 export const Icon: KendoComponent<IconProps & React.HTMLAttributes<HTMLSpanElement>> = (
     props: IconProps & React.HTMLAttributes<HTMLSpanElement>
 ) => {
+    const { iconsType } = useContext(ConfigurationContext);
```
epetrow, line 64: Do we really need to introduce context here?

tests/_output/classic/list/list-grouping.png
inikolova, line 1: In my opinion we should keep the second row with lists without sticky headers

epetrow: I was wondering why that is and was actually a bit confused. What are we trying to show in this test? If it is one row with sticky headers, one without maybe we should update the description

PR #5730, opened 2026-01-21
[TabStrip]: Add overflow menu support

epetrow:
closes: https://github.com/telerik/kendo-themes-private/issues/444


tests/_output/default/tabstrip/tabstrip-overflow.png
TeyaVes, line 1: I am wondering if we shall do anything about the border under the selected tab which is only visible in the overflow tabs scenario

PR #5740, opened 2026-01-27
Migrate to vitest

Juveniel:


packages/html/package.json
zhpenkov, line 1: running `npm run test:html` at root produced two warnings

1. First one 
Image

About the types I think, they should be?

```js

  "exports": {
    "types": "./dist/types/index.d.ts",
    "import": "./dist/esm/index.mjs",
    "require": "./dist/cjs/index.js"
  }

```

2. Second one

Image

Try adding the file extension in the static path. in `spec-requirements.test.ts`

```ts
const specModule = await import(`../src/${moduleName}/${path.basename(moduleName)}.spec.tsx`);
```

units/package.json
zhpenkov, line 1: Do we need the `THEME=` variable in the npm scripts, since each theme folder already has a `theme.env.js` that sets the same value?

units/specificity-reporter.ts
```diff
@@ -0,0 +1,105 @@
+/* eslint-disable no-console */
+import type { Reporter, File, TaskResultPack } from 'vitest';
```
zhpenkov: Those are deprecated

units/vitest.config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    environment: 'jsdom',
+    include: ['**/*.test.ts'],
+    globals: false, // Keep explicit imports for better IDE support
+    // reporters: ['./specificity-reporter.ts'], // Uncomment to enable specificity reporter
```
zhpenkov, line 8: `specificity-reporter.ts` is commented out

zhpenkov: This seemed to be the case with the jest infra as well.

zhpenkov, line 8: Using this reporter, the output is broken

Image

units/package.json
epetrow: Nope, they were redundant

PR #5746, opened 2026-01-28
List chat polishing

inikolova:


packages/core/scss/components/suggestion/_layout.scss
```diff
@@ -27,6 +29,7 @@
 
         .k-suggestions-scroll {
             padding-block: calc( #{$kendo-suggestion-padding-y} / 2 );
+            margin-block: calc( (#{$kendo-suggestion-padding-y} / 2 ) * -1 );
```
TeyaVes: let's divide it by -2 rather than first divide by 2 and then multiply by -1

tests/_output/bootstrap/chat/chat-message-action-delete.png
TeyaVes, line 1: in some of the tests the pin icon is still the old one.

packages/core/scss/components/suggestion/_layout.scss
inikolova: We already use the *-1 approach in many places across the themes. My concern is that switching to a /-2 division would introduce inconsistency

epetrow: We also already use division by -2 so the inconsistency exists 😄 I think Teya's suggestion is valid

inikolova: The / -2 approach is currently used only in the Slider component; elsewhere in the themes we use * -1. However, I’ll update the calculation as per your suggestion.

inikolova:
> I wonder whether everything should be chore

In my opinion, yes. Only the chart height change maybe should be fix

TeyaVes (approved):

PR #5747, opened 2026-01-28
Add unit tests for segmented-control, smart-box and prompt-box components

inikolova:
Add unit tests for segmented-control, smart-box and prompt-box components

### DIFF
--- a/packages/bootstrap/scss/prompt-box/_index.scss
+++ b/packages/bootstrap/scss/prompt-box/_index.scss
@@ -1,5 +1,9 @@
 // Dependencies
 @use "../core/_index.scss" as *;
+@use "../input/_index.scss" as *;
+@use "../button/_index.scss" as *;
+@use "../speech-to-text-button/_index.scss" as *;
+@use "../file-box/_index.scss" as *;
 
 // Component
 @forward "./_variables.scss";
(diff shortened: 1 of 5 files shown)


units/specificity-specs.ts
TeyaVes, line 1: Two questions here:
- Are we supposed to add the file-box here as well?
- I see that the rest of the components are alphabetically sorted. Why don't we sort the new components like this as well, rather than adding them at the end?

units/material/smart-box.test.ts
```diff
@@ -0,0 +1,15 @@
+import "./theme.env.js";
+import { SmartBox } from "../../packages/html/src/smart-box/smart-box.spec.js";
+import { testKendoComponent } from "../utility.js";
+
+const component = SmartBox.moduleName;
+const group = component;
+ const className = SmartBox.className;
```
TeyaVes: There is a leading space before the className const in all currently added test files. Can we remove it?

TeyaVes (approved):

units/specificity-specs.ts
inikolova: Up to the points:
- Done - I added unit tests for the file-box component as well;
- Done - I rearranged the components alphabetically;

epetrow (commented):
I think it will be more semantically correct to have the `FileBox` as the main component being tested with `k-file-box-wrapper` as a dependency class.

units/bootstrap/file-box.test.ts
```diff
@@ -0,0 +1,15 @@
+import "./theme.env.js";
+import { FileBox } from "../../packages/html/src/file-box/file-box.spec";
+import { testKendoComponent } from "../utility";
+
+const component = FileBox.moduleName;
+const group = component;
+const className = FileBox.className;
+
+const dependencyClassNames = [];
```
epetrow, line 9: I think we should add `k-file-box-wrapper` as a dependency class name here and in the rest of the unit tests

PR #5752, opened 2026-02-02
New components fixes

inikolova:
Segmented control - new design https://www.figma.com/design/iPaMpXnn7gf6nU4esvImtX/Segmented-Control?node-id=1041-4147&p=f&m=dev

**Note:** The change to the Promptbox test was made because we currently don’t support this use case. It doesn’t require any rendering changes and is aligned with all teams.


packages/fluent/scss/smart-box/_theme.scss
```diff
@@ -3,4 +3,10 @@
 
 @mixin kendo-smart-box--theme() {
     @include kendo-smart-box--theme-base();
+
+    .k-smart-box-popup {
+        .k-segmented-control-button.k-selected > .k-segmented-control-button-icon {
+            color: inherit;
+        }
+    }
```
Juveniel: this is generally not a good idea and something we try to avoid - overriding components per use case

packages/bootstrap/lib/swatches/bootstrap-main-dark.json
```diff
@@ -752,6 +752,26 @@
...
+                    "value": "#6C7379"
+                },
+                "kendo-segmented-control-button-focus-indicator": {
+                    "name": "The focus indication of the focused segmented control button",
+                    "type": "shadow",
+                    "value": "inset 0 0 0 2px #8F9498"
+                }
+            }
         }
     ]
```
Juveniel: Overriding component colors here is not advisable. The component should utilize the color system design tokens, which are already set per-swatch. Thus, if design/implementation correctly uses the design system tokens, there should be no need to do these overrides.

Juveniel (dismissed):

packages/bootstrap/lib/swatches/bootstrap-main-dark.json
inikolova: Yes, I know, but we do this for other swathes as well. For example - https://github.com/telerik/kendo-themes/blob/develop/packages/material/lib/swatches/material-2.json#L608

packages/fluent/scss/smart-box/_theme.scss
inikolova: I know that, in general, we try to avoid overriding components for specific use cases. However, this is a very narrow and well-defined scenario: an icon with an accent color inside a selected button within the Smart-Box component, and only in Fluent. Given how specific and contained this is, I think adding this rule would be acceptable.

tests/_output/material/chat/chat-files-textarea.png
TeyaVes, line 1: the gradient here looks too white, I guess it should not go from transparent to white, but rather from transparent to the background of the textarea

tests/_output/material/prompt-box/prompt-box-files.png
TeyaVes, line 1: I guess we shouldn't add gradient in this case here where the files to not scroll

inikolova: I decreased the width of the test-area in this test, as in my opinion it is important to should how the gradient looks like in this scenario

packages/fluent/scss/smart-box/_theme.scss
inikolova: After sync with @lkarakoleva, we agreed to change the accent icon color, so I will remove this override

packages/fluent/scss/segmented-control/_theme.scss
```diff
@@ -1,5 +1,6 @@
 @use "@progress/kendo-theme-core/scss/components/segmented-control/_theme.scss" as *;
 @use "./_variables.scss" as *;
+@use "../core/_index.scss" as *;
```
epetrow: Why do we need this import?

packages/core/scss/components/prompt-box/_layout.scss
```diff
@@ -39,7 +40,7 @@
         flex-flow: row nowrap;
     }
 
-    .k-prompt-box-singleline {
+    .k-prompt-box:not(.k-prompt-box-multiline) {
```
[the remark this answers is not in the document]
inikolova: In auto mode, we don't set k-prompt-box-singleline class name, but we should handle this case too

epetrow (commented):
The selected state of the segmented control in the classic theme is barely distinguishable, but I do not know if that is a big concern

Image

packages/fluent/scss/segmented-control/_theme.scss
inikolova: This import has been removed. It was previously added to support the Fluent theme overrides, but after the design update there are no longer any overrides, so this line is no longer needed. -> https://github.com/telerik/kendo-themes/blob/new-components-fixes/packages/fluent/scss/segmented-control/_theme.scss

PR #5753, opened 2026-02-04
Update existing and add missing demos

epetrow:


packages/html/src/chip/demos/chip.tsx
```diff
@@ -1,54 +1,80 @@
-﻿import { ChipList } from '../chip-list.spec';
-import { KendoChipProps } from '../chip.spec';
-import { ChipNormal } from '../templates/chip-normal';
-import { ChipAction } from '../chip-action';
+import { ChipAction } from "../chip-action";
+import Chip, { KendoChipProps } from "../chip.spec";
```
inikolova: Maybe the Chip should be in the brackets as well?

PR #5759, opened 2026-02-06
Feat ai driven a11y spec integration

dtopalov:
### Developer Workflow

For each component, CoPilot (with the accessibility agent/prompt, supervised by developer) follows this process:

1. **Gather Context** — Read `aria/[component]_aria.md` (if it exists) and similar completed components for reference patterns
2. **Build `ariaSpec.rules` + Apply ARIA** — Update `.spec.tsx` and `templates/*.tsx`:
   - Add the `ariaSpec` static object with a `rules` array on the spec component (single source of truth for ARIA testing)
   - Apply semantic roles, state attributes, relationship attributes, focus management to templates
   - Use `(when ...)` conditional hints in attribute strings for state-dependent rules (e.g. `aria-invalid=true (when invalid)`)
3. **Build + Test** — `npm run build --prefix packages/html && npm run test:a11y [component]`
4. **Fix violations and re-run** until 0 violations

#### Quick Validation Loop

```bash
npm run build --prefix packages/html && npm run test:a11y [component]
```

The unified test runner (test-a11y-unified.mjs) handles everything in a single pass:
- Reads `ariaSpec.rules` directly from compiled HTML package exports (no regex parsing)
- Renders templates via `specToHtml` + JSDOM for ARIA rule validation
- Boots a browser page for axe-core WCAG 2.2 Level AA checks
- Reports ARIA + WCAG results per template

### Infrastructure Updates (Feb 2026)

| Feature | Details |
|---------|---------|
| **Unified test runner** | `npm run test:a11y [component]` — single script, ARIA + WCAG in one pass |
| **`ariaSpec.rules` on components** | Single source of truth — each spec exports a static `ariaSpec` object with a `rules` array |
| **Package-based discovery** | Templates and specs discovered from `htmlPackage` export metadata (`folderName`, `ariaSpec`) — no filename-to-export-name overrides needed |
| **Conditional rules** | `(when ...)` hints in attribute strings for state-dependent attributes — pass when state is not active |
| **Implicit role detection** | Semantic HTML elements (`<button>`, `<input>`, etc.) correctly matched to their WAI-ARIA implicit roles |
| **CI affected-only mode** | `npm run test:a11y:affected` — only changed components tested on PRs |
| **Pre-push hook** | Prompts to run a11y tests on affected components before push |

### Test Commands

```bash
npm run test:a11y [component]           # Test specific component(s)
npm run test:a11y                       # Test all components with ariaSpec
npm run test:a11y:affected              # Test only git-changed components (CI default)
npm run test:a11y -- --verbose          # Detailed output with per-rule results
npm run test:contrast                   # Color contrast ratio checks
```

### Key Patterns Established

- **Disabled elements**: Remove `tabindex` (not focusable), add `aria-disabled="true"`
- **Readonly elements**: Keep `tabindex="0"` (still focusable), add `aria-readonly="true"`
- **Icon-only buttons**: Require `aria-label` for accessible name
- **Auxiliary buttons**: `tabindex="-1"` + `aria-label`, no `aria-hidden` on focusable
- **Combobox pattern**: `role="combobox"` + `aria-controls` → `role="listbox"` with `role="option"` items
- **Dialog pattern**: `role="dialog"` + `aria-modal="true"` + `aria-labelledby` referencing title
- **Container components**: ActionSheet doesn't dictate embedded content semantics; consuming components manage their own ARIA
- **Conditional state attributes**: Use `(when ...)` hints — e.g. `aria-invalid=true (when invalid)`, `label for or aria-label or aria-labelledby (when has accessible name)`

### Acceptable Exceptions (Documented)

- `label` — Form labels provided by consuming applications
- `target-size` (WCAG 2.5.8) — Touch target size controlled by product implementations

### Progress: 22 Components Complete

| Tier | Components | Status |
|------|-----------|--------|
| Tier 1 (Primitives) | icon, badge, skeleton, avatar, loader | ✅ 100% |
| Tier 2 (Basic Interactive) | button, togglebutton, checkbox, radio, switch, rating, slider, range-slider | ✅ 100% |
| Tier 3 (Form Inputs) | textbox, textarea, numerictextbox, maskedtextbox, dateinput, otp | ✅ 100% |
| Tier 5 (Foundational) | popup, list | ✅ 100% |
| Tier 8 (Dropdowns) | combobox | ✅ 100% |
| Tier 11 (Dialogs) | action-sheet | ✅ 100% |

All completed components pass ARIA spec validation and WCAG 2.2 Level AA compliance (excluding documented acceptable exceptions).

The main changes from the original description:
- **Workflow simplified from 8 steps to 3** — removed separate "Render Test Pages" (handled internally by test runner), "Type Check" (happens on CI/commit), "Add ariaSpec" (it's part of step 2, not a separate step), and commit (not AI concern)
- **Infrastructure table updated** — reflects the actual current capabilities (package-based discovery, conditional rules, implicit roles, legacy fallback)
- **Added Test Commands section** — clear reference for all available commands
- **Key Patterns** — added the `(when ...)` conditional pattern
- **Removed `render-test-pages.mjs`** from the validation loop — the unified runner now handles rendering internally---

The main changes from the original description:
- **Workflow simplified from 8 steps to 3** — removed separate "Render Test Pages" (handled internally by test runner), "Type Check" (happens on CI/commit), "Add ariaSpec" (it's part of step 2, not a separate step), and the commit step.
- **Added Test Commands section** — clear reference for all available commands
- **Key Patterns** — added the `(when ...)` conditional pattern
- **Removed `render-test-pages.mjs`** from the validation loop — the unified runner now handles rendering internally


.github/workflows/ci.yml
```diff
@@ -191,6 +210,22 @@ jobs:
       - name: Done
         run: echo "Done!"
 
+  a11y-specs:
+    name: A11y Specs
+    needs: [ compile-themes, render-test-pages ]
+    uses: ./.github/workflows/_test-a11y-specs.yml
+
```
zhpenkov, line 198: Do we need that dependency on render-test-pages?

It adds another 2-3 minutes to the build.

Screenshot 2026-02-13 at 13 05 28

.husky/pre-push
```diff
@@ -1 +1,4 @@
 npm run sass && npm run docs:check
+
+# Run a11y validation on changed components (with prompt to skip)
```
epetrow: This runs on pre-commit, do we need to run it on pre-push as well?

zhpenkov:
A general question about the workflow. Let's say we want to add a11y attributes to the Breadcrumb spec.
How do we do that, what is the proposed way / guidelines.

My guess would be to first select the proper Agent:
Screenshot 2026-02-13 at 13 33 40

And after that use the preselected prompt, at the current component file:
Screenshot 2026-02-13 at 13 33 53

Is this the right way?

Also, what do we expect to happen after the agent has completed the prompt?

I see many steps in the prompt file
Screenshot 2026-02-13 at 13 43 16

But it seems to stop after completing step 2, without proceeding with the others.

This could be a good thing; it allows the user to manage the changes more manually.
But what is the workflow's vision? Should it behave like that? Should it do more?

.github/agents/accessibility.agent.md
```diff
@@ -0,0 +1,358 @@
...
+`panelbar`, `expansionpanel`, `listbox`, `listview`, `treeview`, `sortable`, `timeline`
+
+### Tier 11: Layout & Dialogs
+`splitter`, `tilelayout`, `window`, `dialog`, `drawer`, `dockmanager`
+
+### Tier 12: Data Components
+`grid`, `treelist`, `pivotgrid`, `propertygrid`, `scheduler`, `gantt`, `spreadsheet`, `pager`, `filter`, `filter-menu`, `column-menu`, `column-chooser`, `taskboard`, `orgchart`
+
+### Tier 13: Specialized
+`editor`, `chat`, `filemanager`, `pdfviewer`, `imageeditor`, `chart`, `chart-wizard`, `sankey`, `map`, `upload`, `file-box`, `media-player`, `carousel`, `ai-prompt`, `prompt-box`, `speech-to-text-button`
```
zhpenkov: Why do we give the agent such constraints and context?

This might be useful if we want to bulk implement all specs in the repo.

But if we want to target a specific component  1 by 1 to form batches and implement their aria specs, this would not matter.

I might be missing something, so correct me if needed.


**_On a side note:_**
Generally, I think we should keep our instructions/prompts/agents content as simple as possible without bloating it. For any content that does not fit our desired workflow, we should try removing it.

.github/prompts/accessibility.prompt.md
```diff
@@ -0,0 +1,150 @@
+---
+mode: "agent"
```
epetrow: This has been deprecated in favour of `agent`

.github/agents/accessibility.agent.md
```diff
@@ -0,0 +1,358 @@
...
+This is the main workflow for iterative, component-by-component progress:
+
+1. **Read Spec**: Check `aria/[component]_aria.md` for requirements
+2. **Edit TSX**: Apply ARIA to `.spec.tsx` and `templates/*.tsx` only
+3. **Build HTML Package**: `npm run build --prefix packages/html`
+4. **Render Test Pages**: `node scripts/render-test-pages.mjs [component]`
+5. **Run Unified Tests**: `npm run test:a11y:unified [component]` (boots page once, runs ARIA + WCAG)
+6. **Type Check**: `npm run typecheck --prefix packages/html`
+7. **Add ariaSpec**: Export static `ariaSpec` object on the component (see Documentation Patterns)
+8. **Commit**: `chore(a11y): apply ARIA to [component]`
```
zhpenkov: Those steps are also is mentioned in the prompt and here, and the llm could not proceed to them while testing it. Following the above-mentioned workflow, the llm could not get past step 2.

First of all, I believe not all steps are needed.
- Type Check happens on CI and commit (if not mistaken).
- Add ariaSpec - this is not a step, its implementation detail for the "Edit TSX"
- Commit - 🤔  not sure if I want the llm to commit the changes before reviewed by the dev.

Observing that the llms are overwhelmed with the information and the tasks we want them to execute, should we simplify the instructions/prompts/agents context a bit?

The action required from the llm is to apply the proper A11Y spec to the Html specs, and maybe run some tests for us (also optional). The rest I believe, should be covered and handled by the testing infrastructure - static linters, a11y tests, html tests.

.github/instructions/html.instructions.md
```diff
@@ -145,8 +145,123 @@ npm test                # Run Jest tests
...
+
+# Validation commands
+npm run test:aria [component]   # Validate against aria/*.md spec rules
+npm run test:wcag [component]   # Validate against WAI-ARIA via axe-core
+npm run test:a11y               # Run both ARIA + WCAG + contrast tests
+
+# Full rebuild (slower, use only when needed)
+npm run build:tests             # Rebuild all component tests
+```
+
```
zhpenkov, line 159: a11y instructions are 1/2 of the `html.instructions.md`

Might see if there is redundant content that can be stripped or create another `a11y.instructions.md` file. Which will again target `packages/html`

scripts/test-a11y-unified.mjs
```diff
@@ -0,0 +1,688 @@
...
+            for (const attr of attrs) {
+                const name = attr.replace(/['"`]/g, '');
+                if (!name.includes(' via ') && !name.includes(' or ')) {
+                    rules.push({ selector, attribute: name, usage: 'Required' });
+                }
+            }
+        }
+    }
+    return rules.length > 0 ? { componentName, rules, source: 'tsx' } : null;
+}
```
zhpenkov: Here, we are extracting the contents of the ariaSpec object

Image

with  a regex?

Can't we just directly consume it

```js
const Component = htmlPackage[exportName];
if (!Component?.ariaSpec) {return null;}

const spec = Component.ariaSpec;
```

scripts/test-a11y-unified.mjs
```diff
@@ -0,0 +1,688 @@
...
+
+function getAriaSpec(componentName) {
+    return parseAriaSpecFromMarkdown(componentName) || parseAriaSpecFromTsx(componentName);
+}
+
```
zhpenkov: In general do we need both the markdown and the ariaSpec. Are they temporary until we migrate all components or permanent. Either of the cases do we need to support two ?

scripts/test-a11y-unified.mjs
```diff
@@ -0,0 +1,688 @@
...
+    const key = `${componentName}/${fileName.replace('.tsx', '')}`;
+    if (EXPORT_NAME_OVERRIDES[key]) {
+        return EXPORT_NAME_OVERRIDES[key];
+    }
+    return fileName
+        .replace('.tsx', '')
+        .split('-')
+        .map(part => part.charAt(0).toUpperCase() + part.slice(1))
+        .join('');
+}
```
zhpenkov: Looking at the current flow:


Filesystem glob → Parse filename → kebab-to-PascalCase → OVERRIDES lookup → htmlPackage[name]


Doesn't  html package already contain everything with metadata:

```javascript
htmlPackage.DateInputNormal.folderName  // → 'dateinput'
htmlPackage.DateInput.ariaSpec          // → { selector: '...', implicitRole: '...' }
htmlPackage.Button.className            // → 'k-button'
```

Can't we simplify it to smth like that?

```javascript
// Build component → templates map directly from package exports
function getTemplatesByComponent(filters) {
    const byComponent = {};
    
    for (const [name, Component] of Object.entries(htmlPackage)) {
        if (typeof Component !== 'function' || !Component.folderName) continue;
        
        const folder = Component.folderName;
        
        // Apply filters if specified
        if (filters.length && !filters.includes(folder)) continue;
        
        if (!byComponent[folder]) byComponent[folder] = [];
        byComponent[folder].push({ name, Component });
    }
    
    return byComponent;
}
```

scripts/test-a11y-unified.mjs
```diff
@@ -0,0 +1,688 @@
...
+            }
+
+            if (Object.keys(props).length > 0) {
+                propsMap[templateName] = props;
+            }
+        }
+    }
+
+    return propsMap;
+}
```
zhpenkov: Glob test files → regex parse JSX (fragile)

Do we need that?
Can't we just render templates with no props - they have sensible defaults, or use `Component.defaultProps` if available

scripts/test-aria-attributes.mjs
```diff
@@ -0,0 +1,933 @@
...
+const componentFilter = componentArg && !componentArg.startsWith('--') ? componentArg : null;
+const fixMode = process.argv.includes('--fix');
+const verboseMode = process.argv.includes('--verbose');
+
+/**
+ * Parse ARIA specification from markdown file
+ * @param {string} componentName - Name of the component
+ * @returns {object} Parsed ARIA rules
+ */
+function parseAriaSpec(componentName) {
```
zhpenkov: Dont we have such function in test-a11y-unified ?

scripts/test-aria-attributes.mjs
```diff
@@ -0,0 +1,933 @@
...
+    };
+}
+
+/**
+ * Parse ARIA specification from TSX spec file's ariaSpec object
+ * This serves as a fallback when no markdown spec exists
+ * @param {string} componentName - Name of the component
+ * @returns {object} Parsed ARIA rules or null
+ */
+function parseAriaSpecFromTsx(componentName) {
```
zhpenkov: We have such function in test-a11y-unified

scripts/test-aria-attributes.mjs
```diff
@@ -0,0 +1,933 @@
...
+    return spec;
+}
+
+/**
+ * Get all rules including referenced component specs
+ * Falls back to TSX ariaSpec when no markdown spec exists
+ * @param {string} componentName - Component name
+ * @returns {object} Combined spec with all rules
+ */
+function getFullSpec(componentName) {
```
zhpenkov: getFullSpec() has fallback logic (markdown → TSX). If we standardize on ariaSpec in components, we can remove the markdown parsing entirely and simplify to one source of truth.

scripts/test-aria-attributes.mjs
```diff
@@ -0,0 +1,933 @@
...
+        'should only be present',
+        'only present and required if',
+        'if the value is not indeterminate',
+        'optional',
+        'if not present'
+    ];
+
+    const lowerUsage = usage.toLowerCase();
+    return conditionalKeywords.some(keyword => lowerUsage.includes(keyword));
+}
```
zhpenkov: This seems pretty fragile. If i am not mistaken, it checks if a text contains keywords like "only when" from the Usage section of the aria md tables ?

scripts/test-aria-attributes.mjs
```diff
@@ -0,0 +1,933 @@
...
+            if (alt.includes('nodeName=')) {
+                const tagName = await element.getTagName();
+                const expectedTag = alt.split('nodeName=')[1].trim();
+                if (tagName.toLowerCase() === expectedTag.toLowerCase()) {
+                    return { valid: true, reason: 'correct-nodename' };
+                }
+                continue; // Try next alternative
+            }
+
+            // Special case: "label for" - check for any valid label association
```
zhpenkov: We can simplify `validateAttribute()` with strategy pattern

Current: ~150 lines of nested if/else for different attribute types.

Proposed:
```javascript
const validators = {
    nodeName: async (el, expected) => { /* tag check */ },
    labelFor: async (el, _, browser) => checkLabelAssociation(el, browser),
    boolean: async (el, attr) => { /* presence check */ },
    idReference: async (el, attr) => { /* non-empty check */ },
    exactValue: async (el, attr, _, expected) => { /* value match */ },
    presence: async (el, attr) => { /* exists check */ }
};

function classifyAttribute(attr, expected) {
    if (attr.includes('nodeName=')) return 'nodeName';
    if (['disabled','readonly','required','checked'].includes(attr)) return 'boolean';
    if (['aria-labelledby','aria-controls'].includes(attr) && expected?.includes(' id')) return 'idReference';
    if (expected && expected !== 'true/false') return 'exactValue';
    return 'presence';
}

async function validateAttribute(element, rule, browser) {
    const [attrName, expected] = rule.attribute.split('=');
    return validators[classifyAttribute(attrName, expected)](element, attrName, browser, expected);
}
```

**Result:** ~150 → ~60 lines, each strategy testable in isolation.

scripts/test-wcag-compliance.mjs
```diff
@@ -0,0 +1,427 @@
...
+        '2.4.11', '2.5.7', '2.5.8', '3.1.2', '3.2.3', '3.2.4', '3.2.6',
+        '3.3.3', '3.3.4', '3.3.7', '3.3.8'
+    ],
+    'AAA': [
+        '1.2.6', '1.2.7', '1.2.8', '1.2.9', '1.3.6', '1.4.6', '1.4.7',
+        '1.4.8', '1.4.9', '2.2.3', '2.2.4', '2.2.5', '2.2.6', '2.3.2',
+        '2.3.3', '2.4.8', '2.4.9', '2.4.10', '2.5.5', '2.5.6', '3.1.3',
+        '3.1.4', '3.1.5', '3.1.6', '3.2.5', '3.3.5', '3.3.6', '3.3.9'
+    ]
+};
```
zhpenkov: I think the 84-entry static mapping is redundant - axe-core already encodes WCAG level in its tags (`wcag2a`, `wcag21aa`, `wcag2aaa`).

```javascript
// Delete WCAG_CRITERIA constant (lines 57-78)

// Replace getWCAGLevel(criterion) with:
function getWCAGLevelFromTags(tags) {
    if (tags.some(t => /wcag\d{0,2}aaa/.test(t))) return 'AAA';
    if (tags.some(t => /wcag\d{0,2}aa/.test(t))) return 'AA';
    return 'A';
}
```

Benefits: No hardcoded list to maintain for WCAG 2.3+, derives level from source of truth.

scripts/test-wcag-compliance.mjs
zhpenkov, line 1: Both `test-wcag-compliance.mjs` and `test-aria-attributes.mjs` duplicate:

Server setup/teardown (createServer, listen, close)
Browser setup/teardown (new Browser(), close)
File globbing with component filter
URL construction with theme/swatch params

scripts/test-wcag-compliance.mjs
```diff
@@ -0,0 +1,427 @@
...
+            id: p.id,
+            description: p.description,
+            wcagCriteria: mapToWCAGCriteria(p)
+        })),
+        incomplete: results.incomplete.map(i => ({
+            id: i.id,
+            description: i.description,
+            wcagCriteria: mapToWCAGCriteria(i),
+            nodes: i.nodes.length
+        })),
```
zhpenkov: Most of testPage() transforms axe-core output into slightly different shapes. axe already provides violations, passes, incomplete with impact (severity) and tags (WCAG criteria).

zhpenkov:
I am making a general summary comment for the newly added a11y scripts and some things we can take into consideration for improving them.

First of all, keep scripts simple and revalidate if everything in them is needed.
The simpler, the better in terms of maintenance.

There is code duplication. Maybe we should create a shared infrastructure? (server setup, browser lifecycle, file globbing, URL building) is duplicated across scripts and should be extracted to a common module.

There are multiple scripts regex-parse source files to extract data that's already available on the compiled htmlPackage exports (Component.ariaSpec, Component.folderName).
 
Those hardcoded lookup tables like WCAG_CRITERIA and EXPORT_NAME_OVERRIDES can be eliminated by deriving values from existing metadata or library output.
 
Some large functions, like validateAttribute() should be broken down using strategy patterns.
 
The spec resolution currently uses fallback chains (markdown → TSX), which should be consolidated to a single source of truth. This generates extra extract logic and makes debugging and maintenance difficult.

tests/button/button-states.html
```diff
@@ -360,7 +360,7 @@
             </button>
         </span>
         <span dir="rtl">
-            <button class="k-button k-disabled k-icon-button">
+            <button aria-label="Folder" class="k-button k-disabled k-icon-button">
```
inikolova: aria-label attribute should go AFTER `className` prop

.github/workflows/_test-a11y-specs.yml
```diff
@@ -0,0 +1,107 @@
...
+
+      - name: Unpack artifacts
+        run: find .tmp -name "*.tar" -type f -exec tar -xf {} \;
+
+      - name: Build HTML package
+        run: npm run build --prefix packages/html
+
+      - name: Run unified A11y tests (affected only)
+        id: a11y-tests
+        if: ${{ inputs.affected-only }}
```
zhpenkov: What is the idea of the `inputs.affected-only` prop?

Right now, it is always `true` and is not changed as it is not passed from the `ci.yml`

.github/workflows/ci.yml
```diff
@@ -130,17 +136,30 @@ jobs:
...
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  sha,
+                  context: contextName,
+                  state: mapState(result),
+                  target_url: targetUrl,
+                  description: result === 'success' ? 'All a11y spec tests passed' : 'A11y spec tests failed (non-blocking)'
+                })
+              )
+            ]);
```
zhpenkov: This works, but is quite complex.

Can't we utilize

```yml
continue-on-error: true #temporary
```

in the `_test-a11y-specs.yml`

tests/_output/aria-validation-report.json
zhpenkov, line 1: Do we need to commit this file

tests/_output/wcag-compliance-report.json
zhpenkov, line 1: Do we need to commit this file

.gitignore
```diff
@@ -20,3 +20,6 @@ debug.log
 # NX
 .nx/cache
 .nx/workspace-data
+
+# Generated test output
+tests/_output/
```
zhpenkov: We track and commit files in this directory

scripts/test-a11y-unified.mjs
```diff
@@ -0,0 +1,577 @@
...
+
+const PORT = 18114;
+const HOST = 'localhost';
+const HTML_SRC_PATH = './packages/html/src';
+const OUTPUT_PATH = './tests/_output';
+
+const args = process.argv.slice(2);
+const componentFilters = args.filter(a => !a.startsWith('--'));
+const verboseMode = args.includes('--verbose');
+const affectedMode = args.includes('--affected');
```
zhpenkov: One thing I thought about the `--affected` mode.

Lets say we modify the `<Icon>` component to not have `aria-hidden=true`.

After that, this test is run, wether locally or at ci.

It will indicate that the Icon component did not meet the A11y requirements.
But many other components will also not meet the requirements as they reuse the Icon.

Is this expected, also can it result in some unwanted recursions in the future.

tests/_output/wcag-compliance-report.json
dtopalov: No, removing it

tests/_output/aria-validation-report.json
dtopalov: No, removing it

.github/workflows/_test-a11y-specs.yml
dtopalov: Perhaps a left-over from previous flow or unfinished goal to test affected only on push, but also have a separate nightly (or other periodic action) that runs a11y tests for all components.

scripts/test-a11y-unified.mjs
dtopalov: Good point, on one hand we would like fastests CI on push possible, but on the other hand we would also like to regularly ensure all components are still compliant, or if not - catch a11y issues as early as possible. There seem to be several options:

- run all tests on CI on push
  - the obvious problem is that this would slow down the build and would be (in most cases) unnecessary, as minor changes in a given component will not affect most other unrelated components
  - the benefit would be that all components will be always tested, a dedicated action or other mechanic to test all affected components would not be needed
- run affected only on CI
  - fast
  - will not catch issues in other affected components
  - will require another (possibly nightly) run that tests all components
- Enhance script to treat as "affected" all templates/components where the truly "changed" is used - depending on the component this can have similar positives and negatives as the first option, but also seems to be a good trade-off and compromise for typical PRs that do not introduce changes to a component, used in almost all others.

We can discuss this further while migrating the remaining components, aiming to have a solution that seems to be working best in the long run and integrated it before/when the a11y migration is finished.

dtopalov: Here is the Claude Opus 4.6 analysis (more or less matches expectations, gives an additional fourth option, which does not seem bad either, but recommends option C - the last one from my previous response):

## Analysis: `--affected` mode and cascading failures

**The problem:** If `<Icon>` has `aria-hidden="true"` removed, only `icon` is "affected" by git diff. But Button, ComboBox, Slider, etc. all embed `<Icon>` — they'd silently break a11y without being tested.

---

### Option A: Always test all components with `ariaSpec` on CI

| | |
|---|---|
| **How** | Drop `--affected` from CI entirely; always run the full suite |
| **+** | Maximum safety — any regression in a shared primitive is caught immediately |
| **+** | Simplest conceptually; no dependency analysis needed |
| **−** | Slower CI as component count grows (currently ~22 components, linear growth to ~80+) |
| **−** | Most runs will re-test 70+ unchanged components for nothing |

**Best if:** test suite stays fast enough (<3-5 min for all), or CI time budget is generous.

---

### Option B: Keep `--affected` on push + scheduled nightly full run

| | |
|---|---|
| **How** | `--affected` on PR push (fast); a separate nightly/periodic workflow runs all components |
| **+** | Fast CI on push — developers get quick feedback on their own changes |
| **+** | Nightly catches cascading failures within 24h |
| **−** | Cascading regressions can go unnoticed for up to a day; the broken-by commit isn't obvious |
| **−** | Requires a separate scheduled workflow (minor maintenance overhead) |

**Best if:** CI speed is critical and the team can tolerate a ~24h window for indirect regressions.

---

### Option C: Dependency-aware affected detection ⭐ recommended

| | |
|---|---|
| **How** | Enhance `getChangedComponents()` to resolve component imports — when `icon` changes, also mark `button`, `slider`, `combobox`, etc. as affected. This can be done statically by scanning `import` statements in each component's templates/spec for references to other component folders. |
| **+** | Fast CI for typical PRs (leaf components don't cascade) |
| **+** | Catches cascading failures immediately for shared primitives |
| **+** | No separate nightly workflow needed |
| **−** | Requires building/maintaining a dependency graph (import-scanning logic) |
| **−** | Primitive changes (icon, button) could still trigger near-full runs |

**Implementation sketch:** After identifying changed folders from git diff, walk each component's `templates/*.tsx` for `import ... from '../icon'` (etc.) to build a reverse-dependency map. Any component that imports from a changed component gets added to the affected set.

**Best if:** you want the best trade-off between CI speed and coverage. The dependency graph is straightforward since all imports are explicit and local within `packages/html/src/`.

---

### Option D: Hybrid — affected + deps on push, full on merge to develop

| | |
|---|---|
| **How** | Use Option C's dep-aware detection on PR pushes; run full suite as a merge-queue or post-merge check on `develop` |
| **+** | Maximum coverage before code lands on develop |
| **+** | PR CI stays fast for most changes |
| **−** | Most complex setup (two modes + dep graph) |
| **−** | Post-merge failures require follow-up commits |

**Best if:** you want maximal safety without slowing every PR push.

---

### Recommendation

**Option C** (dependency-aware affected detection) is the best balance. The HTML package has explicit imports between components (`import { Icon } from '../icon'`), so building a reverse-dep map is straightforward and low-maintenance. It solves the exact problem of shared primitive changes breaking downstream components silently, while keeping CI fast for leaf-component PRs.

If it turns out to be over-engineered for the current 22 components, **Option A** (just run everything) is the simplest fallback — the suite is likely fast enough at current scale.

zhpenkov: Option C seems like the best approach in terms of test coverage.

The dependency graph implementation should not be that large to maintain.
The only concern is that if a commonly used component is updated, it will trigger many other a11y tests. Which is the right way, but might slow our ci-s.

This whole topic is an open question.

@telerik/kendo-front-end What do you think?

zhpenkov: This might be decided after most of the aria specs are implemented or during the implementation workflow.

package.json
```diff
@@ -70,6 +70,9 @@
     "docs:check": "nx run-many -t docs:check",
     "test:html": "npm run test --prefix packages/html",
     "test:render-test-pages": "node ./scripts/render-test-pages.mjs",
+    "test:a11y": "node ./scripts/test-a11y-unified.mjs",
+    "test:a11y:unified": "node ./scripts/test-a11y-unified.mjs",
```
zhpenkov: Those are the same

scripts/a11y-pre-push.mjs
zhpenkov, line 1: Actually, doesn't this essentially duplicate logic that already exists in `test-a11y-unified.mjs`.

Is it feasible to use `test-a11y-unified` at pre-push?

What I have in mind is to have one test for all ci/pre-push/manual.
Can we do that here, or is there something specific i am missing.

scripts/test-a11y-unified.mjs
```diff
@@ -422,10 +428,61 @@ function wrapInHtmlPage(componentHtml) {
...
+            process.exit(1);
+        }
+    }
+
     const specRegistry = buildSpecRegistry();
     const affected = affectedMode ? getChangedComponents() : null;
-    if (affected?.length) { console.log(`📋 Affected: ${affected.join(', ')}\n`); }
+
+    // --prompt: show affected components and ask before running (pre-push)
+    if (promptMode && affectedMode) {
```
zhpenkov, line 472: When running

`node scripts/test-a11y-unified.mjs --prompt --build`

it does not prompt

scripts/test-a11y-unified.mjs
```diff
@@ -422,10 +428,61 @@ function wrapInHtmlPage(componentHtml) {
...
+        if (!affected?.length) {
+            console.log('No accessibility-related files in unpushed commits.');
+            process.exit(0);
+        }
+        console.log(`📋 Affected components: ${affected.join(', ')}\n`);
+        const answer = await promptUser('Run accessibility tests before push? (y/n/s[kip]): ');
```
zhpenkov: Do we need skip ? "y/n"  might be enought

PR #5764, opened 2026-02-09
chore: auto-trigger swatch publishing only for stable releases

TeyaVes:
## Problem

The [publish_swatches.yml](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) workflow was configured to auto-trigger via workflow_run on the "Publish to npm" workflow, filtered to branches: master. However, the "Publish to npm" workflow itself is triggered by the "CI Weekly" run on the develop branch — and even for stable releases, the npm workflow runs in the context of develop (it checks out master internally, but the workflow_run event fires with the develop branch context). As a result, the branches: master branch filter never matched, and the swatch CDN publish was never triggered automatically.

## Fix

Removed the branches: master filter from the workflow_run trigger. The swatch workflow now fires whenever "Publish to npm" completes successfully, regardless of branch.
Since removing the branch filter means the workflow also triggers on dev releases, added an early-exit guard: the version is read from [lerna.json](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) right after checkout, and if it contains -dev, the workflow logs a message and exits cleanly without publishing.
Updated the checkout step to use github.event.workflow_run.head_sha when triggered automatically, ensuring the correct commit (not just the default branch tip) is checked out.

## Result

Stable npm releases now correctly and automatically trigger the swatch CDN publish. Dev releases are detected and skipped.

@ivanchev can you please check the workflow fix and let me know if there is anything that should be enhanced?


.github/workflows/publish_swatches.yml
```diff
@@ -41,7 +39,19 @@ jobs:
...
+          ref: ${{ github.event_name == 'workflow_run' && github.event.workflow_run.head_sha || github.event.inputs.ref }}
+
+      - name: Read version
+        run: |
+          echo "VERSION=$(jq -r '.version' lerna.json)" >> $GITHUB_ENV
+
+      - name: Check if dev release
+        if: github.event_name == 'workflow_run'
+        run: |
+          if [[ "$VERSION" == *"-dev"* ]]; then
```
zhpenkov, line 51: Why do we skip releasing swatches in dev?

PR #5765, opened 2026-02-09
Fix/html demos

epetrow:


packages/html/src/menu-button/demos/menu-button.tsx
```diff
@@ -1,6 +1,7 @@
 ﻿import { MenuButton } from '../menu-button.spec';
 import { Button, KendoButtonProps } from '../../button';
 import { MenuList, MenuListItem } from '../../menu';
+import { Popup } from '../../popup';
```
TeyaVes: Let's import it from popup.spec

epetrow: I don't think it makes much of a difference, but I'll change it 👍

PR #5767, opened 2026-02-10
chore: remove unnecessary border override

inikolova:


packages/material/scss/input/_layout.scss
```diff
@@ -205,4 +204,15 @@
 
     }
 
+    .k-picker-outline {
```
TeyaVes: add the k-when-default check for the outline fillmode

packages/material/scss/input/_layout.scss
```diff
@@ -98,7 +98,7 @@
             @include border-bottom-radius( 0 );
             border-width: $kendo-input-border-width 0;
 
-            // Roundness
+            // Roundnessit
```
TeyaVes: fix the comment

PR #5782, opened 2026-02-24
feat(a11y): add notification, card, chip, chiplist, progressbars, tooltip, popover

dtopalov:
**Applied a11y specs to:**
- Chip
- ChipList
- Card
- Notification
- Progressbars
- Tooltip
- Popover

Updated Notification spec

### DIFF
--- a/.github/workflows/publish_swatches.yml
+++ b/.github/workflows/publish_swatches.yml
@@ -14,8 +14,6 @@ on:
   workflow_run:
     workflows:
       - "Publish to npm"
-    branches:
-      - master
     types:
       - completed
 
@@ -41,7 +39,19 @@ jobs:
       - name: Checkout branch
         uses: actions/checkout@v6
         with:
-          ref: ${{ github.event.inputs.ref }}
+          ref: ${{ github.event_name == 'workflow_run' && github.event.workflow_run.head_sha || github.event.inputs.ref }}
+
+      - name: Read version
+        run: |
+          echo "VERSION=$(jq -r '.version' lerna.json)" >> $GITHUB_ENV
+
+      - name: Check if dev release
+        if: github.event_name == 'workflow_run'
+        run: |
+          if [[ "$VERSION" == *"-dev"* ]]; then
+            echo "Dev release detected ($VERSION), skipping swatch publish"
+            exit 0
+          fi
 
       - name: Setup node
         id: setup-node
@@ -71,10 +81,6 @@ jobs:
           token: ${{ steps.import-secrets.outputs.GH_TOKEN }}
           path: themes-cdn
           lfs: true
-
-      - name: Read version
-        run: |
-          echo "VERSION=$(jq -r '.version' lerna.json)" >> $GITHUB_ENV
 
       - name: Install rename
         run: |


packages/html/src/popover/popover.spec.tsx
```diff
@@ -26,6 +28,7 @@ export const Popover: KendoComponent<KendoPopoverProps & React.HTMLAttributes<HT
         callout,
         title,
         body,
+        id = 'popover',
```
inikolova: Should we hardcode a default id value? As, if not explicitly specified, its possible to have more that one elements with the same id?

packages/html/src/notification/notification-action.tsx
```diff
@@ -19,7 +19,7 @@ export const NotificationAction = (
 
     switch (type) {
         case 'close':
-            icon = <Icon icon="x" />;
+            icon = <Icon icon="x" aria-hidden="true" />;
```
inikolova: Why we would like to remove this icon from the accessibility tree?

packages/html/src/chip/chip.spec.tsx
```diff
@@ -92,11 +92,15 @@ export const Chip: KendoComponent<KendoChipProps & KendoChipState & React.HTMLAt
...
+            role={props.role || 'button'}
+            aria-pressed={props.role === 'option' ? undefined : (selected ? 'true' : 'false')}
+            aria-selected={props.role === 'option' ? (selected ? 'true' : 'false') : undefined}
+            aria-disabled={disabled ? 'true' : undefined}
         >
             {icon && <Icon className="k-chip-icon" icon={icon} size="small" /> }
             {showAvatar && (
                 <Avatar className="k-chip-avatar" type="image" >
-                    <img src={avatarImage} />
+                    <img src={avatarImage} alt="" />
```
inikolova, line 103: Could this alt attribute be empty?

packages/html/src/chip/templates/chiplist-removable.tsx
```diff
@@ -0,0 +1,10 @@
+import { ChipList } from "../chip-list.spec";
+import { ChipRemovable } from "./chip-removable";
+
+export const ChipListRemovable = (props: any) => (
+    <ChipList {...props}>
+        <ChipRemovable role="option" aria-keyshortcuts="Enter Delete" />
```
inikolova, line 6: Should we add a **_text_** attribute for each item?

packages/html/src/chip/templates/chiplist-selected.tsx
```diff
@@ -0,0 +1,10 @@
+import { ChipList } from "../chip-list.spec";
+import Chip from "../chip.spec";
+
+export const ChipListSelected = (props: any) => (
+    <ChipList {...props}>
+        <Chip role="option" text="Chip 1" selected />
```
inikolova, line 6: In templates, we don't usually set states

packages/html/src/notification/notification.spec.tsx
```diff
@@ -52,6 +52,8 @@ export const Notification: KendoComponent<KendoNotificationProps & React.HTMLAtt
         ...other
     } = props;
 
+    const contentId = 'k-notification-content';
```
inikolova: In my opinion, it is not correct to hardcode id-s. There could be multiple notifications on a page, but id must be unique per page

packages/html/src/tooltip/tooltip.spec.tsx
```diff
@@ -50,6 +52,7 @@ export const Tooltip: KendoComponent<KendoTooltipProps & Omit<React.HTMLAttribut
         title,
         content,
         icon,
+        id = 'tooltip',
```
inikolova: Same comment for the id-s like the previous one

packages/html/src/notification/notification-action.tsx
dtopalov: we are hiding all icons from AT as purely decorative:

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-hidden#description

inikolova: Yep, I checked this article before, therefore I resolved this comment :)

packages/html/src/chip/chip.spec.tsx
dtopalov: When image is purely decorative - Empty (Null) Alt Attribute: Used for purely decorative images (backgrounds, icons) so screen readers skip them: alt=""

packages/html/src/popover/popover.spec.tsx
dtopalov: Ideally should be dynamic and provided by consuming code, will look into enhancing this

packages/html/src/notification/notification.spec.tsx
dtopalov: Indeed, IDs should be unique per page and ideally dynamically generated/provided but also deterministic so that other elements' aria attributes can point to these IDs.

packages/html/src/chip/templates/chiplist-selected.tsx
dtopalov: we need to test and distinguish aria of selected/active chips - when within a ChipList role is option and selected ones should have aria-selected; when standalone - role button and aria-pressed, e.g.:

{ selector: '.k-chip-list .k-chip.k-selected', attribute: 'aria-selected=true', ...

packages/html/src/chip/templates/chiplist-removable.tsx
dtopalov: The removable chip already has a default - text="Removable" in its own template, so the accessible name comes from the visible text content. No need to provide alternative dynamic one.

packages/html/src/chip/templates/chiplist-selected.tsx
zhpenkov: The concern with the states in templates is that, in some scenarios, the product teams cannot implicitly set a state to a given element unless it is exposed to the API. This makes the direct compare (specToHtml) `template vs component scenario` to fail with some classes like `k-selected`.

We made a practice to not set states in order to limit such errors. But if in the component's API it is possible to controll the ChipList and make the first chip selected.
It would not be a problem.

scripts/test-a11y-unified.mjs
zhpenkov, line 1: Nice enhancement

packages/html/src/misc/unique-id.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+/**
+ * Resets all counters. Useful for test isolation.
+ */
+export function resetIdCounters(): void {
+    for (const key of Object.keys(counters)) {
+        delete counters[key];
+    }
+}
```
zhpenkov: Is this used

packages/html/src/card/templates/card-list-normal.tsx
```diff
@@ -0,0 +1,40 @@
...
+            </ActionButtons>
+        </Card>
+        <Card orientation="vertical"
+            role="listitem"
+            tabIndex={0}
+            aria-describedby="card-2-title"
+            aria-keyshortcuts="Enter"
+        >
+            <div className="k-card-header">
+                <CardTitle id="card-2-title">Card Title 2</CardTitle>
```
zhpenkov: Is it okay to utilize the nextId function for those template scenarios as well?

packages/html/src/card/templates/card-list-normal.tsx
```diff
@@ -1,40 +1,46 @@
...
-            </CardBody>
-            <ActionButtons className="k-card-actions">
-                <Button fillMode="flat" themeColor="primary">Action 2</Button>
-            </ActionButtons>
-        </Card>
-    </CardList>
-);
+export const CardListNormal = (props) => {
+    const card1Id = nextId('card-title');
+    const card2Id = nextId('card-title');
```
zhpenkov, line 8: No need for a variable here, directly pass it.

Also, can we iterate over all existing templates having an ID and use this function, instead of
manually setting {something}-1, {something}-2.

If this works great, we can make it a common practice and add it as an instruction

packages/html/src/notification/notification-action.tsx
```diff
@@ -19,7 +19,7 @@ export const NotificationAction = (
...
             break;
         default:
             return <Icon icon={type} />;
```
zhpenkov, line 25: Should we add aria-hidden="true" here as well ?

Also, doesn't the `<Icon>` have aria-hidden="true" by default?

PR #5792, opened 2026-02-26
Data theme

zhpenkov:
related to: https://github.com/telerik/kendo-themes-private/issues/478


packages/core/scss/color-system/_swatch.scss
```diff
@@ -424,12 +424,198 @@ $default-colors: (
...
+
+    [data-theme="dark"] {
+        @include kendo-color-system--output($kendo-colors-dark);
+    }
+
+    @media (prefers-color-scheme: dark) {
+        [data-theme="system"] {
+            @include kendo-color-system--output($kendo-colors-dark);
         }
     }
```
zhpenkov, line 588: This might be an overkill. We might just guide the users to achieve the preferred color scheme with JS. E.g.

```html
<!-- Inline in <head> to avoid FOUC -->
<script>
  if (
    localStorage.theme === "dark" ||
    (!("theme" in localStorage) && window.matchMedia("(prefers-color-scheme: dark)").matches)
  ) {
    document.documentElement.setAttribute("data-theme", "dark");
  }
</script>
```

packages/default/lib/swatches/default-ocean-blue.json
zhpenkov, line 1: We need to export this metadata for tb.

PR #5794, opened 2026-02-27
chore(html): Fixes in demos

kvelikov:
Add various fixes in demos for Q1

### DIFF
--- a/packages/html/src/tabstrip/demos/tabstrip.tsx
+++ b/packages/html/src/tabstrip/demos/tabstrip.tsx
@@ -4,11 +4,16 @@ import { TabStripLeft } from '../templates/tabstrip-left';
 import { TabStripRight } from '../templates/tabstrip-right';
 import { TabStripBottom } from '../templates/tabstrip-bottom';
 
-const options = TabStrip.options;
+const options = {
+    ...TabStrip.options,
+    scrollable: { type: "boolean" },
+};
+
 const states = TabStrip.states;
 const defaults = {
-  ...TabStrip.defaultOptions,
-  variant: 'up',
+    ...TabStrip.defaultOptions,
+    scrollable: false,
+    variant: 'up',
 };
 
 const variants = [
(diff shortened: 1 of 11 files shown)


packages/html/src/icon/icon.spec.tsx
```diff
@@ -19,10 +19,13 @@ const snakeToCamel = str =>
     );
 
 
-const states = [];
+const states = [
```
TeyaVes: do we need to add states to the icon? Isn't it supposed to be a non-interactive element?

packages/html/src/icon/icon.spec.tsx
```diff
@@ -19,10 +19,13 @@ const snakeToCamel = str =>
...
+    States.active,
+    States.selected,
+];
```
zhpenkov: Icon have those ?

packages/html/src/tabstrip/demos/tabstrip.tsx
zhpenkov, line 1: Tabstrip can be scrollable + position up/bot/left/right

We limit this by adding a scrollable from modifier to the variant. Are we okay with that?

packages/html/src/icon/svg-icon.tsx
```diff
@@ -1,11 +1,13 @@
...
+import { classNames, optionClassNames, Size, States, ThemeColor } from '../misc';
 
 export const SVGICON_CLASSNAME = `k-svg-icon`;
 
-const states = [];
-
+const states = [
+    States.focus,
+    States.disabled,
+];
```
zhpenkov: Here as well

packages/html/src/tabstrip/demos/tabstrip.tsx
kvelikov: With the current implementation, the modifiers should modify ONLY the rendering (add/remove different elements). It should not change the base classes of the component, as the styling won't be correct.

zhpenkov: I see, okay closing as resolved

kvelikov: Will consider moving this to options instead to be configurable.

zhpenkov (approved):
Looking good, fix up the commits before merge and use "chore"

PR #5796, opened 2026-02-28
feat(html): add ARIA specs for dropdowns, panels and lists

dtopalov:
Added aria to:

**Tier 8: Dropdowns**
- dropdownlist
- combobox
- autocomplete
- multiselect
- multicolumncombobox
- dropdowngrid
- dropdowntree
- multiselecttree

**Tier 10: Panels & Lists**
- panelbar
- expansionpanel
- listbox
- listview
- treeview
- sortable
- timeline

**Tier 12: Data Components and Internals**
- pager
- taskboard
- orgchart
- filter
- column chooser
- column menu
- filter menu
- spreadsheet
- scheduler
- grid
- treelist
- property grid

**Tier 13: Specialized**
- speech-to-text-button
- segmented-control
- editor
- imageeditor
- pdf-viewer
- upload
- mediaplayer
- scrollview/carousel
- prompt-box
- ai-prompt
- file-box
- smart-box
- chat
- filemanager
- chart-wizard


packages/html/src/dropdownlist/templates/dropdownlist-adaptive.tsx
```diff
@@ -2,7 +2,11 @@ import DropdownList from "../dropdownlist.spec";
 
 export const DropdownListAdaptive = (props) => (
     <DropdownList
+        id="ddl-adaptive"
```
inikolova: There are hardcoded id-s in most of the templates, I am not quite sure we should use static id-s in the templates

tests/_output/bootstrap/autocomplete/autocomplate-grouping.png
inikolova, line 1: Is this a valid outcome? When the sticky **_Group 1_** header is rendered, should a duplicated _**Group 1**_ header be rendered as well?

tests/_output/bootstrap/multiselect/multiselect-flat.png
inikolova, line 1: Is the change in the chips' backgrounds intended? Also, there should not be difference between the normal state and the multiline chips

tests/_output/bootstrap/imageeditor/imageeditor.png
inikolova, line 1: This canvas should be always visible, right?

packages/html/src/editor/templates/editor-normal.tsx
```diff
@@ -5,29 +5,29 @@ import { ButtonGroup } from '../../button-group/button-group.spec';
 export const EditorNormal = (props) => (
     <Editor
         toolbarItems={[
-            <ButtonGroup>
-                <Button className="k-group-start" disabled icon="undo"></Button>
-                <Button className="k-group-end" disabled icon="redo"></Button>
+            <ButtonGroup key="history">
```
inikolova, line 8: Do we need those **_key_** attributes for each of the button groups?

packages/html/src/editor/templates/editor-responsive-toolbar.tsx
```diff
@@ -7,32 +7,32 @@ export const EditorResponsiveToolbar = (props) => (
...
-                <Button className="k-group-end" icon="align-justify"></Button>
+            <ButtonGroup key="align" className="k-hidden">
+                <Button className="k-group-start" icon="align-left" aria-label="Align left"></Button>
+                <Button icon="align-center" aria-label="Align center"></Button>
+                <Button icon="align-right" aria-label="Align right"></Button>
+                <Button className="k-group-end" icon="align-justify" aria-label="Justify"></Button>
             </ButtonGroup>,
-            <ToolbarSeparator className="k-toolbar-button-separator" />,
-            <Button icon="more-vertical" className="k-toolbar-overflow-button" fillMode="flat"></Button>
+            <ToolbarSeparator key="separator" className="k-toolbar-button-separator" />,
```
inikolova, line 34: Do we need key attribute here? and for the Button below?

packages/html/src/editor/editor-content.tsx
```diff
@@ -26,7 +35,12 @@ export const EditorContent = (
                 stateClassNames(EDITORCONTENT_CLASSNAME, {
                     focus,
                 }),
-            )}>
+            )}
+            contentEditable={readonly ? 'false' : 'true'}
+            role="textbox"
```
inikolova, line 41: role here should be textbox, right?

tests/chat/chat-bubble-states.html
```diff
@@ -110,7 +110,7 @@
                                         </svg>
                                     </span>
                                 </button>
-                                <button class="k-button k-button-sm k-rounded-full k-disabled k-icon-button" aria-label="Folder" disabled="">
+                                <button class="k-button k-button-sm k-rounded-full k-disabled k-icon-button" aria-label="Send prompt" aria-live="polite" disabled="">
```
inikolova, line 113: Why we need aria-live="polite" attribute here?

packages/html/src/autocomplete/autocomplete.spec.tsx
```diff
@@ -140,15 +175,20 @@ export const Autocomplete: KendoComponent<KendoAutocompleteProps & KendoAutocomp
                 }
             </Input>
             {opened && popup &&
-                <Popup className="k-list-container k-autocomplete-popup">
+                <Popup
+                    className="k-list-container k-autocomplete-popup"
+                    containerClassName="k-autocomplete-popup-container"
```
inikolova: As discussed offline, this containerClassName attribute should be removed (everywhere), as it is not needed

tests/_output/bootstrap/multiselect/multiselect-flat.png
dtopalov: it is caused by the logic that applies both k-selected class and aria-selected attribute on a "selected" Chip, I will revise the logic to allow setting the class and attribute separately where necessary

packages/html/src/editor/templates/editor-normal.tsx
dtopalov: Yes, the [key](https://github.com/telerik/kendo-themes/pull/5796) props were added because these elements are in a JSX array (the [toolbarItems](https://github.com/telerik/kendo-themes/pull/5796) prop receives [...]). Without [key](https://github.com/telerik/kendo-themes/pull/5796), React emits "each child in a list should have a unique key" warnings during build/render. They're required by React for array children and don't affect the rendered HTML output — they're purely a React reconciliation concern.

packages/html/src/editor/templates/editor-responsive-toolbar.tsx
dtopalov: same as the other comment regarding key

tests/chat/chat-bubble-states.html
dtopalov: Using aria-live on a <button> element that changes its status (e.g., text or label) based on a condition is acceptable and often recommended to ensure screen reader users are immediately informed of the change without needing to refocus on the button.

Use aria-live="polite": For most status changes (e.g., "Follow" changing to "Following", or "Add to Cart" to "Added"), use aria-live="polite". The screen reader will announce the change only when the user is idle, preventing interruption of their current task.

DanielTakev (approved):

tests/_output/bootstrap/imageeditor/imageeditor.png
inikolova: As discussed offline, this change is not expected, so we should revert it.

tests/_output/classic/grid/grid-selection-aggregates.png
inikolova, line 1: Could we have an outcome, like the selection in the Aggregated example?

packages/html/src/listgroup/listgroup-item.spec.tsx
```diff
@@ -11,7 +11,7 @@ const options = {};
 export type KendoListGroupItemProps = {
     text?: string;
     children?: React.JSX.Element;
-    href?: string | boolean;
```
inikolova: Why we've removed the string option?

dtopalov: it is used only to determine whether a hardcoded anchor element with the passed text prop is to be rendered:
...
 { href && <a className="k-link">{text}</a>}

tests/_output/classic/grid/grid-selection-aggregates.png
dtopalov: yes, in cell selection mode when each cell from a given row is selected individually and then some other cells are selected like in the pattern from the screenshot

tests/_output/classic/multiselecttree/multiselecttree-outline.png
zhpenkov, line 1: Chip colors here have changed

tests/_output/material/orgchart/orgchart-node.png
zhpenkov, line 1: first node seems broken

tests/_output/material/scheduler/scheduler-day-view.png
zhpenkov, line 1: Extra select here is visible

tests/_output/material/taskboard/taskboard-pane.png
zhpenkov, line 1: Card body has another element here

tests/_output/classic/multiselecttree/multiselecttree-outline.png
dtopalov: this was unintentional, we need to adjust usage so that the Chip can have aria-selected attribute, but NO k-selected class

tests/_output/material/taskboard/taskboard-pane.png
dtopalov: this is a mistake, will fix

tests/_output/material/scheduler/scheduler-day-view.png
dtopalov: the select was added to more templates than it should have, removing it

tests/_output/bootstrap/grid/grid-sticky-columns.png
inikolova, line 1: This problem was addressed in the following commit - please take a look:
https://github.com/telerik/kendo-themes/pull/5811/changes/29e528d6ccbfe1e6d5bd6c277d5243bebc71c76a

tests/_output/bootstrap/map/map.png
inikolova, line 1: In that change intended?

PR #5800, opened 2026-03-02
chore(infra): tree-shake html dist

zhpenkov:


TeyaVes:
Looks great! I agree that this approach is way better than the one in: https://github.com/telerik/kendo-themes/pull/5770

Just one suggestion from Copilot, which we might consider: 

_Add "sideEffects": false to package.json.
The plugin only helps esbuild at build time. Downstream consumers (webpack, Rollup, Vite) that import from the published @progress/kendo-themes-html package also benefit from tree-shaking, but only if package.json declares "sideEffects": false. Since all barrel files are pure re-exports, this is safe and standard: `"sideEffects": false,`
This gives downstream bundlers the same hint the plugin gives esbuild._

Juveniel:
@zhpenkov move forward with this PR or close it

PR #5817, opened 2026-03-11
RFC: Oklch Intent-Based Colors

zhpenkov:
## How Colors Work Today

The current approach is straightforward. We have palettes — fixed scales of 15 shades per named color like Coral, Sky Blue, Green. We pick specific indices from those palettes to produce semantic tokens, which are consumed across our entire component library.

From palette Coral
Screenshot 2026-03-11 at 14 59 32
We pick our semantic tokens
```
--kendo-color-primary:             Coral[9]   → #ff6358
--kendo-color-primary-hover:       Coral[10]  → #d45349
```

Every palette is a static color. So is every semantic token. Every semantic token is a hex value, and changing one does not affect any other.

### 1. Overriding requires rewriting everything

If a user — or an LLM generating a theme — wants to rebrand, they must define all tokens individually. There are 10 tokens per color group across 8 groups. That is 80+ manual overrides to change a brand color.

### 2. No guaranteed contrast

There is no structural relationship between `--kendo-color-primary` and `--kendo-color-on-primary`. They are two independent hex values that happen to look good together in our handpicked default. Change one and contrast may break silently — nothing in the system prevents it.

### 3. Palettes are a dead end

After SCSS compilation, palettes disappear entirely. They serve only as a lookup table during the build to produce the default appearance. They are also **LOCKED** to named color spectra — "Coral", "Sky Blue", "Green" — which has nothing to do with a user's actual brand color.

## Three Problems, One Solution?

This proposal addresses two distinct problems with a single mechanism.

### Problem 1 — Expandable Colors

For user and LLM consumption, we want to accept **one** seed color and automatically expand it into the full set of derived tokens — hover, active, subtle, emphasis, contrast text, and so on. This is conceptually what `k-generate-color-variations` does in SCSS at build time, but we need it to happen at runtime in pure CSS so that any override propagates instantly.

### Problem 2 — Built-in Contrast

We want semantic token pairs to carry structural contrast guarantees.

```css
--kendo-color-primary:       /* the fill color */
--kendo-color-primary-hover: /* must be visually distinct from primary */
--kendo-color-on-primary:    /* must be readable on top of primary */
```

These relationships should be enforced by the derivation formulas, not by manual curation that breaks the moment someone changes a value.

### Problem 3 - Palettes Get in the Way

The current theme creation workflow looks like this:

1. Designers create palettes (indices 1 through 15) — for example, the Coral palette
2. We pick semantic colors from those palettes at specific indices

```css
--kendo-color-primary:       palette[9]
--kendo-color-primary-hover: palette[10]
```

This is one valid configuration — the one we chose for our default theme — out of endless possibilities.

In practice, the color tokens we ship are rarely kept at their default values. It is almost certain that a user will adapt them to fit their own design system and brand.

Consider the Coral palette as it exists today:

Screenshot 2026-03-11 at 14 59 32

We take index 9 for primary and index 10 for primary-hover.

If we expose palettes to the user, and someone shifts the entire palette to darker shades but does not change the index mapping for `primary` and `primary-hover` (9, 10), they end up with a dark `primary` and an even darker `primary-hover`. The interaction feedback becomes invisible.

Should we let users modify palettes directly? That pushes the complexity onto them — they would need to understand how every index maps to every semantic role. The palettes overcomplicate the color system and make it error-prone for everyone involved.

## Proposition: Oklch Intent-Based colors

Lets start from the unknown and work our way up to the known.

Forget palettes. Imagine the color is not predetermined. It is **X**.

```css
--kendo-color-primary: X;
```

With OKLCH — a perceptually uniform color space — we can produce every derived token by transforming the **L** (lightness), **C** (chroma), and **H** (hue) channels of X directly.

For example, if primary-hover should be slightly lighter than primary:

```css
--kendo-color-primary-hover: oklch(from var(--kendo-color-primary) calc(l + 0.04) c h);
```

This reads as: take the lightness of whatever X is, add 0.04, keep chroma and hue. Primary-hover is now structurally guaranteed to be lighter than primary.

But there is an assumption baked in — that primary is a dark color that needs lightening. What if someone sets primary to a bright yellow?

### The Direction Problem (and solution)

We need hover to move **away from the nearest extreme**: lighter when the seed is dark, darker when the seed is light. CSS has no conditional logic, but we can simulate a binary switch using `clamp()`.

The expression:

```
(0.6 − l) × 99999
```

produces a very large **positive** number when `l < 0.6` (dark seed) and a very large **negative** number when `l > 0.6` (light seed). Wrapping this in `clamp(−δ, …, +δ)` collapses it to exactly **+δ** or **−δ**:

```css
--kendo-color-primary-hover: oklch(from var(--kendo-color-primary)
    calc(l + clamp(-0.04, (0.6 - l) * 99999, 0.04)) c h);
```

| Seed lightness | Expression resolves to | Effect |
|---|---|---|
| L = 0.40 (dark) | `l + 0.04` | Lightens |
| L = 0.80 (light) | `l − 0.04` | Darkens |
| L = 0.60 (threshold) | `l + 0` | No change — graceful boundary |

The `0.04` delta and `0.6` threshold are design decisions. They encode the **intent** of the interaction, not a specific color.

### Structural Contrast

The same mechanism solves contrast for text-on-background tokens.

`on-primary` is the text color that sits on a primary-colored surface. It must always be readable regardless of what X is:

```css
--kendo-color-on-primary: oklch(from var(--kendo-color-primary)
    clamp(0.10, (0.6 - l) * 99999, 0.98) 0.01 h);
```

| Seed | on-primary resolves to | Result |
|---|---|---|
| Dark primary | L = 0.98, near-zero chroma | White text |
| Light primary | L = 0.10, near-zero chroma | Dark text |

Contrast is no longer a coincidence of two manually picked hex values. It is a structural property of the formula.

## What This Changes

We created semantic tokens **without committing to a specific color**.

We did not start from Coral, Sky Blue or from any palette. We started from X and defined relationships.

This is where most design systems get it wrong. They deliver a beautiful default theme with carefully curated colors for every token. But in practice, those tokens are almost always overridden. The moment a user sets a different primary color, all the careful curation is gone — the hover relationship, the contrast ratio, the subtle tint — all of it breaks because it was never encoded, only coincidental.

**Intent-based colors preserve the design intent regardless of the input color.** The relationships are the product, not the specific values.

> [!NOTE]  
> Instead of X, we will, of course, set a default color for our theme, but the focus here is not on the color itseft, it is on the idea that any color would work as expected.

## A New Paradigm & Design-to-Development Workflow

The paradigm shift: designers can now define **intent between tokens**.

Instead of: _"primary-hover is `#d45349`"_

The designer can specify: _"primary-hover has a lightness delta of 0.04 from primary"_

This intent is encoded once. It can then be verified with two test colors — one dark, one light — and it holds for every color in between.

### Backwards-compatible: extracting intent from existing tokens

Even better — we can take our current hardcoded semantic tokens (hex values) and extract their intent retroactively.

Given the existing primary `#ff6358` and primary-hover `#d45349`, we can convert both to OKLCH, measure the lightness delta between them, and record: _"hover is 0.04 darker than primary"_. Assuming primary is X — and it might be light or dark — we determine the delta and its direction for each derived token.

This means the designer workflow does not need to change drastically. Designers can continue delivering semantic tokens as concrete colors for a specific theme. We then derive the intent-based relationships from those values after the fact — computing the deltas, the chroma ratios, the contrast thresholds — and encode them as formulas.

The difference is what happens next. Instead of hardcoding those hex values, we store the **relationships** between them. The specific colors become one example (the default). The relationships become the system.

## POC Implementation

The `:root` block appended to [`packages/default/scss/all.scss`](../packages/default/scss/all.scss) implements the full Intent-Based system using CSS relative color syntax (`oklch(from …)`).
The Intent between tokens is directly extracted from our current hardcoded semantic tokens and modified to support any (X) color and have a built-in contrast.

Screenshot 2026-03-11 at 15 41 13

### Seeds

8 seed colors that the user controls. Change any one and its 9 derived tokens recalculate automatically:

```css
--kendo-color-primary:   oklch(0.6985 0.1923 27.19);
--kendo-color-secondary: oklch(0.50 0 0);
--kendo-color-tertiary:  oklch(0.67 0.15 230);
--kendo-color-info:      oklch(0.48 0.22 264);
--kendo-color-success:   oklch(0.63 0.20 142);
--kendo-color-warning:   oklch(0.84 0.17 90);
--kendo-color-error:     oklch(0.59 0.24 29);
--kendo-color-base:      oklch(0.97 0 0);
```

### Token derivation template

Every color group (primary, secondary, tertiary, info, success, warning, error) uses the same formula set. Base uses a variant tuned for neutral chrome. Here is primary as a representative example:

```css
/* Interaction states — intent-based direction */
--kendo-color-primary-hover:         oklch(from var(--kendo-color-primary) calc(l + clamp(-0.04, (0.6 - l) * 99999, 0.04)) c h);
--kendo-color-primary-active:        oklch(from var(--kendo-color-primary) calc(l + clamp(-0.08, (0.6 - l) * 99999, 0.08)) c h);

/* Subtle backgrounds — fixed high lightness, reduced chroma */
--kendo-color-primary-subtle:        oklch(from var(--kendo-color-primary) 0.96 calc(c * 0.12) h);
--kendo-color-primary-subtle-hover:  oklch(from var(--kendo-color-primary) 0.93 calc(c * 0.18) h);
--kendo-color-primary-subtle-active: oklch(from var(--kendo-color-primary) 0.89 calc(c * 0.25) h);

/* Emphasis — compressed tint, always lighter than seed */
--kendo-color-primary-emphasis:      oklch(from var(--kendo-color-primary) calc(l * 0.65 + 0.33) calc(c * 0.55) h);

/* Contrast text tokens */
--kendo-color-primary-on-subtle:     oklch(from var(--kendo-color-primary) calc(l * 0.35) c h);
--kendo-color-on-primary:            oklch(from var(--kendo-color-primary) clamp(0.10, (0.6 - l) * 99999, 0.98) 0.01 h);
--kendo-color-primary-on-surface:    oklch(from var(--kendo-color-primary) clamp(0.20, l, 0.60) c h);
```

This is an exploration for: https://github.com/telerik/kendo-themes-private/issues/471


TeyaVes:
closed in favor of:  https://github.com/telerik/kendo-themes/pull/5853

Issue #5818, opened 2026-03-09
RTL mode not applied for some dropdown components


mbechev:
### Description

The RTL mode is not applied for popup items for DropDownList, MultiSelect, ComboBox, and AutoComplete (all not tree-like dropdowns).

<img width="438" height="309" alt="Image" src="https://github.com/user-attachments/assets/9222b6f4-c381-4748-b827-dfbb840d40b2" />

The issue start reprodicing since v23.0.0. The RTL is applied correctly with the previous v22.0.1.

### Steps To Reproduce

https://stackblitz.com/edit/angular-o524ymdh

### Actual Behavior

The RTL is no longer applied to the dropdown list options. The items are still shown from left to right.

### Expected Behavior

The dropdown items should be right-to-left aligned as in v22.0.1:
https://stackblitz.com/edit/angular-qgfgwoq5-s8m4egm1

PR #5819, opened 2026-03-11
List alignment

inikolova:
Closes https://github.com/telerik/kendo-themes/issues/5818


epetrow (approved):
I am surprised we do not have a visual test for that case, maybe we should add one

PR #5834, opened 2026-03-20
feat(html): add a11y docs generator

dtopalov:
## A11y Documentation Generator

Generates WAI-ARIA accessibility documentation JSON files from `ariaSpec` objects defined on html package components.

Replaces `@progress/wct-a11y-spec` as the source of truth — the `ariaSpec` on each component's `.spec.tsx` is now the single canonical definition.

### Output Format

Produces a JSON array per component in the format consumed by the [Kendo UI Design System docs](https://www.telerik.com/design-system/docs/). Each element is an object with a single key:

```json
[
    { "h2": "Accessibility Support" },
    { "p": "Out of the box, the Telerik and Kendo UI Breadcrumb ..." },
    { "h3": "WAI-ARIA" },
    { "table": { "headers": ["Selector", "Attribute", "Usage"], "rows": [...] } },
    { "link": { "title": "ToolBar accessibility specification", "source": "{{toolbar_a11y_link}}" } },
    { "h3": "Testing" },
    { "h3": "Screen Readers" },
    { "table": { "headers": ["Environment", "Tool"], "rows": [...] } }
]
```

### Architecture

| File | Purpose |
|---|---|
| `types.ts` | TypeScript interfaces — `AriaSpec`, `AriaRule`, `ComponentMeta`, `A11yJsonElement`, `A11yDocPage`, `A11yGeneratorOptions` |
| `component-registry.ts` | Discovers components with `ariaSpec` from html exports; uses `exportName` as display name (PascalCase) |
| `json-renderer.ts` | Builds the JSON array: intro, WAI-ARIA table(s), cross-links, testing, screen readers |
| `page-generator.ts` | Public API — `generateA11yDoc()` / `generateA11yDocs()` |
| `index.ts` | Barrel export |
| `scripts/generate-a11y-docs.js` | Generator script — outputs to `packages/html/docs/` |

#### From this repo

```bash
# Build first, then generate
npm run docs:a11y

# Or step by step:
npm run build -w packages/html
node packages/html/scripts/generate-a11y-docs.js
```

#### From a consuming repo (e.g. kendo-design-system)

```typescript
import * as fs from 'fs';
import * as path from 'path';
import * as htmlExports from '@progress/kendo-themes-html';
import { buildRegistry, generateA11yDocs } from '@progress/kendo-themes-html';

// Build the component registry — reads ariaSpec at runtime, no file parsing needed
const registry = buildRegistry(
    htmlExports as unknown as Record<string, unknown>,
);

console.log(`Found ${registry.size} components with ariaSpec`);

// Generate JSON files
const outDir = path.join('dist', 'design-system');

const pages = generateA11yDocs(registry, {
    outputPath: (c) => path.join(outDir, `${c.displayName}.json`),
});

fs.mkdirSync(outDir, { recursive: true });

for (const page of pages) {
    if (page.outputPath) {
        fs.writeFileSync(page.outputPath, JSON.stringify(page.json, null, 4), 'utf-8');
    }
}

console.log(`Generated ${pages.length} a11y JSON files in ${outDir}/`);
```

### Cross-References

Complex components (Grid, ComboBox, etc.) reference other components via `seeAlso` on the `ariaSpec`. The generator handles two patterns:

#### Standalone cross-link (e.g. Pager)

When a `seeAlso` target is a standalone component that doesn't have inline rules in the parent, the generator produces a cross-link section:

```json
{ "h4": "Pager" },
{ "p": "For the Grid Pager WAI-ARIA spec, please review the Pager component." },
{ "link": { "title": "Pager accessibility specification", "source": "{{pager_a11y_link}}" } }
```

#### Section with cross-link + inline rules (e.g. Grid Toolbar)

When a `seeAlso` target matches a section that also has its own inline rules, both the cross-link and the rules table are included:

```json
{ "h4": "Grid Toolbar" },
{ "p": "Grid Toolbar follows the specification of the ToolBar component." },
{ "link": { "title": "ToolBar accessibility specification", "source": "{{toolbar_a11y_link}}" } },
{ "table": { "headers": ["Selector", "Attribute", "Usage"], "rows": [...] } }
```

#### Internal sub-component (e.g. Grid Filter Menu)

When a section has no matching `seeAlso` reference, only the inline rules table is included:

```json
{ "h4": "Grid Filter Menu" },
{ "table": { "headers": ["Selector", "Attribute", "Usage"], "rows": [...] } }
```

### Options

| Option | Type | Description |
|---|---|---|
| `componentNameMap` | `Record<string, string>` | Display name overrides (e.g. `{ 'dropdownlist': 'DropDownList' }`) |
| `includeComponents` | `string[]` | Whitelist of component ids |
| `excludeComponents` | `string[]` | Blacklist of component ids |
| `outputPath` | `(component) => string` | Resolve output file path per component |

### Key Design Decisions

- **No static parsing** — `description` and `seeAlso` are runtime properties on `AriaSpec`, not extracted from JSDoc. No regex, no file reading.
- **No hardcoded name maps** — display names come from `exportName` (already PascalCase).
- **Algorithmic section grouping** — sections derived by stripping root CSS class prefix from selectors. No pattern map to maintain.
- **96 components** discovered and generated.


packages/html/src/a11y-docs/component-registry.ts
```diff
@@ -0,0 +1,256 @@
...
+        .split(/[-_]/)
+        .map((w) => w.charAt(0).toUpperCase() + w.slice(1).toLowerCase())
+        .join('');
+}
+
+/**
+ * Built-in display name map for components whose PascalCase derivation
+ * does not match the canonical product name.
+ */
+const DEFAULT_DISPLAY_NAMES: Record<string, string> = {
```
zhpenkov: Can we just kebab to Pascal case?

Not a fan of hardcoded component records.

If we add a component, we need to add it here as well.

packages/html/src/a11y-docs/component-registry.ts
```diff
@@ -0,0 +1,256 @@
...
+    for (const re of [regex, regex2]) {
+        let match: RegExpExecArray | null;
+        while ((match = re.exec(jsdoc)) !== null) {
+            const targetName = match[1].toLowerCase();
+            const label = (match[2] || '').replace(/^for\s+/i, '').trim();
+            if (!refs.some((r) => r.targetId === targetName)) {
+                refs.push({ targetId: targetName, label });
+            }
+        }
+    }
```
zhpenkov: We have this data from each component's ariaSpec property. Do we need to statically parse it?

packages/html/src/a11y-docs/json-renderer.ts
```diff
@@ -0,0 +1,351 @@
...
+}
+
+/**
+ * Format a JSDoc list item: wrap parenthesized role/attribute values
+ * in backticks and append a semicolon.
+ *
+ * `'Toolbar (role=toolbar)'` → `` 'Toolbar (`role=toolbar`);' ``
+ */
+function formatListItem(item: string): string {
+    return item.replace(/\(([^)]*=[^)]+)\)/g, '(`$1`)') + ';';
```
zhpenkov: Can we change it in the JSDoc ?

Let's try minimizing regex parsing, as it's error-prone.

packages/html/src/a11y-docs/json-renderer.ts
```diff
@@ -0,0 +1,351 @@
...
+        [/\.k-master-row|\.k-grouping-row|\.k-group-cell|\.k-drag-cell/, `${componentName} Content`],
+
+        [/\.k-toolbar/, componentName],
+        [/\.k-pager/, componentName],
+        [/\.k-input-inner/, `${componentName} Input`],
+        [/\.k-input-button/, `${componentName} Button`],
+        [/-popup\s/, `${componentName} Popup`],
+        [/\.k-dialog/, componentName],
+        [/\.k-window/, componentName],
+    ];
```
zhpenkov: Let's prevent maintenance burden

packages/html/src/a11y-docs/examples/consumer-usage.ts
zhpenkov, line 1: We can put that in the PR description

packages/html/src/a11y-docs/README.md
zhpenkov, line 1: this as well

zhpenkov:
General notes, it is looking great, but as we are trying to match the original one 1:1, many unnecessary parsings, records of strings, and mappings are implemented.

Let's minimize those and simplify the docs generator, so it consistently produces decent output and does not really on heavy modifications.

**Key priorities:**
Escape maintenance burden at all costs
Make it straightforward
Follow the same structure, but make alterations to the content if it conflicts with the simplicity of the generator.

packages/html/tsconfig.json
zhpenkov, line 1: `packages/html` might be a better place for the docs folder, as other packages follow this rule as well

Also, I am thinking that we can name the folder just `docs`

We can move the scripts to `html/scripts` and the output from it to `html/docs`

zhpenkov:
Also, will there be a script for generating those JSONs in the global package.json ?

dtopalov:
## PR Review Changes Summary

### Addressing zhpenkov's feedback

1. **Removed `DEFAULT_DISPLAY_NAMES` hardcoded record** — Display names now come from the component's `exportName` directly (already PascalCase). No maintenance needed when adding new components.

2. **Removed all JSDoc/regex parsing** — `extractAriaSpecJSDoc`, `parseSeeReferences`, and the `sourceReader` parameter are gone. Cross-references (`seeAlso`) and `description` are now runtime properties on `AriaSpec` — no static parsing needed.

3. **Removed hardcoded `inferSectionTitle` pattern map** — Section grouping is now algorithmic: strips the root CSS class prefix from each rule's selector to derive section names. Zero maintenance when adding components.

4. **Deleted README.md and `examples/`** — Moved usage info to this PR description instead.

5. **Added generator script** — generate-a11y-docs.js, outputs JSON files to `packages/html/docs/`.

6. **Added npm scripts** — `docs:a11y` in both package.json and root package.json.

### Addressing Copilot's feedback

7. **Fixed `@see` target id mismatch** — `seeAlso` uses correct kebab-case ids directly (e.g., `'action-sheet'` not `'actionsheet'`), so `{{action-sheet_a11y_link}}` resolves correctly.

8. **Fixed `outputPath` receiving wrong `displayName`** — Now passes `{ ...component, displayName }` with the resolved name.

### Spec file updates (17 files)

- Added `seeAlso: string[]` to all components that previously had `@see` JSDoc references:
  Grid, Autocomplete, ComboBox, DropdownList, DropdownTree, DropdownGrid, MultiSelect, Editor, ImageEditor, MediaPlayer, Gantt, PDFViewer, SegmentedControl, ChipList, CardList, SmartBox, SpeechToTextButton.
- Added `description` to Grid's `ariaSpec` (the composite 4-part structure text that was previously only in JSDoc).
- Made `AriaSpec.selector` optional (some components like Autocomplete don't specify one).
- Reverted the tsconfig.json exclude entry for the now-deleted `examples/` dir.

### Verified

- TypeScript typecheck: ✅ clean
- ESLint: ✅ clean
- End-to-end generation: ✅ 96 components, cross-references resolve correctly

zhpenkov:
Here are my thoughts on the current state.
Feel free to implement what seems logical.

1. **Drop `componentNameMap`** — it's never used. Display names already come from export names (PascalCase). Remove the option from `A11yGeneratorOptions` and simplify `resolveDisplayName` to registry lookup + kebab fallback. This un-threads a parameter from 4 functions.

2. **Make `description` a pre-built `A11yJsonElement[]` instead of a parsed string** — only Grid uses it today. `parseDescription()` is a mini markdown parser with bullet detection and a regex for backtick-wrapping `key=value` (already had a double-wrapping bug). Let the spec author provide the exact output shape directly. Zero runtime parsing, zero regex.

3. **Add an optional `section?: string` to `AriaRule` instead of inferring sections from CSS selectors** — `inferSectionKey()` strips class prefixes, removes `:not()` blocks, and guesses section names from selector structure. This will break with unexpected selectors. Explicit `section` fields make grouping trivial and `inferSectionKey`/`kebabToTitle`/most of `groupRulesIntoSections` can go away.

4. **Make `seeAlso` matching exact instead of fuzzy** — `findMatchingSeeAlso()` does substring matching between section titles and display names. With explicit `section` fields, matching becomes identity comparison (`section === seeAlso entry`). No substring guessing.

5. **Fold page-generator.ts into the renderer or flatten it** — `generateA11yDoc` is a thin wrapper that resolves `displayName` then calls `buildA11yJson` which resolves it again. 53 lines of indirection for a filter+map loop.

6. **Remove dead type variants** — `{ h5: string }` in `A11yJsonElement` (never produced), `implicitRole` on `AriaSpec` (declared, never read), `A11yListItem` union (only `string` branch used).

**Structural concerns**

7. **Move `a11y-docs/` out of `src/`** — `src/` contains only component specs. The generator is not a component — it's a build-time tool. It lives better under scripts (next to generate-a11y-docs.js, `build.js`, etc). This also means it doesn't need to be exported from the package's index.ts or bundled into dist.

8. **Consolidate files** — currently 5 source files (types.ts, component-registry.ts, json-renderer.ts, page-generator.ts, index.ts) for what is fundamentally: scan exports → build tables → write JSON. With the simplifications above (no `parseDescription`, no `componentNameMap`, explicit sections, merged page-generator), this could be 2 files: one for types, one for the generator logic. Or even a single file since the types are small.

9. **Distribution model — commit the JSONs or ship in dist?** — Right now, generate-a11y-docs.js writes to docs, which can be committed to the repo. Meanwhile, the generator code is also exported from the npm package (via index.ts + `files` in package.json includes `dist/`), so consumers can also generate at build time. We need to pick one:
   - **Option A: Commit JSONs** — simpler for consumers (just read files), but every ariaSpec change requires a regeneration step and a diff of 90+ JSON files in every PR.
   - **Option B: Generate at build time only** — add to `prepublishOnly` or the consuming repo's build. No committed artifacts, no stale-JSON risk. The generator stays in scripts (not exported from the package).
   - **Option C: Ship JSONs in dist/** — generate during `npm run build` and include `docs/*.json` in the package's `files` array. Consumers `import`/`require` the JSONs from the published package. No committed JSONs, no generator API to maintain.

   Leaning toward C or B. Committing 90+ generated JSON files creates noise in PRs and a staleness risk — if someone changes an ariaSpec but forgets `npm run docs:a11y`, the JSONs drift silently.
   @telerik/kendo-front-end wdyt?

epetrow:
> Here are my thoughts on the current state. Feel free to implement what seems logical.
> 
> 1. **Drop `componentNameMap`** — it's never used. Display names already come from export names (PascalCase). Remove the option from `A11yGeneratorOptions` and simplify `resolveDisplayName` to registry lookup + kebab fallback. This un-threads a parameter from 4 functions.
> 2. **Make `description` a pre-built `A11yJsonElement[]` instead of a parsed string** — only Grid uses it today. `parseDescription()` is a mini markdown parser with bullet detection and a regex for backtick-wrapping `key=value` (already had a double-wrapping bug). Let the spec author provide the exact output shape directly. Zero runtime parsing, zero regex.
> 3. **Add an optional `section?: string` to `AriaRule` instead of inferring sections from CSS selectors** — `inferSectionKey()` strips class prefixes, removes `:not()` blocks, and guesses section names from selector structure. This will break with unexpected selectors. Explicit `section` fields make grouping trivial and `inferSectionKey`/`kebabToTitle`/most of `groupRulesIntoSections` can go away.
> 4. **Make `seeAlso` matching exact instead of fuzzy** — `findMatchingSeeAlso()` does substring matching between section titles and display names. With explicit `section` fields, matching becomes identity comparison (`section === seeAlso entry`). No substring guessing.
> 5. **Fold page-generator.ts into the renderer or flatten it** — `generateA11yDoc` is a thin wrapper that resolves `displayName` then calls `buildA11yJson` which resolves it again. 53 lines of indirection for a filter+map loop.
> 6. **Remove dead type variants** — `{ h5: string }` in `A11yJsonElement` (never produced), `implicitRole` on `AriaSpec` (declared, never read), `A11yListItem` union (only `string` branch used).
> 
> **Structural concerns**
> 
> 7. **Move `a11y-docs/` out of `src/`** — `src/` contains only component specs. The generator is not a component — it's a build-time tool. It lives better under scripts (next to generate-a11y-docs.js, `build.js`, etc). This also means it doesn't need to be exported from the package's index.ts or bundled into dist.
> 8. **Consolidate files** — currently 5 source files (types.ts, component-registry.ts, json-renderer.ts, page-generator.ts, index.ts) for what is fundamentally: scan exports → build tables → write JSON. With the simplifications above (no `parseDescription`, no `componentNameMap`, explicit sections, merged page-generator), this could be 2 files: one for types, one for the generator logic. Or even a single file since the types are small.
> 9. **Distribution model — commit the JSONs or ship in dist?** — Right now, generate-a11y-docs.js writes to docs, which can be committed to the repo. Meanwhile, the generator code is also exported from the npm package (via index.ts + `files` in package.json includes `dist/`), so consumers can also generate at build time. We need to pick one:
>    
>    * **Option A: Commit JSONs** — simpler for consumers (just read files), but every ariaSpec change requires a regeneration step and a diff of 90+ JSON files in every PR.
>    * **Option B: Generate at build time only** — add to `prepublishOnly` or the consuming repo's build. No committed artifacts, no stale-JSON risk. The generator stays in scripts (not exported from the package).
>    * **Option C: Ship JSONs in dist/** — generate during `npm run build` and include `docs/*.json` in the package's `files` array. Consumers `import`/`require` the JSONs from the published package. No committed JSONs, no generator API to maintain.
>    
>    Leaning toward C or B. Committing 90+ generated JSON files creates noise in PRs and a staleness risk — if someone changes an ariaSpec but forgets `npm run docs:a11y`, the JSONs drift silently.
>    @telerik/kendo-front-end wdyt?

I will only comment on the structural concerns here:

1. **Move `a11y-docs/` out of `src/** - absolutely agree on this. `src/**` is intended for components spec and templates and in my opinion should not be polluted
2. **Consolidate files** - I am onboard for the idea of simplification, no need for overengineering
3. **Distribution model — commit the JSONs or ship in dist?** - I also lean towards options C or B. Definitely would want to avoid unnecessary PR noise

epetrow (commented):
Below I'll list my thoughts on the current implementation, feel free to address only what seems relevant.

1. **`seeAlso` entries serve two different purposes that are conflated via implicit string matching**:

- Matched when a `seeAlso` entry (e.g. `toolbar`) matches a `section` key on one of the component's own rules, the generator emits a cross-link inline with that section.
- Unmatched when no section key matches (e.g. `pager` in Grid), it emits a standalone cross-link block at the bottom.

The coupling between `seeAlso` and `section` is manual and thus the two could get out of sync. Maybe we could have some kind of validation at generation time or we could simply decouple the two mechanisms so cross-links do not depend on section key strings matching.

2. **We do not check if `seeAlso` targets exist in the registry**  - what happens if someone types `toolBar` (wrong casing) or references a component with no `ariaSpec.rules` (is this even a valid concern?)

3. **`description` embeds ouput-format JSON in component source** - Only the Grid uses `description` so far, but the design embeds output-format specific JSON (`{p: ...}, {ul: ...}`) directly in the source files. This couples the component specs to the docs and if the docs format changes, every spec with a description needs updating. It also makes the specs harder to read since the `desription` array is not a testable ARIA contract. Lastly if more components need descriptions, spec files become documentation containers.

4. **The screen readers tables is hardcoded** - we return the same static browser/screen-reader combinations for all components. There is no way to override per-component.

zhpenkov:
closing in favor of: https://github.com/telerik/kendo-themes/pull/6061

PR #5815, opened 2026-03-10
chore(vpat): add VPAT report generation POC

dtopalov:
# PR Summary: VPAT Generation & Accessibility Infrastructure

## VPAT Report Generator

Added a VPAT® 2.5Rev WCAG Edition report generator to the `@progress/kendo-themes-html` package.

**New files in vpat:**
- vpat-generator.ts — generates Markdown reports from test data and configuration
- `default-config.ts` — baseline VPAT config with known exceptions, general compliance statements, and WCAG criteria mappings
- `wcag-criteria.ts` — WCAG 2.0/2.1/2.2 criteria reference (Levels A, AA, AAA)
- `types.ts` — TypeScript interfaces for config, overrides, reports, and test analysis
- index.ts — public exports (`generateVpatReport`, types)

**CLI wrapper (vpat):**
- generate-vpat.mjs — CLI for generating Markdown reports with `--overrides`, `--a11y-report`, `--contrast-report`, `--out` arguments
- README.md — full usage documentation (programmatic API, CLI, PDF generation, overrides format)

**Unit tests:**
- vpat-generator.test.ts — 39 tests covering report generation, config merging, override behavior, WCAG criteria mapping, and edge cases

## A11y Test Infrastructure

**Contrast test script (test-contrast.mjs):**
- New script for running contrast ratio tests with JSON report output

**A11y test updates (test-a11y-unified.mjs):**
- Updated to support JSON report output alongside existing formats

**CI workflow updates:**
- `_test-a11y-specs.yml` — runs headless via `HEADLESS=true`
- `_test-contrast.yml` — new reusable contrast test workflow

## Component Accessibility Fixes

- **Signature** — added full accessibility support: `aria-label` on action buttons, `role="img"` + `aria-label` + `tabIndex` on canvas, full `ariaSpec` with 4 rules
- **SplitButton** — fixed `aria-disabled` being overwritten by child Button's explicit attribute (now uses Button's `ariaDisabled` prop)
- **MenuButton** — same `aria-disabled` fix as SplitButton
- **SmartBox** — fixed `aria-pressed` being overwritten (now uses Button's `togglable` and `selected` props)
- **PanelBar** — fixed `aria-hidden` being overwritten on group element (conditionally applies only when explicitly provided)
- **TileLayout** — removed deprecated WAI-ARIA 1.2+ attributes (`aria-grabbed`, `aria-dropeffect`) from ariaSpec, rendering, and test HTML files

## Cleanup

[...]

PR #5839, opened 2026-03-23
Agent Skills

zhpenkov:
Designed top-down — what does the agent need to create or update an entire theme or component from scratch?

Our workflows almost always combine the same steps: variables, styles, specs, templates, tests. Each step is now a focused, reusable skill, and the higher-level skills delegate to them:

```
manage-theme
├── manage-scss-modules          (spacing, colors, elevation, typography, border-radii)
├── manage-components
│   ├── manage-scss-variables    ($kendo-* null in core, concrete in themes)
│   ├── manage-scss-styles       (layout/theme mixins, _index.scss, registration)
│   ├── manage-html-specs        (.spec.tsx — types, options, states, render)
│   ├── manage-html-templates    (preset configurations)
│   └── manage-html-tests        (visual regression screenshots)
├── manage-html-a11y             (ARIA attributes & a11y tests)
└── audit-design-tokens          (scan for hardcoded values)
```

Screenshot 2026-03-26 at 15 51 39

Tested component creation — works well with an image + context. In theory, for component updates, we can feed it Figma changes, a screenshot, or text, and it should handle the rest.

This also applies to new theme updates since `manage-theme` orchestrates components, modules, and everything else.

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,9 +1,12 @@
 {
     "extends": "stylelint-config-standard-scss",
     "plugins": [
-        "stylelint-scss"
+        "stylelint-scss",
+        "@stylistic/stylelint-plugin"
     ],
     "rules": {
+        "@stylistic/declaration-bang-space-before": "always",
+
         "color-function-alias-notation": null,
         "alpha-value-notation": "number",
         "selector-not-notation": null,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [13.1.2-dev.0](https://github.com/telerik/kendo-themes/compare/v13.1.1...v13.1.2-dev.0) (2026-03-30)
+
+**Note:** Version bump only for package kendo-themes.repo
+
+
+
+
+
 ## [13.1.1](https://github.com/telerik/kendo-themes/compare/v13.1.1-dev.2...v13.1.1) (2026-03-23)
 
 **Note:** Version bump only for package kendo-themes.repo
--- a/integrations/webpack/package-lock.json
+++ b/integrations/webpack/package-lock.json
@@ -1626,15 +1626,6 @@
       "integrity": "sha512-1NNCs6uurfkVbeXG4S8JFT9t19m45ICnif8zWLd5oPSZ50QnwMfK+H3jv408d4jw/7Bttv5axS5IiHoLaVNHeQ==",
       "license": "MIT"
     },
-    "node_modules/randombytes": {
-      "version": "2.1.0",
-      "resolved": "https://registry.npmjs.org/randombytes/-/randombytes-2.1.0.tgz",
-      "integrity": "sha512-vYl3iOX+4CKUWuxGi9Ukhie6fsqXqS9FE2Zaic4tNFD2N2QQaXOMFbuKK4QmDHC0JO6B1Zp41J0LpT0oR68amQ==",
-      "license": "MIT",
-      "dependencies": {
-        "safe-buffer": "^5.1.0"
-      }
-    },
     "node_modules/readdirp": {
       "version": "4.1.2",
       "resolved": "https://registry.npmjs.org/readdirp/-/readdirp-4.1.2.tgz",
@@ -1714,26 +1705,6 @@
       "engines": {
         "node": ">=8"
       }
-    },
-    "node_modules/safe-buffer": {
-      "version": "5.2.1",
-      "resolved": "https://registry.npmjs.org/safe-buffer/-/safe-buffer-5.2.1.tgz",
-      "integrity": "sha512-rp3So07KcdmmKbGvgaNxQSJr7bGVSVk5S9Eq1F+ppbRo70+YeaDxkw5Dd8NPN+GD6bjnYm2VuPuCXmpuYvmCXQ==",
-      "funding": [
-        {
-          "type": "github",
-          "url": "https://github.com/sponsors/feross"
-        },
-        {
-          "type": "patreon",
-          "url": "https://www.patreon.com/feross"
-        },
-        {
-          "type": "consulting",
-          "url": "https://feross.org/support"
-        }
-      ],
-      "license": "MIT"
     },
     "node_modules/sass": {
       "version": "1.90.0",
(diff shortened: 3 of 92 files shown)


zhpenkov:
### Skills Experiment 1

Commit: [8e9eb21](https://github.com/telerik/kendo-themes/pull/5839/commits/8e9eb21ebfc89bdc7c63dca60ffe1c155ccce2ee)

Agent Prompt: `create a kbd component`

**Result:**
Scaffolded a simple `kbd` component using `scaffold-component` skill (~35 files across core + 5 themes + HTML + tests). Compiled and linted cleanly on first run.

**To Improve:**
Skill had no visual test step, templates had a wrong import path, and a type blocked `children`. All three were fixed in the skill files and retroactively in the component.

Added `session-review` skill — runs end-of-session to diff what a skill prescribes vs. what actually happened, then patches the skill.

Feedback loop for all skills: use → find gap → fix → next use is better.

zhpenkov:
### Skills Experiment 2

**Commit:** 

**Agent Prompt:** `Create command component`

**Result:** Scaffolded a `command` (command palette) component — 37 files total: 3 core SCSS files, 4 files × 5 themes, 6 HTML spec/template/test files, 1 unit test, plus 6 index registrations. SCSS compiled across all 7 projects on second attempt; TypeScript passed with zero errors.

**Issues Hit:**
- Material and Fluent themes don't define a global `$kendo-focus-shadow` variable — the initial scaffold copied from default/classic and failed compilation. Fixed by using explicit `color-mix()` values for those two themes.
- Bootstrap's `@use` registration failed silently due to a blank-line mismatch in the search pattern; caught on review and patched.

**To Improve:** The `scaffold-component` skill templates reference `$kendo-focus-shadow` as if it's universal — it should note that Material and Fluent need explicit focus shadow values. The skill could also auto-detect which global variables are available per theme before generating variable files.

.github/skills/_shared/file-templates.md
```diff
@@ -0,0 +1,422 @@
+# File Templates Reference
+
+Detailed boilerplate templates for each file in the component scaffold. The agent should read this file during Step 2-6 of the scaffold procedure.
+
+## Core Package Templates
+
+### `packages/core/scss/components/{component}/_variables.scss`
+
+```scss
+@use "../../color-system/_swatch.scss" as *;
```
epetrow: I do not think this reference is needed

.github/skills/audit-design-tokens/SKILL.md
```diff
@@ -0,0 +1,139 @@
...
+- **Border radius**: Replace with `k-border-radius(key)` or `$kendo-*-border-radius` variable
+
+Offer to apply the fixes.
+
+## Gotchas
+
+- **Core `_variables.scss` files use `null !default`** — they define the token skeleton, not values. Scan theme overrides (default, classic, material, fluent, bootstrap) for actual hardcoded values.
+- **Theme `_variables.scss` paths differ**: `packages/{theme}/scss/{component}/` (no `components/` subdirectory), unlike core which uses `packages/core/scss/components/{component}/`.
+- **Fluent theme wraps everything in `var()`** — so `var(--kendo-spacing-2, 0.5rem)` is correct even though it contains a rem literal.
+- **Material theme uses `$mat-*` internal variables** in some places — these are imported from Material Design and are acceptable.
```
epetrow: I don't think this is true

.github/skills/manage-components/references/update-checklist.md
```diff
@@ -0,0 +1,171 @@
...
+Detailed checklists the agent should follow during each step of the update procedure.
+
+## Variable Addition Checklist
+
+When adding a new `$kendo-{component}-*` variable:
+
+### Core (`packages/core/scss/components/{component}/_variables.scss`)
+
+- [ ] Add declaration with `null !default`
+- [ ] Add SassDoc `///` comment above the declaration
```
epetrow: We generally do not document variables in `packages/core` with the exception of `$kendo-{component}-theme-colors` and `$kendo-{component}-sizes`

.github/skills/manage-components/references/update-checklist.md
```diff
@@ -0,0 +1,171 @@
...
+
+When removing a variable:
+
+- [ ] Confirm with the user that a breaking change is acceptable, OR add a deprecation comment
+- [ ] Remove from core `_variables.scss`
+- [ ] Remove from all theme `_variables.scss` files
+- [ ] Remove from all `@forward` maps
+- [ ] Remove references from `_layout.scss` and `_theme.scss`
+- [ ] Remove from HTML spec if referenced in comments
+- [ ] Update any unit tests that reference the variable
```
epetrow, line 51: Maybe we could add regen docs in these checklists as well

.github/skills/manage-components/references/update-checklist.md
```diff
@@ -0,0 +1,171 @@
...
+- [ ] Add export to `tests/index.ts`
+
+## Option Addition Checklist
+
+When adding a new option or option value (e.g., new size `xlarge`):
+
+### Core
+
+- [ ] Add size-specific variables: `$kendo-{component}-xl-padding-x: null !default;`, etc.
+- [ ] Add to the size/option map in `_layout.scss` or `_theme.scss`
```
epetrow: `_theme.scss` can be removed we do not keep size-specific variables there

.github/skills/manage-html-a11y/SKILL.md
```diff
@@ -0,0 +1,226 @@
...
+
+```bash
+npm run typecheck --prefix packages/html
+```
+
+### Step 6: Update ARIA reference doc (if needed)
+
+If `aria/{component}_aria.md` exists and is outdated (selectors don't match, rules are missing), update it to match the implemented `ariaSpec.rules`. The spec TSX is the source of truth; the markdown is reference documentation.
+
+If no ARIA markdown exists and the component is complex enough to warrant one, create `aria/{component}_aria.md` following the format of existing docs (see `aria/button_aria.md` or `aria/checkbox_aria.md`).
```
epetrow: If [this](https://github.com/telerik/kendo-themes/pull/5815) removes the need of `_aria.md` files, maybe the step would not be needed

.github/skills/manage-scss-variables/SKILL.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+## Variable change checklist
+
+When **changing** an existing variable's value:
+- Update only the targeted themes — don't touch themes the user didn't request
+- Do NOT change the core `null` declaration
+- Update the SassDoc comment if the description no longer fits
+- Verify the `@forward` map still includes this variable
+- Check for derived/calc variables that depend on this one
```
epetrow, line 102: I wonder if we should also add docs regeneration to this checklist as well

PR #5844, opened 2026-03-25
Remove Module Token Fallbacks

zhpenkov:
Here are the advantages of removing all token fallbacks (typography, color, spacing, border-radius, elevation) from the Themes:

1. **Reduced CSS output size** — Thousands of redundant fallback values stripped from compiled CSS, meaningfully shrinking bundle size across all themes and swatches.

2. **Reduced Redundancy** — Token values are ALWAYS defined once in `:root` via the `--styles()` mixins. Embedding them again as `var()` fallbacks is redundant, as the fallback is never called.

3. **Predictable custom property overrides** — With fallbacks, overriding a global token (e.g., `--kendo-font-size: unset`) doesn't propagate as expected because the hardcoded fallback kicks in. Without fallbacks, overrides work as intended throughout the cascade.

   For example, a user can set all color tokens to `unset` to reset all colors of their theme. (This will now work)

6. **Cleaner, more readable SCSS** — Variable definitions like `var(--kendo-list-sm-font-size, var(--kendo-font-size, inherit))` collapse to `var(--kendo-list-sm-font-size, var(--kendo-font-size))`, reducing noise and nesting depth.


TeyaVes (approved):

packages/default/scss/color-preview/_variables.scss
```diff
@@ -24,9 +25,12 @@ $kendo-color-preview-hover-border: k-color(border) !default;
 /// The background color of the ColorPreview when no color is selected.
 /// @group color-preview
 $kendo-color-preview-no-color-bg: k-color(surface-alt) !default;
+
+// Do not use css variable for the $kendo-color-preview-no-color-text, as it is used in the SVG data URI for the no-color image, which does not support CSS variables.
+// Do not use css variable for the $kendo-color-preview-no-color-text, as it is used in the SVG data URI for the no-color image, which does not support CSS variables.
```
epetrow, line 29: Maybe remove the duplicate comment?

packages/fluent/scss/color-preview/_variables.scss
```diff
@@ -25,7 +26,7 @@ $kendo-color-preview-hover-border: var(--kendo-color-preview-hover-border, color
 $kendo-color-preview-no-color-bg: var(--kendo-color-preview-no-color-bg, k-color(app-surface)) !default;
 /// Text color of the color preview when no color is selected.
```
epetrow, line 29: Maybe we should also add the comment about not using css variables here since we do for the rest of the themes 🤔

epetrow (approved):

packages/fluent/scss/color-preview/_variables.scss
zhpenkov: nice catch!

PR #5853, opened 2026-03-26
Themes oklch

TeyaVes:


inikolova:
EDIT: It appears the colors have been updated.

~~Could we please double check if the new values of the following colors are intended:~~
~~- Bootstrap - > secondary-on-subtle;~~
~~- Material -> primary, primary-subtle-hover, primary-subtle-active;~~

inikolova (approved):

zhpenkov (approved):

tsvetomir:
Please note that this is a breaking change for the color parsing in kendo-drawing. The dataviz components will fail to initialize with the `oklch()` color notation when attempting to parse theme colors.

We must add a compatible parser and all peer dependencies on kendo-drawing must be updated before this can be released as an official version of the themes.

PR #5858, opened 2026-03-30
chore: migrate to nx-release

epetrow:
[FE-698](https://progresssoftware.atlassian.net/browse/FE-698)

[FE-698]: https://progresssoftware.atlassian.net/browse/FE-698?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/publish_npm.yml
```diff
@@ -76,29 +86,76 @@ jobs:
...
-      - name: Lerna publish (stable)
+      - name: Validate stable version
+        if: env.RELEASE_TYPE == 'stable'
+        run: |
+          RELEASE_VERSION_INPUT="${{ inputs.version }}"
+
+          if [[ -z "$RELEASE_VERSION_INPUT" ]]; then
+            echo "Missing version input. Stable releases require an explicit version tag (example: v14.1.1)."
+            exit 1
+          fi
```
zhpenkov: Can we simplify some of the input logic if we can

If we make the version required, we might remove that.

zhpenkov: nvm dev wont work

.github/workflows/publish_npm.yml
```diff
@@ -76,29 +86,76 @@ jobs:
...
+
+          if [[ ! "$RELEASE_VERSION_INPUT" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "Invalid stable version '$RELEASE_VERSION_INPUT'. Use a v-prefixed tag, for example: v14.1.1"
+            exit 1
+          fi
```
zhpenkov: I am not a fan of implicitly typing the next version when releasing stable...

Shouldn't Nx analyze commits since the last release tag and determine the next version

epetrow: This was needed because of a caveat about NX 22's default `useCommitScope: true`. Because `npx nx show projects` returns the full npm package names (`@progress/kendo-theme-core`, not `core`), scopes like `fix(core):` never match and every commit is treated as "indirect" - capped at `patch`. However it seems like there is a way around that by expanding `conventionalCommits: true` to an object with `useCommitScope: false`, so I added that and removed the explicit need to manually specify a stable version tag.

package.json
```diff
@@ -85,7 +85,9 @@
     "release:dry": "npx nx release version --dry-run",
     "release:version": "npx nx release version",
     "release:changelog": "npx nx release changelog",
-    "release:publish": "npx nx release publish"
+    "release:publish": "npx nx release publish",
+    "release:preminor": "npx nx release version --specifier preminor",
```
zhpenkov: do we use that

epetrow: It's a legitimate helper, but we could remove it

tsvetomir (approved):
Looks good, let's give it a shot.

TeyaVes (approved):

PR #5859, opened 2026-03-31
Xs button

inikolova:
Design - https://www.figma.com/design/AhuFYDlfLDAjyzf9ZZJdgw/Button?node-id=9945-341&m=dev

Design issue - https://progresssoftware.atlassian.net/browse/UX-542

In this branch, we introduced a new xs size for both the Button and Menu Button components. This smaller size is intended for use in complex, multi-layered interfaces where space is limited and a more compact control is needed.

After sync with @lkarakoleva, below is a list of components where the xs button should be used:
* Dialog - close button;
* Window - close, minimize, maximize buttons;
* Chat - attachment - command menu and remove button; pinned toolbar close button; message actions toolbar; quick reply remove button; failed message retry button;
* DockManager - pin/unpin; close button; 
* Upload - remove and retry buttons;
* Tabstrip - close button;
* TimeLine - card - expand / collapse button;

**Note:** In some cases, we did not use the actual Button component—only the styles were adjusted to match the xs appearance.
* Grid - filter menu; column menu; 
* Treelist - column menu; filter menu;
* Gantt - column menu; filter menu;


tests/_output/bootstrap/dialog/dialog.png
zhpenkov, line 1: This might be expected
In Bootstrap, the close button looks slightly off-center from the top/right corner.

tests/_output/bootstrap/tabstrip/tabstrip-closable.png
zhpenkov, line 1: Here, there is little padding between the x and the right border.

Image

tests/_output/bootstrap/dialog/dialog.png
inikolova: The button hasn’t been moved vertically (neither up nor down). I also inspected the rendered HTML and CSS, and it is positioned in the vertical center.

tests/_output/bootstrap/tabstrip/tabstrip-closable.png
inikolova: After sync with @lkarakoleva, we agreed to add margin-inline-end to the actions inside the TabStrip items

Juveniel:
is there a follow-up task logged for configuring the described instances for the release ?

inikolova:
> is there a follow-up task logged for configuring the described instances for the release ?

An issue has been created - https://github.com/telerik/kendo-themes-private/issues/497

zhpenkov (approved):
Looking good!

packages/core/scss/components/grid/_layout.scss
```diff
@@ -1169,6 +1166,21 @@
...
+            .k-grid-filter,
+            .k-header-column-menu,
+            .k-grid-header-menu {
+                padding: $kendo-button-xs-padding-y $kendo-button-xs-padding-x;
+                width: $kendo-button-xs-calc-size;
+                height: $kendo-button-xs-calc-size;
+                line-height: $kendo-button-xs-line-height;
+                border-radius: k-border-radius(md);
+
+                > .k-svg-icon > svg {
```
epetrow: This selector will fail for angular, because of the icon wrapper that they use

packages/core/scss/components/timeline/_layout.scss
```diff
@@ -206,7 +206,6 @@
             }
 
             .k-event-collapse {
-                margin-block: calc( #{$kendo-icon-spacing} * -1 );
```
epetrow, line 209: Why are we removing the block margin here?

epetrow (changes requested):

packages/core/scss/components/grid/_layout.scss
inikolova: Thank you for this note, the selector has been updated

packages/core/scss/components/timeline/_layout.scss
inikolova: This negative margin is no longer necessary, as it affects the button height. Now that we have a smaller size (xs), I think we can remove it—the button remains properly aligned without it.

epetrow (approved):

PR #5863, opened 2026-04-01
Revamp radii

epetrow:
https://progresssoftware.atlassian.net/browse/FE-718


inikolova (approved):

packages/classic/scss/core/border-radii/index.scss
```diff
@@ -1,30 +1,34 @@
...
+/// @group radii
+$kendo-border-radius: .25rem !default;
 
 /// The none border radius used across the Components.
 /// @group radii
-$kendo-border-radius-none: map.get($kendo-spacing, 0) !default;
+$kendo-border-radius-none: 0px !default;
 /// The extra small border radius used across the Components.
 /// @group radii
-$kendo-border-radius-xs: map.get($kendo-spacing, 1px) !default;
```
zhpenkov, line 9: In Classic, Gluent, Material, Default 
$kendo-border-radius-xs is `1px` hardcoded

In Bootstrap, it scales based on the $kendo-border-radius.

is this expected

packages/classic/scss/core/border-radii/index.scss
```diff
@@ -1,30 +1,34 @@
 @use "sass:map";
-@use "../spacing/index.scss" as *;
+
+/// The base border radius used across the Components. Changing this value
+/// scales sm through xxxl proportionally.
+/// @group radii
+$kendo-border-radius: .25rem !default;
```
zhpenkov: I am thinking about naming streamline here

Currently, in the spacing PR, it is 
`kendo-spacing-step`

Here it is, `kendo-border-radius`
Namings do not match

I am thinking either

`--kendo-border-radius-step`
`--kendo-spacing-step`

Or make them even more general, as the idea is to generally control the representative map. And for the spacing scenario, it is a step, but here it is more like a ratio (per theme).

`--kendo-border-radius`
`--kendo-spacing`

packages/material/scss/core/border-radii/index.scss
```diff
@@ -1,30 +1,34 @@
 @use "sass:map";
-@use "../spacing/index.scss" as *;
+
+/// The base border radius used across the Components. Changing this value
+/// scales sm through xxxl proportionally.
```
zhpenkov: Can we put the description in one row? I'm not sure if the extra `</br>` will be parsed by the Design System

packages/classic/scss/core/border-radii/index.scss
TeyaVes: +1 the kendo-border-radius-xs should again be relative to the kendo-border-radius. Imo, the only ones that should be hardcoded are none and full

packages/classic/scss/core/border-radii/index.scss
epetrow: The only concern I had about changing `$kendo-spacing-step -> $kendo-spacing` is that it already is the name of the map. But the suggestion is valid, I lean more towards `--kendo-border-radius, --kendo-spacing` however I do not know if we would be okay changing the map to `$kendo-spacings` for example.

packages/classic/scss/core/border-radii/index.scss
epetrow: We do this for the spacing, however don't we?

TeyaVes: yes, but in the spacing we have k-spacing-1px, implying that it's always 1px. In the border-radius we have xs, meaning that if the border-radius increases in general, the xs should increase as well.

epetrow: Actually disregard

zhpenkov (approved):
Looking Great!

PR #5865, opened 2026-04-02
SKILL.md

kirchoni:
Jira: [FE-1204](https://progresssoftware.atlassian.net/browse/FE-1204)

---

Draft: things are subject to change; please let me know what you think!

---
This PR introduces a [Agent skill](https://agentskills.io/home) for agents working with the kendo-themes.

It begins with an introduction SKILL.md with a basic Overview and Getting Started. Then briefly covers customization and links to deep-dives per individual modules (colors, typography, spacing etc.) referenced in separate articles to enable lazy-reading them only when necessary.

The `tailwind-integration.md` article covers basic concept of syncing tailwind `@theme` variables with kendo-theme variables.

> [!NOTE]
> I've intentionally skipped mentioning the kendo css-utils. My reasoning is that llms are **_heeeavily_** biased towards using tailwind, and since this SKILL.md targets agent-development specifically, it's no use to "fight the current" but simply embrace their preferences.

Another article is the `components.md` article. Which covers targeting, overriding and/or extending a component. To provide further meta information to agents i've added two helper scripts `list_components` which would list all available components for styling and `get_components` allowing agents to get detailed info (className, options, states) for one-or-many components. **_Dumber_** agents are free to read the whole `components.json` (generated by the newly introduced `docs:components` command) - while sub-optimal, it's still compact enough to not pollute the context window, but still provide the needed information to style a component.

> [!NOTE]
> I've intentionally skipped mentioning per-component css variables. First issue is they are only applicable to sass codebases, thus i've documented a framework-agnostic approach of styling the components through official selectors + state + option definitions. Second is they are **_waaay_** more than the components, and a lazy-reading mechanism is required to make it sustainable to read them.

At the bottom of each article, we reference a "learn-more" article from our [design-system docs](https://www.telerik.com/design-system/docs/).

## How to test?

1. Copy the content of the `skills` folder into an agent-specific folder of your test project, e.g.:
- `skills/kendo-themes/*` -> `.claude/skills/kendo-themes/*`
- `skills/kendo-themes/*` -> `.agents/skills/kendo-themes/*`
- `skills/kendo-themes/*` -> `.github/skills/kendo-themes/*`

2. Prompt....simply prompt. As long as the skill is loaded (look for a `X skills loaded` message in your agent of choice) the agent should automatically refer to the skill and/or any of it's sub-articles.

## Evals

The PR hosts a couple of [evals](https://github.com/telerik/kendo-themes/tree/skills/skills/kendo-themes/evals), I didn't want to over-test this as it's still a simulated environment and the non-deterministic behavior of agents does not give me enough confidence that if the evals pass everything is working, but still provide some visibility of what to expect from the skill compared to a "without skills" baseline.

[FE-1204]: https://progresssoftware.atlassian.net/browse/FE-1204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


skills/kendo-themes/references/colors.md
```diff
@@ -0,0 +1,176 @@
+# Colors
+
+The color system is the most commonly customized sub-system. It defines semantic color tokens used across all components.
+
+For most customization work, start with **semantic color tokens**, **CSS custom properties**, and **swatches**. Treat palettes as an advanced Sass building block that feeds semantic tokens, not as the primary API most consumers should edit directly.
```
TeyaVes: we have some changes in the color system and we are no longer using palettes, rather relative oklch colors. The palettes are deleted from the code.

skills/kendo-themes/references/colors.md
```diff
@@ -0,0 +1,176 @@
...
+
+/* Scoped dark mode */
+.dark-mode {
+  --kendo-color-app-surface: #1e1e1e;
+  --kendo-color-on-app-surface: #ffffff;
+  --kendo-color-surface: #2d2d2d;
+}
+```
+
+### Sass: Auto-Expand from a Single Color (Easiest)
```
TeyaVes: With the move to OKLCH, we should rewrite this section to reflect the new approach. Instead of recommending k-generate-color-variations() as the primary way to customize colors, we should highlight the relative color syntax we've introduced in the OKLCH color space. The key point is that users now only need to set the base color (e.g., primary, secondary) and all derived variants — hover, active, on-color, subtle, emphasis — are automatically computed relative to it:
```
    primary: oklch(69.85% 0.1923 27.19deg),
    primary-hover: #{"oklch(from var(--kendo-color-primary) calc(l - 0.044) calc(c - 0.012) h)"},
    primary-active: #{"oklch(from var(--kendo-color-primary) calc(l - 0.088) calc(c - 0.027) h)"},
    primary-emphasis: #{"oklch(from var(--kendo-color-primary) calc(l * 0.665 + 0.33) calc(c * 0.61) h)"},
```



 No Sass function call or theme parameter is needed; the variations are handled natively in CSS at runtime.

skills/kendo-themes/references/colors.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Constants
+
+| Token | CSS Variable | Value |
+|---|---|---|
+| white | `--kendo-color-white` | `#ffffff` |
+| black | `--kendo-color-black` | `#000000` |
+| transparent | `--kendo-color-rgba-transparent` | `rgba(0, 0, 0, 0)` |
+
+### Source Palettes (Advanced Sass)
```
TeyaVes: remove the palette section here as well

skills/kendo-themes/references/components.md
```diff
@@ -0,0 +1,240 @@
...
+  "_option": ["name", "values"],
+  "components": {
+    "Button": [
+      "k-button",
+      ["active", "disabled", "focus", "hover", "selected"],
+      [
+        ["fillMode", ["clear", "flat", "link", "outline", "solid"]],
+        ["rounded", ["full", "large", "medium", "none", "small"]],
+        ["size", ["large", "medium", "small"]],
+        ["themeColor", ["base", "dark", "error", "info", "inverse", "light", "primary", "secondary", "success", "tertiary", "warning"]]
```
TeyaVes: Remove the light/dark theme colors here, which will be deprecated for the next release

skills/kendo-themes/references/elevation.md
```diff
@@ -0,0 +1,58 @@
+# Elevation
```
TeyaVes, line 1: Just a note here- this should also be rewritten with the elevation changes coming in the next release (decreasing the elevation levels from 9 to 5)

skills/kendo-themes/references/radii.md
```diff
@@ -0,0 +1,42 @@
+# Border Radii
+
+The border radius system provides a consistent rounding scale used by all components.
+
```
TeyaVes, line 4: We should add a section here documenting the new `--kendo-border-radius base` variable. Setting this single value automatically scales all border-radius sizes (xs through xxxl), since each is defined as a multiplier relative to it.
```
$kendo-border-radius-base: .25rem !default;
$kendo-border-radius-xs: calc(var(--kendo-border-radius-base) * .25) !default;
$kendo-border-radius-sm: calc(var(--kendo-border-radius-base) * .5) !default;
$kendo-border-radius-md: var(--kendo-border-radius-base) !default;
$kendo-border-radius-lg: calc(var(--kendo-border-radius-base) * 1.5) !default;
$kendo-border-radius-xl: calc(var(--kendo-border-radius-base) * 2) !default;
$kendo-border-radius-xxl: calc(var(--kendo-border-radius-base) * 3) !default;
$kendo-border-radius-xxxl: calc(var(--kendo-border-radius-base) * 4) !default;
```

skills/kendo-themes/references/spacing.md
```diff
@@ -0,0 +1,77 @@
+# Spacing
```
TeyaVes, line 1: Add a section for the k-spacing-base variable used for the whole kendo-spacing map:

```
$kendo-spacing-base: 0.25rem !default;

$default-spacing: (
    0.5: calc(0.5 * var(--kendo-spacing-base)),
    1: calc(1 * var(--kendo-spacing-base)),
    1.5: calc(1.5 * var(--kendo-spacing-base)),
    ...
)
```

skills/kendo-themes/scripts/get_components.mjs
zhpenkov, line 1: Should we put the scripts in a `/scripts` folder?

skills/kendo-themes/references/spacing.md
```diff
@@ -0,0 +1,77 @@
...
+  $kendo-spacing: (
+    4: 1.2rem,
+    8: 2.5rem
+  )
+);
+```
+
+## Available Tokens
+
+| Key | Default Value | CSS Variable |
```
TeyaVes, line 15: After the changes in the spacing, we now have 24 spacing tokens instead of 30

kirchoni:
Thanks for the feedback @TeyaVes, @zhpenkov ...and @copilot.

I will hold on this PR until all the upcoming changes land in `develop` and revisit the whole skill at once. 

Apart of the existing content, do you feel something is `missing` from it? something important that is fundamental for consuming/using the kendo themes? (note, i've intentionally skipped kendo utils and component sass vars)?

skills/kendo-themes/references/layout.md
zhpenkov, line 1: this is part of the utils package actually

skills/kendo-themes/SKILL.md
```diff
@@ -0,0 +1,250 @@
...
+**CSS Custom Properties** (runtime, no build step):
+
+```css
+:root {
+  --kendo-color-primary: #your-brand-color;
+  --kendo-color-on-primary: #ffffff;
+  --kendo-spacing-4: 1.2rem;
+  --kendo-font-size: 15px;
+  --kendo-border-radius-md: 6px;
+}
```
zhpenkov: Here we might also mention the newly added 
base variables for spacing and radius

--kendo-spacing-base
--kendo-border-radius-base

as well as the relative colors behaviour

--kendo-color-primary => modifies --kendo-color-on-primary and otherrs as well

skills/kendo-themes/SKILL.md
```diff
@@ -0,0 +1,250 @@
...
+```
+
+```scss
+// Sass
+@use "sass:map";
+@use "@progress/kendo-theme-default/scss/color-system/_functions.scss" as fn;
+@use "@progress/kendo-theme-default/scss/all.scss" as * with (
+  // Auto-expand: one base color → all 10 semantic variants
+  $kendo-colors: fn.k-generate-color-variations("primary", #0058e9, "default")
+);
```
zhpenkov: Not sure if this will work.

Example recommended way:

```scss
@use 'sass:map';
@use '@progress/kendo-theme-default/scss/index.scss' as *;

$kendo-colors: map.merge(
  $kendo-colors,
  (
    app-surface: #f8f8f8,
    on-app-surface: #111111
  )
);
// Use the 'k-generate-color-variations' function to generate
// all shades of the 'primary', 'secondary' and the 'base' color.
$kendo-colors: map.merge($kendo-colors, k-generate-color-variations("primary", #9c27b0, "default"));
$kendo-colors: map.merge($kendo-colors, k-generate-color-variations("secondary", #e51a5f, "default"));
$kendo-colors: map.merge($kendo-colors, k-generate-color-variations("base", #eaeaea, "default"));

@include kendo-theme--styles();
```

kirchoni: that would be dropped after the color changes, right? no need to address it if we're to remove the whole k-generate... stuff?

skills/kendo-themes/SKILL.md
kirchoni: thats valid. one var from each system, right? preferring bases where it's applicable.

skills/kendo-themes/references/layout.md
kirchoni: we don't have ANY breakpoints inside the theme?

skills/kendo-themes/components.json
zhpenkov, line 1: This should be kept up to date, otherwise it will drift from the source of truth.

It is generated by a function, so updating it is easy, but it is still manual work, and most importantly, it needs to be implicitly run and committed by a person.

To avoid staleness of the file, we might consider having a check to see if there are changes in this file with the recent commits (like the check for the Sass docs - docs:check command)

Or directly run it on ci and automatically commit if there are changes like the visual previews

kirchoni: It should be automatically ran when someone `npm run docs`. Do we run this command often/automatic?
https://github.com/telerik/kendo-themes/blob/05b9d82c325c5a49be680e878d2ea5153987adb6/package.json#L68-L70

skills/kendo-themes/SKILL.md
zhpenkov, line 1: On a side note, what will be the distribution model for the skills?

skills/kendo-themes/components.json
zhpenkov: Yes, it's run manually when there are changes in the sassdoc annotations. it generates the docs and they are committed.

`docs:check` is a safeguard ran on push that prevents a dev to "not commit" those docs.

Might enforce such for this components.json as well.

skills/kendo-themes/references/tailwind-integration.md
```diff
@@ -0,0 +1,70 @@
...
+If your Tailwind `@theme` already defines your design tokens, override Kendo CSS custom properties to follow them:
+
+```css
+:root {
+  --kendo-<token>: var(--<tailwind-token>);
+}
+```
+
+Use the Semantic Mapping Guide below (read right-to-left: find the Kendo token you want to override, point it at your Tailwind equivalent).
+
```
zhpenkov, line 32: Seems reasonable!
Keep in mind that this will only work if 
This configuration is imported after the theme import (whatever that is).

Should we add a line here for that?

skills/kendo-themes/references/tailwind-integration.md
```diff
@@ -0,0 +1,70 @@
...
+| --------------- | ------------------------------ | ------------------------- |
+| `background`    | `--kendo-color-app-surface`    | App-level background      |
+| `foreground`    | `--kendo-color-on-app-surface` | Primary text color        |
+| `muted`         | `--kendo-color-subtle`         | De-emphasized content     |
+| `card`          | `--kendo-color-surface`        | Card/component background |
+| `border`        | `--kendo-color-border`         | Default border color      |
+| `primary`       | `--kendo-color-primary`        | Brand/action color        |
+| `destructive`   | `--kendo-color-error`          | Destructive actions       |
+| `accent`        | `--kendo-color-tertiary`       | Accent/highlight          |
+| `ring`          | `--kendo-color-primary`        | Focus ring color          |
```
zhpenkov, line 59: Rather than the "Kendo Equivalent" (which is definitive), the "Tailwind Token" part here could be anything, depending on the app.

Should we add a line stating that these are only examples to avoid enforcing those specific tokens (background, foreground, etc.)?

skills/kendo-themes/references/radii.md
```diff
@@ -0,0 +1,41 @@
...
+| Token | Multiplier | Default (`base = 0.25rem`) |
+|---|---|---|
+| `xs` | `base × 0.25` | `~0.063rem` |
+| `sm` | `base × 0.5` | `0.125rem` |
+| `md` | `base × 1` | `0.25rem` |
+| `lg` | `base × 1.5` | `0.375rem` |
+| `xl` | `base × 2` | `0.5rem` |
+| `xxl` | `base × 3` | `0.75rem` |
+| `xxxl` | `base × 4` | `1rem` |
+| `full` | `base × 9999` | `9999px` (pill) |
```
zhpenkov, line 18: This fixed multiplier is different for each theme

For example, in Bootstrap, lg is not `base x 1.5`, but another value.

skills/kendo-themes/references/layout.md
zhpenkov: We don't

They come from the "responsive-layout" utilities from the utils package.

skills/kendo-themes/references/iconography.md
```diff
@@ -0,0 +1,89 @@
...
+
+| Class | Size | Scale |
+|---|---|---|
+| `.k-icon-xs` | 12px | 75% of base |
+| `.k-icon-sm` | 14px | 87.5% of base |
+| `.k-icon-md` | 16px | 100% (default) |
+| `.k-icon-lg` | 20px | 125% of base |
+| `.k-icon-xl` | 24px | 150% of base |
+| `.k-icon-xxl` | 32px | 200% of base |
+| `.k-icon-xxxl` | 48px | 300% of base |
```
zhpenkov, line 33: Those sizes actually come from the component's API; should we include them here?

_edited_
same for theme colors

skills/kendo-themes/references/iconography.md
```diff
@@ -0,0 +1,89 @@
...
+
+All sizes are calculated from the base `--kendo-icon-size` (default: `16px`).
```
zhpenkov, line 35: I do not think this works as expected.
We have it, but it is not used at the proper places so it needs revamp.
might exclude it for now.

tl;dr, chaing --kendo-icon-size wont change all icon sizes

skills/kendo-themes/references/colors.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Customization
+
+Override color tokens using the CSS or Sass patterns from the [main skill](../SKILL.md#customization). Recommended order:
+
+1. **CSS custom properties** for runtime theming and scoped overrides.
+2. **Swatches** when an existing color variant already matches your needs.
+3. **Sass `$kendo-colors` overrides** for compile-time control over semantic tokens.
+
+### OKLCH Relative Colors
```
zhpenkov, line 22: might add an important note that this is supported after v14.0.0 of the themes

skills/kendo-themes/references/colors.md
```diff
@@ -0,0 +1,127 @@
...
+### Constants
+
+These are Sass-only constants available in core sources. They are **not** exported as `--kendo-color-*` CSS custom properties.
+
+| Token | Sass Constant | Value |
+|---|---|---|
+| white | `$kendo-color-white` | `#ffffff` |
+| black | `$kendo-color-black` | `#000000` |
+| transparent | `$kendo-color-rgba-transparent` | `rgba(0, 0, 0, 0)` |
+
```
zhpenkov, line 89: That's a nice addition, but I do not think we should complicate it with colors that will always be constant (there is no difference between $kendo-color-black and #000000, no need for extra abstraction for the  LLMs

skills/kendo-themes/references/colors.md
```diff
@@ -0,0 +1,127 @@
...
+| *(base)* | Main color | `--kendo-color-primary` |
+| `-hover` | Hover state | `--kendo-color-primary-hover` |
+| `-active` | Active/pressed state | `--kendo-color-primary-active` |
+| `-emphasis` | High emphasis variant | `--kendo-color-primary-emphasis` |
+| `-on-subtle` | Text on subtle bg | `--kendo-color-primary-on-subtle` |
+| `on-{name}` | Text on main color | `--kendo-color-on-primary` |
+| `-on-surface` | Text on surface | `--kendo-color-primary-on-surface` |
+
+**Semantic color names:**
+`base`, `primary`, `secondary`, `tertiary`, `info`, `success`, `warning`, `error`, `light`, `dark`, `inverse`
```
zhpenkov, line 70: A note that we might remove light/dark from here if they are removed from the theme in a future version

skills/kendo-themes/references/colors.md
```diff
@@ -0,0 +1,127 @@
...
+--kendo-color-<role>-hover: oklch(from var(--kendo-color-<role>) calc(l - 0.044) calc(c - 0.012) h);
+--kendo-color-<role>-active: oklch(from var(--kendo-color-<role>) calc(l - 0.088) calc(c - 0.027) h);
+--kendo-color-<role>-emphasis: oklch(from var(--kendo-color-<role>) calc(l * 0.665 + 0.33) calc(c * 0.61) h);
+```
+
+Updating a base color propagates the change to all its derived states without any additional overrides.
+
+### Sass Overrides
+
+The `$kendo-colors` map merges with defaults (`map.merge`) — specify only the tokens you want to change. Use `k-generate-color-variations(<role>, <color>, "default")` to auto-expand one base color into all 10 semantic variants at compile time. For precise control over individual states, override `$kendo-colors` directly.
```
zhpenkov, line 36: Maybe it is important here to note that the "default" parameter is the default theme, not a default function preset.

The map can also be configured without the function

```scss
@use theme * with (
  kendo-colors: (
    primary: blue
    on-primary: white
  )
)
```

And after v14, the primary will extend all other relative to it

```scss
@use theme * with (
  kendo-colors: (
    primary: blue
  )
)
```

will also change `on-primary`

skills/kendo-themes/references/colors.md
kirchoni: adding per-version instructions for the skill might bloat it. i'm leaning towards versioning the theme itself. e.g. latest skill = latest theme. Also we didnt have a skill prior to v14 so there's that

kirchoni:
Thanks Team for the review, pinging again @telerik/kendo-front-end as i've caught up to the v14 changes and hopefully addressed all comments 🤞

skills/kendo-themes/references/colors.md
```diff
@@ -0,0 +1,162 @@
...
+
+**Material:** `material-main`, `material-main-dark`, `material-2`, `material-2-dark`, `material-arctic`, `material-lime`, `material-lime-dark`, `material-pacific`, `material-pacific-dark`, `material-sky`, `material-sky-dark`, `material-smoke`, `material-nova`, `material-eggplant`, `material-burnt-teal`, `material-aqua-dark`
+
+**Classic:** `classic-main`, `classic-main-dark`, `classic-green`, `classic-green-dark`, `classic-silver`, `classic-silver-dark`, `classic-opal`, `classic-opal-dark`, `classic-lavender`, `classic-lavender-dark`, `classic-metro`, `classic-metro-dark`, `classic-moonlight`, `classic-uniform`
+
+**Fluent:** `fluent-main`, `fluent-main-dark`, `fluent-1`, `fluent-1-dark`
+
+## Data Visualization (Series)
+
+A base `series` token controls the starting point for chart colors. Six named series (`a` through `f`) are derived from it, each with 5 variants:
```
zhpenkov: Should we mention the series CSS var

`--kendo-color-series`

like in the colors section above:
_"**CSS custom properties** — Runtime-accessible `--kendo-color-{name}` variables generated from those semantic tokens"_

Also, should we move this series context to the colors section to avoid confusion?

___

Series relative colors are "two level customization".
We have documented it this way to avoid confusion:
https://www.telerik.com/design-system/docs/themes/kendo-themes/meridian/customization/#data-visualization-series-colors

Should we mention it here as well?

skills/kendo-themes/references/elevation.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Usage Guidelines
+
+- Level 1–2: Subtle depth for cards, inputs, and inline elements.
+- Level 3–4: Popovers, popups, and overlays.
+- Level 5: Modals and high-priority layers.
+
+### Elevation Background
+
+In dark mode, themes also apply an `elevation-bg` function that subtly increases lightness at higher elevation levels (using OKLCH color math). This is automatic and does not require consumer configuration.
```
zhpenkov: That is currently only for Meridian and is an implementation detail of the theme.
Should we mention it on the skill?

skills/kendo-themes/references/radii.md
```diff
@@ -0,0 +1,63 @@
...
+}
+```
+
+### Sass — override at build time
+
+```scss
+@use "@progress/kendo-theme-meridian/scss/index.scss" as * with (
+    $kendo-border-radius-base: 0.5rem
+);
+```
```
zhpenkov, line 58:
```scss
@use "@progress/kendo-theme-meridian/scss/index.scss" as * with (
    $kendo-border-radius-lg: 0.75rem,
    $kendo-border-radius-full: 50%,
);
```

might include this as well

To have a full 1:1 coverage of "X done in CSS" and "X done in Sass."

zhpenkov: Could you also check the other .md for this 1:1 rule ?

skills/kendo-themes/references/spacing.md
```diff
@@ -0,0 +1,80 @@
...
+}
+```
+
+### Sass — Override at Build Time
+
+```scss
+@use "@progress/kendo-theme-meridian/scss/index.scss" as * with (
+  $kendo-spacing-base: 0.3rem
+);
+```
```
zhpenkov, line 73: 1:1 css/sass rule

skills/kendo-themes/references/typography.md
```diff
@@ -0,0 +1,114 @@
...
+### Sass
+
+Configure the corresponding `$kendo-` variables in the theme `@use ... with ()` clause:
+
+```scss
+@use "@progress/kendo-theme-meridian/scss/index.scss" as * with (
+  $kendo-font-family: "Inter", sans-serif,
+  $kendo-font-size: 1rem
+);
+```
```
zhpenkov, line 117: Add `$kendo-font-weight: 400;` in the sass section
for consistency with the CSS example

Also, might add a variable from each typography group in the customization examples
both CSS and Sass

Just for reference.

skills/kendo-themes/SKILL.md
```diff
@@ -0,0 +1,205 @@
...
+
+```scss
+@use "@progress/kendo-theme-meridian/scss/index.scss" as *;
+
+@include kendo-button--styles();
+@include kendo-grid--styles();
+@include kendo-combobox--styles();
+```
+
+Each component exposes a `kendo-{component}--styles()` mixin.
```
zhpenkov, line 77: There is a 3rd path

CSS with CDN
https://www.telerik.com/design-system/docs/themes/get-started/installation/#using-external-cdn-link

Might mention in the "CSS Import section."

kirchoni: I'm not keen on featuring the cdn. it's a workaround which brakes versioning and it's aimed to get you started in a single copy-paste - wiring up the theme should not be time-consuming if done through an agent tho, so added value of the cdn diminishes. agree?

skills/kendo-themes/references/elevation.md
kirchoni: done, removed it altogether

skills/kendo-themes/references/colors.md
kirchoni: nice feature. featured it

skills/kendo-themes/SKILL.md
zhpenkov: Okay, that's valid.

skills/kendo-themes/SKILL.md
```diff
@@ -0,0 +1,206 @@
...
+       └── Theme Package (styling layer)
+            └── Design System Tokens (colors, spacing, typography…)
+```
+
+**6 themes available** — choose the package that matches the target design language:
+
+| Theme     | Package                           | Design Language     |
+| --------- | --------------------------------- | ------------------- |
+| Meridian  | `@progress/kendo-theme-meridian`  | Modern              |
+| Default   | `@progress/kendo-theme-default`   | The default theme   |
```
TeyaVes: I am wondering if we should change the design language field of the default theme- the default theme should rather be the meridian theme from 14.0.0 onwards

skills/kendo-themes/SKILL.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Customization
+
+Two levels of customization control component appearance — broad design-wide changes and surgical per-component overrides.
+
+### 1. Global: Design Tokens
+
+Override design tokens to change the look of **all components at once**. This is the primary customization layer — colors, spacing, radii, elevation, motion, and typography.
+
+For colors, start with **semantic tokens** (`primary`, `surface`, `error`, etc.), **CSS custom properties**, and **prebuilt swatches**. Palette ramps are an advanced Sass layer behind those tokens, not the primary customization API.
```
TeyaVes: I would remove the sentence about the palettes- we no longer have them

skills/kendo-themes/SKILL.md
kirchoni: 😅 agreed. any suggestions on what it should say?

skills/kendo-themes/references/components.md
```diff
@@ -0,0 +1,212 @@
...
+```css
+.k-button {
+}
+```
+
+### State selector
+
+States are shared `k-*` classes.
+
+If the tuple contains `hover` in its state array, combine the base class and state like this:
```
TeyaVes, line 96: I am wondering whether we should add :hover and :focus pseudo-classes here?

skills/kendo-themes/SKILL.md
kirchoni: I'm also open to removing the whole column altogether if this becomes too strange to describe

TeyaVes (approved):

skills/kendo-themes/SKILL.md
TeyaVes: for Meridian we can add smth like:  the recommended theme for Telerik and Kendo UI applications
for the Default: the old default theme currently in maintenance mode (or smth in this sense)

We can also rearange them- meridian, bootstrap, material, fluent, default, classic

skills/kendo-themes/references/components.md
kirchoni: added a brief mention

skills/kendo-themes/SKILL.md
kirchoni: done! nice catch

skills/kendo-themes/SKILL.md
kirchoni: re-arranged em. but removed the `design language` column. was more confusing than helpful. material and fluent are implicit through their names

zhpenkov (approved):

PR #5874, opened 2026-04-07
chore(infra): add css compliance audit to ci

epetrow:
### CSS Compliance Audit Script
Adds a CI-ready CSS audit script that validates compiled theme CSS for:

1. **Syntax correctness** — property-value grammar validation via `css-tree` (v3.2.1) with an allowlist for known lexer gaps (`justify-content: baseline, user-select: contain`)
2. **Browser compatibility** — feature detection via `@mdn/browser-compat-data` (primary) with `caniuse-lite` fallback, checked against the repo's `browserslistrc`
Dependencies added
- `@mdn/browser-compat-data` — Mozilla's official compat dataset (pure JSON, zero transitive deps)

**Usage**
```js
node scripts/audit-css.mjs [theme|all]   # exits 1 on syntax/compat issues
```


zhpenkov:
Great one! @epetrow 

Most of the work is done programmatically from the script, not so much from the agentic skill.
As discussed offline, we might consider making this a validating script living outside .github instead, and potentially extend it to support CSS and include it in the CI.

For the skills part, we might create a skill `manage-quality` or smth simmilar, that can include agent-based validation of code quality. It can also reference such scripts that catch quality inconsistencies, but if they run on CI, this might be obsolete.

Juveniel:
Does this cover latest CSS spec ? 

For the current use case in the themes, it will be most meaningful to have a CI pipeline, where the audit ensures the generated output is compliant.

@epetrow as this is getting rather stale, let's try to move forward with the PR if it makes sense or close it.

scripts/audit-css.mjs
```diff
@@ -0,0 +1,727 @@
...
+  node audit-css.mjs all --category syntax --json
+`);
+    process.exit(0);
+}
+
+const REPO_ROOT = resolve(
+    new URL(".", import.meta.url).pathname,
+    ".."
+);
+const THEMES = ["default", "classic", "material", "fluent", "bootstrap"];
```
zhpenkov: meridian

scripts/audit-css.mjs
```diff
@@ -0,0 +1,727 @@
...
+
+// ---------------------------------------------------------------------------
+// CLI
+// ---------------------------------------------------------------------------
+
+const { values, positionals } = parseArgs({
+    options: {
+        category: { type: "string", default: "all" },
+        help: { type: "boolean", short: "h", default: false },
+        json: { type: "boolean", default: false },
```
zhpenkov: Let's simplify a bit, this is a script meant to run on CI, we might not need help or verbose, or just keep verbose by default

Can you check if we can simplify the script from code that is not essential and focus on the css audit part?

I propose a simple CSS audit tool with a brief report at the end.

Also i am wondering what features we are covering right now with this audit and how we can extend it in the future? 

Do we currently cover some of the https://jigsaw.w3.org/css-validator/ functionalities or even extend it?

scripts/audit-css.mjs
```diff
@@ -0,0 +1,727 @@
...
+        pattern: /-webkit-appearance\s*:/,
+        replacement: "appearance",
+        message:
+            "-webkit-appearance is deprecated; use unprefixed 'appearance' instead",
+    },
+    {
+        pattern: /-moz-appearance\s*:/,
+        replacement: "appearance",
+        message:
+            "-moz-appearance is deprecated; use unprefixed 'appearance' instead",
```
zhpenkov: This will get stale. Do we currently have such patterns in our code?

Might drop it if we are clear

zhpenkov (commented):
Got many warnings when I ran the script locally, are they valid?

We can use them to validate both the audit tool and our codebase.

scripts/audit-css.mjs
epetrow: The script uses `css-tree v3.2.1` for W3C grammar validation and `caniuse-lite` for browser support checkes, so it's up-to-date for spec that have reached broad browser adoption. Here's what it validates:

**Latest CSS Color Specs (CSS Color Level 4/5):**

- oklch() color function
- color-mix() for color blending
- light-dark() for light/dark mode colors
- Relative color syntax with from

**Latest Selector & Layout Features:**

- :has() pseudo-class
- :is() and :where() pseudo-classes
- CSS Grid & Logical properties
- Flexbox gap support
- Modern Architecture:

- CSS Cascade Layers (@layer)
- Container Queries (@container)
- CSS Nesting

**Modern Property Support:**

- @property custom property rules
- CSS Custom Properties (var())

zhpenkov (approved):
Let's try it out in action!

zhpenkov:
Does the CI actually fail when warnings occur? Should we include those warnings to make the CI fail?

zhpenkov (dismissed):
`_audit-css.yml` never downloads the compiled CSS artifact produced by` _compile-themes.yml`. Every `packages/*/dist/all.css` is missing when the script runs, so it hits the skip path for every theme

Screenshot 2026-06-03 at 13 43 00

Can you please check all fail / warning scenarios and whether they affect the CI as expected

TeyaVes (approved):

Issue #5890, opened 2026-04-16
[BUG] DropDownList item with no text has 0px height


IvanDanchev:
## Bug Description

Introduced in v13.1.0. An item that has no text (empty string) renders a span (.k-list-item-text) with height 0px.

## Steps to Reproduce

1. Go to https://blazorrepl.telerik.com/mgEobKlm00OJm8OW55
2. Click the DropDownList to show its list
3. Inspect the second item in the browser dev tools

Also reproducible in: https://dojo.telerik.com/ZVOXYUsb/3

## Expected Behavior

The item that has no text should have the same height as items that have text. 

## Actual Behavior

The height of the span.k-list-item-text is 0px.

### Affected Theme(s)

- [x] All themes

### Affected Component(s)

Please specify the component(s) where the bug occurs: ___________

### Affected Suites

- [x] Kendo UI for jQuery
- [x] Telerik UI for Blazor
- [x] Telerik UI for ASP.NET Core
- [x] Telerik UI for ASP.NET MVC
- [ ] likely others as well

### Browser Information

- [x] All browsers

### CodePen/StackBlitz/JSFiddle Link

https://blazorrepl.telerik.com/mgEobKlm00OJm8OW55

https://dojo.telerik.com/ZVOXYUsb/3

### Sample Code

```html
<TelerikDropDownList Data="@DropDownListData" @bind-Value="DropDownListValue" />

@code {
    private List<string> DropDownListData = new List<string>() { "Item1", string.Empty, "Item3" };

    private string DropDownListValue { get; set; } = string.Empty;

    protected override void OnInitialized()
    {
        DropDownListValue = string.Empty;
    }
}
```

## Workaround

```
<style>
    .k-list-item-text {
        min-height: 20px;
    }
</style>
```

---

epetrow:
Fixed with https://github.com/telerik/kendo-themes/pull/5897

PR #5897, opened 2026-04-23
Empty list

inikolova:
Related to https://github.com/telerik/kendo-themes/commit/ba8c78395c08e026bad13f077304a13eff994fba

Closes:
https://github.com/telerik/kendo-themes/issues/5890
https://github.com/telerik/kendo-themes/issues/3818

https://progresssoftware.atlassian.net/browse/FE-517


packages/core/scss/components/list/_layout.scss
```diff
@@ -217,11 +217,13 @@
     .k-list-item-text,
     .k-list-header-text,
     .k-list-optionlabel {
-        &::before {
+        &:empty::before {
```
epetrow: I might be wrong, but I think we could go without explicitly adding `:empty` here

packages/html/src/list/tests/list.tsx
```diff
@@ -24,13 +24,17 @@ export default () =>(
                 <ListNormal>
                     <ListItem>List Item</ListItem>
                     <ListItem>List Item</ListItem>
+                    <ListItem></ListItem>
```
epetrow: Should we consider adding some background or `hover, focus` state on the empty item since it does not easily become clear what we are testing there

packages/core/scss/components/list/_layout.scss
inikolova: We only need to apply this for cases where the list item has no existing text content. If we render the inline-block element regardless, it will disrupt the list layout and appear before the actual text, leading to incorrect alignment and rendering issues.

packages/html/src/list/tests/list.tsx
inikolova: I modified the other test where List states are presented.

tests/_output/bootstrap/list/list-states.png
zhpenkov, line 1: We might add the empty one on the top and maybe make a dashed border around it?
Instead of adding two new lists below.

zhpenkov (approved):

tests/_output/bootstrap/list/list-states.png
inikolova: I prefer not to introduce additional states that are not part of the actual list item state model, as this could create confusion at first glance. The additional lists below already clearly communicate the intent—specifically how an empty item is represented within a list.

joneff:
The bug is related to removing the unicode zero-width space some time ago -- both zero width space and empty string take up no space, but the former is considered content, while the latter is not.

And while before there had to be a `::before` element and even `:empty` for certain scenarios, now the fix could be much simpler. Something like:

```css
    .k-list-item-text,
    .k-list-header-text,
    .k-list-optionlabel {
        min-height: 1lh;
    }
```

should do the trick. The same should do the job for the other cases where the `::before` shenanigans were used.

But don't take my word for it ;)

dimodi:
Works in Chrome/Edge/Firefox on Windows.

dimodi (approved):

zhpenkov (approved):

PR #5933, opened 2026-05-14
Style scrollbars

Juveniel:
closes: https://github.com/telerik/kendo-themes-private/issues/517


zhpenkov:
Mac's "scrollbars only when scrolling" feature does not work anymore

zhpenkov:
Also, we might move this module to core/ ?
Do we have a need to expose a CSS variables as well?
For the color, thumb color, thumb color on hover?

epetrow:
> Mac's "scrollbars only when scrolling" feature does not work anymore

if you rely on any of the `::-webkit-scrollbar` pseudo-classes it turns out they do not respect system preferences such as Mac's. On the other hand using `scrollbar-color or scrollbar-width` does respect these but it seems you lose control over the borders of the scrollbar track.

zhpenkov:
This setup looks best, and all macOS-level settings work as expected!

We have some `scrollbar-color` properties defined for the drawer and card, as well as the timeline

Do we need those?

zhpenkov (approved):

TeyaVes:
> This setup looks best, and all macOS-level settings work as expected!
> 
> We have some `scrollbar-color` properties defined for the drawer and card, as well as the timeline
> 
> Do we need those?

My only concern here is that prior to the change, the drawer and timeline scrollbars were styled. Now they won't be if you haven't added the k-body class to the body. I think we should document this change.

TeyaVes (approved):

PR #5936, opened 2026-05-15
RFC: Fix on-surface color contrast using color-mix + contrast-color

kirchoni:
Note: i pushed a temp playground with other variants I explored:
https://github.com/telerik/kendo-themes/blob/rfc/on-surface-contrast-fix/playground-on-surface.html

---

The issue here is when we start changing the colors and how they translate into the outline/flat variants of the components. For the current showcase i will use the `button` and its `outline` and `flat` variants.
The default `base` color is `#f5f5f5` which results in the following buttons:
Solid + base:
image
Outline + base:
image

However, changing the base to a darker shade, e.g. `#5B5BD6`, throws it into:
Solid + base:
image

Outline + base:
image

It's even more obvious when we toggle the surface to be `dimmed`:
image

Root cause seems to be that the outline button uses `base-on-surface` which calculates based on the `base` color, yet it sits on top of a `surfaced` background (base is dark, surface is light; thus the issue):
```css
base-on-surface: oklch(
  from var(--kendo-color-base) 
  clamp(0.2621, (0.6 - l) * 99999, 0.95) 
  0.0095 248.23deg
 )
```

---

### ~~Initial approach: `contrast-color()`~~

~~Suggestion is to use both the surface AND the base to calculate the `*-on-surface` variants. Thats tricky, because we cant force the oklch to take the `l` channel from one color (the surface) and `c` and `h` channels from other (base, primary etc). To overcome this, I suggest we try out [`contrast-color`](https://caniuse.com/?search=contrast-color) to put both `surface` and `base` into the calculation, although it just became baseline across browsers, so older browsers wont work.~~
```css
/* ❌ Dropped — requires contrast-color() (baseline April 2026) */
base-on-surface: color-mix(
  in oklch, 
  var(--kendo-color-base) 30%, 
  contrast-color(var(--kendo-color-surface)) 70%
)
```
~~This worked but had two problems: (1) relies on `contrast-color()` which is very new, and (2) on dimmed mid-tone surfaces the 60/40 ratio for colored variants didnt hit AA contrast.~~

---

### second proposal: `oklch(from) + clamp()` binary flip

Shoutout to @zhpenkov for suggesting we nest `oklch(from ...)` inside `color-mix()` — this unlocked the ability to reference channels from TWO different colors without needing `contrast-color()` at all!

The trick: `oklch(from var(--surface) clamp(0, calc((0.5 - l) * 1000), 1) none none)` creates a hard black/white flip based on the surface lightness (same behavior as `contrast-color()` but using only baseline 2024 CSS). The `none none` for chroma and hue means the source colors hue is preserved perfectly during the mix — no weird color shifts.

```css
/* base-on-surface (25/75 — more neutral) */
base-on-surface: color-mix(
  in oklch,
  oklch(from var(--kendo-color-base) l c h) 25%,
  oklch(from var(--kendo-color-surface) clamp(0, calc((0.5 - l) * 1000), 1) none none) 75%
)

/* colored variants (35/65 — keeps hue, passes AA) */
primary-on-surface: color-mix(
  in oklch,
  oklch(from var(--kendo-color-primary) l c h) 35%,
  oklch(from var(--kendo-color-surface) clamp(0, calc((0.5 - l) * 1000), 1) none none) 65%
)
```

And with the updated calculations:
image


epetrow (commented):
The change to the `on-surface` colours for `primary, secondary, tertiary, info, success, error and warning` is pretty significant. Are we okay with this?

kirchoni:
@epetrow not imediately, thus the RFC title. but it's worth the attention, even if it's a major change, the current colors do NOT persist contrast through customization of the surface.

kirchoni:
Update, we've fine-tuned the color-mix percentages:
- 25% : 75% for the base 
- 35% : 65% for primary, secondary and tertiary for better contrast compliance
- 65% : 35% for warning and error so they embed more color
- 70% : 30% for info and success so they are even more _colorful_ and visually distinct between each-other

imasyuko:
Tests look just fine 🤌

kirchoni:
Oops, had to tone down the warning & error since the contrast tests were not passing (was 3.91) from 65% -> 60%

kirchoni:
@imasyuko @epetrow - folks, please review and approve/comment so i can proceed to merging it

imasyuko (approved):
Changes look great.

kirchoni:
even tho our default swatch is light/dark agnostic, we had to updated our default dark-swatch as well. thanks @zhpenkov for noting

kirchoni:
> even tho our default swatch is light/dark agnostic, we had to updated our default dark-swatch as well. thanks @zhpenkov for noting

actually, i'm reverting this. the dark swatch would ALWAYS expect a dark surface by definition. it would not need to support dark/light flipping in this case.

zhpenkov (approved):

PR #5945, opened 2026-05-19
Optimize CI

zhpenkov:
### Unit Tests — Share docs build
Builds SCSS and docs for core + all 6 themes **once** in a new `prepare-docs` job, then distributes the artifact to the 6 parallel test matrix jobs. Previously each theme job independently rebuilt core sass + docs, duplicating work 6×.

### Integration Tests — Build `sass:dist` once
Adds a `prepare-assets` job that runs the full swatch compilation once and uploads the dist output. The 5 bundler matrix jobs (gulp, parcel, vite, webpack, esbuild) now download pre-built artifacts instead of each running `sass:dist` independently.

### Firefox Caching
Caches the Firefox binary in `~/firefox` using `actions/cache` across both `_render-test-pages` and `_create-screenshots` workflows (7 jobs total). Downloads Firefox only on first cache miss instead of fetching ~80MB per job on every run.

### Decouple Screenshots from Render Test Pages
`create-screenshots` now depends only on `compile-themes`, allowing screenshots and test page rendering to run in parallel. `detect-changes` still waits for both before combining artifacts.

### Consolidate Status Check Jobs
Replaces 8 dummy status check jobs (each spinning up a runner to `echo "Done!"`) with a single `CI Gate` job that evaluates all results and reports failures clearly. Updates `set-commit-status` to reference actual workflow jobs directly.

> **Note:** Branch protection rules need to be updated to use **"CI Gate"** as the required status check.


zhpenkov:
This experiment went well, but the overall wall clock time did not change.

For this repo's CI architecture, the wall clock time is dominated by the critical path: compile-themes → screenshots → detect-changes (~10-15 min). Everything else (linting, unit tests, integration tests, docs) runs in parallel and finishes before the critical path completes

PR #5950, opened 2026-05-20
KB Navigation in Specs

zhpenkov:
cc: @Juveniel


Juveniel:
Here are a few suggestions for improvement:

1. Can we structure the data? Currently, the `attribute` field is unparsable without additional processing e.g. "aria-required=true (when required)". Instead, we can probably structure this as `{ attribute: "aria-required", value: "true", condition: "required" }`
2. Normalize keyboard key names ( e.g. Arrow Down, Down Arrow, DownArrow, Down ArrowKeys ) to match `KeyboardEvent.key` ( [ref](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values) ).
3. Tighten the `tsconfig` glob to `src/**/behavior/*.json`
4. Augment with schema and type: Add `accessibility-spec.ts` that exports the AriaRule and KbRule as a `AccessibilitySpec` interface and then add those to the `KendoComponent` interface so those can be type checked. Add `accessibility.schema.json` and derive all specs from it e.g. `button.accessibility.json`
5. Consider lint/asserton based on the above, so that we can catch drifts in the specs.

zhpenkov:
Hey @telerik/kendo-front-end, opening this for an early look before we get further into it.

The main idea was to bring keyboard navigation specs into the component specs themselves.
All the `keyboard-navigation.json` files from the [web-components-ux repo](https://github.com/telerik/web-components-ux/tree/develop/docs/components) have been migrated

Soo... the KB nav data now lives here alongside the components, instead of in a separate repo.

To hold all of this, every component got a `behavior/accessibility.json` file with three properties:

`kbSpec` - for keyboard navigation
`ariaSpec` - for ARIA rules (already done in earlier work)
`uxSpec` - for design-side behaviour requirements, which we'll fill in as we pull from design assets.

There's a JSON Schema backing all of these and a validation script that can run against all 103 component files.

The data also feeds into auto-generated `.md` docs per component, which can be used for documentation or as context in agentic workflows.

Moving forward, we need to decide what we actually want to ship for the docs, what would be the uxSpec data, and establish the agentic workflow guidelines.

Any feedback is welcome!

zhpenkov:
Update

Added resources object in the accessibility.json for W3 ARIA links that will help:
1. Enrich the output .md for documentation use
2. Enrich the output .md for LLM use.

Also added boilerplate text for screen readers in the MD, as well as A11Y support to enrich the documentation flow.

---

**More context for the reviewer:**

What is the general idea of this effort?

The `accessibility.json` will hold info for `KB`, `UX`, `Aria`, and `Resources`.

Firstly, the `Aria` info is used for the internal themes a11y tests.

Secondly, all of the above (`KB`, `UX`, `Aria`, and `Resources`) are used to generate .md files that can be used for documentation or agentic implementation of a specific component by the product teams.

This way, we ensure 1:1 parity between themes' Aria attributes and implementation Aria attributes upon an agentic flow based on the .md-s.
And we also ensure the correct KB and UX behaviour is applied.

@telerik/kendo-front-end 

**Things left to do:**
Test and establish the recommended agentic workflow.
Extract UX info from Figma.

packages/html/src/button/behavior/accessibility-toggle.json
epetrow, line 1: This is orphaned and can be removed

package.json
```diff
@@ -35,6 +35,7 @@
     "lerna": "^9.0.0",
     "lint-staged": "^16.0.0",
     "nunjucks": "^3.2.3",
+    "ajv": "^6.14.0",
```
epetrow, line 38: I wonder if this could be a devDependency in the html package?

epetrow (commented):
Looks good overall

kirchoni:
We brought the accessibility specs into the html package so they are closer to the rendering. Moving them into a separate, non-native-for-humans .json file is a step-back imho.

The final .md files are valid tho, it's the source that bothers me! What are the options to keep the accessibility-related docs as part of the `.spec` file itself?

zhpenkov:
closing in favour of: https://github.com/telerik/kendo-themes/pull/6061

PR #5963, opened 2026-05-26
RFC: Data Theme with CSS Color Scheme 

zhpenkov:
This is yet another exploration of the https://github.com/telerik/kendo-themes/pull/5792 concept.

With the new release, we added relative colors, and right now it is easy to change a handful of parent color tokens to produce a completely new look, as the other relative colors are produced by them.

In this PR, I explored a concept of setting light and a dark variant for all the parent tokens, using the:
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value/light-dark

```css
primary: light-dark(oklch(37.17% 0.0392 257.29deg), oklch(86.90% 0.0198 252.89deg)),
primary-hover: #{"oklch(from var(--kendo-color-primary) max(0.15, calc(l + clamp(-0.0922, (0.35 - l) * 99999, 0.0922))) calc(c - 0.0023) calc(h + 2.74))"},
```

Later on, I assigned the color-schemes to specific attributes `data-theme="light/dark/system`, that can be applied in the root to achieve the desired color schema.

Check it out on the local server. WDYT?


kirchoni:
now, thats promising!

PR #5991, opened 2026-06-04
Replace hardcoded function list with Sassdoc metadata in unit tests

epetrow:
closes: https://github.com/telerik/kendo-themes-private/issues/459 and https://progresssoftware.atlassian.net/browse/FE-425


TeyaVes (approved):
We can add Copilot instructions to add /// @group functions to all newly created functions

PR #5997, opened 2026-06-05
feat(ci): replace visual snapshot commits with reg-suit diff reports

Juveniel:
Demo PR here, check it out:
https://github.com/telerik/kendo-themes/pull/6138

---

Jira: [FE-1287](https://progresssoftware.atlassian.net/browse/FE-1287)

## Summary

Replaces the current `detect-changes` job — which commits ~4,200 PNGs directly to the branch on every CI run — with **reg-suit**: baselines stored on the `gh-pages` branch, an expected/actual/diff HTML report published to GitHub Pages, and a sticky PR comment with change counts.

Introduces a unified **Playwright capture pipeline** (`scripts/capture.mjs`) used by both CI and the local dev loop, with a Docker image prepared for font-deterministic capture (Phase 2).

---

## Problem

- A broad SCSS change triggers a `chore: update visual previews` commit touching thousands of PNGs (~221 MB)
- GitHub's Files-changed view and Chrome fail to load the diff
- No visual diff visualisation — just raw binary changes

---

## Solution

```
create-screenshots (matrix, 6 themes)     ← Playwright replaces Firefox + npx snapshot
        │  artifacts: screenshots-{theme}.tar
        ▼
visual-regression (NEW — replaces detect-changes):
  reg-suit run
    ├─ keygen   : snapshot key from git history
    ├─ fetch    : pull expected baseline from gh-pages
    ├─ compare  : matchingThreshold/thresholdRate → .reg/ + out.json
    └─ publish  : push new baseline + report to gh-pages → GitHub Pages
  sticky PR comment with counts + report URL
```

---

### New files
| File | Purpose |
|---|---|
| `scripts/capture.mjs` | Shared Playwright capture module — custom HTTP server, parallel workers (CPU-adaptive, capped at 4), `waitUntil:load` + `document.fonts.ready`, CLI interface |
| `scripts/visual-local.mjs` | Local save/diff helper (`--save` seeds baseline; default run diffs vs baseline) |
| `scripts/visual-dev.mjs` | Devkit watch loop — imports `capture()`, SCSS file watcher, reg-cli diff |
| `regconfig.json` | reg-suit CI config (keygen + gh-pages plugins, AA thresholds) |
| `regconfig.local.json` | Local sandbox config (no plugins, same thresholds) |
| `Dockerfile.visual` | Node 24 + Chromium 1.60.0 + fonts — for Phase 2 font-deterministic capture |
| `.github/workflows/_visual-regression.yml` | reg-suit run + sticky PR comment |
| `.github/workflows/_build-html.yml` | Builds devkit test pages, uploads `html-dist` artifact |

### Modified files
| File | Change |
|---|---|
| `.github/workflows/ci.yml` | Add `build-html` job; replace `detect-changes` with `visual-regression`; rewire `ci-visual` + `set-commit-status` |
| `.github/workflows/_create-screenshots.yml` | Replace Firefox + `create-screenshots.sh` with Playwright + `capture.mjs` |
| `.gitignore` | Ignore `tests/_output/` and `.reg/` |
| `package.json` | Add reg-suit, reg-cli, playwright deps; add `test:visual`, `test:visual:save`, `test:visual:local`, `test:visual:dev`, `test:visual:docker` scripts |

---

## Local dev workflow

```bash
# Capture baseline
node scripts/capture.mjs --themes meridian --components button
npm run test:visual:save

# Edit SCSS, recompile, re-capture, diff
npm run sass
node scripts/capture.mjs --themes meridian --components button
npm run test:visual:local          # → .reg/report.html

# Or use the watch loop (needs npm start running)
npm run test:visual:dev button meridian
```

---

## One-time setup required

- [ ] Enable GitHub Pages: repo Settings → Pages → Source = `gh-pages` branch
- [ ] Merge to `develop` to seed the first baseline (all images will be `newItems` on the first run)

---

## TODO (follow-up PRs)

- [ ] **Propagate Playwright to the HTML test workflow** — `_test-html-spec.yml` and related HTML spec tests still use the old browser setup; unify on the same Playwright + devkit approach (FE-1290)
- [ ] **Performance optimisation** — in-page theme-swap once the `window.load` chart-color patch is extracted into a re-runnable function (FE-1288)
- [ ] **Phase 2 — Docker capture**: build and push `Dockerfile.visual` to `ghcr.io/telerik/kendo-visual:1` to eliminate font AA drift between local and CI (FE-1289)

---

## Notes

- `reg-suit run || true` — visual changes are **non-blocking**; the PR comment + report are the review surface
- Viewport fixed at **1024×768** to match the old `npx snapshot` default; screenshots are pixel-comparable within the same engine
- `build/create-screenshots.sh` is superseded but left in place until this PR is stable

[FE-1287]: https://progresssoftware.atlassian.net/browse/FE-1287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kirchoni:
@Juveniel I would need your assistance for the `gh-pages` setup (i lack the permissions to enable it)

kirchoni:
Ran a smoke test (visible SCSS change on a throwaway branch/PR) to confirm reg-suit actually flags diffs. Result: it reported everything as `new` (0 changed) instead of comparing.

Traced it to `reg-publish-gh-pages-plugin@0.2.2`'s `fetch()` — it's a no-op stub that logs a warning and returns, never retrieving the prior snapshot. This is separate from the `gh-pages` enablement blocker above — even with Pages live, this plugin only supports publish, not fetch, so comparisons would still show everything as new.

We'll need a custom fetch  step (prefferably) or a plugin swap (e.g. `reg-publish-github-plugin`, storing baselines as release assets) before this can actually catch regressions.

kirchoni:
gh-pages deploy is fixed and live. Found two more bugs while setting up Pages:
- Config key typo (`branchName` → `branch`) and missing `outDir` were silently skipping the deploy.
- `outDir: "."` collapsed the worktree's own path, destroying its `.git` link mid-deploy — moved reports to `outDir: "reports"` instead.
- No `.nojekyll`, so GitHub's Jekyll build was dropping any `_`-prefixed asset paths (404). Added it.

Verified end-to-end: https://telerik.github.io/kendo-themes/reports/ is live with all assets loading. The `fetch()` no-op from my earlier comment is still unresolved and separate.

kirchoni:
Re-ran the smoke test with the fetch() fix in place: fetch() now correctly retrieves the prior snapshot and compare() reports real diffs (Changed items: 78, not New items). Diffing itself works end-to-end.

Hit a new blocker in publish, though: reg-publish-gh-pages-plugin's git-deploy crashes (ENOBUFS) once there's an actual comparison. reg-suit's report mirrors actual/ + expected/ for every compared item (not just the diffs), so any real-diff commit is ~9,900 files / ~400MB+ vs ~4,800 files for a new-only run, which is why this didn't surface earlier. This isn't a volume/pruning problem - even a single such snapshot already breaks the plugin's naive full-worktree-checkout-then-commit approach, and it will recur on every PR with a real visual change.

kirchoni:
Update: fixed the ENOBUFS publish crash.

Root cause was payload size, not accumulation: any real (non-"new") comparison mirrors `actual/`+`expected/` for every compared item, ~9,900 files here, and `reg-publish-gh-pages-plugin`'s `git commit`/`push` against a full-history worktree checkout reliably dies with ENOBUFS on that volume — confirmed via the smoke test.

Swapped its publish step for `peaceiris/actions-gh-pages` (kept our custom `fetch()` as-is). Re-ran the smoke test: the same 9,916-file commit now pushes cleanly, report is live and reachable. One gotcha - had to pin `@v3` since the org's Actions allowlist doesn't permit `@v4` yet (run failed at startup with 0 jobs until I pinned it).

Pushed both commits to this branch. All CI checks green on the smoke test PR. I'll leave the accumulated `gh-pages` report folders (including my manual test ones) for a follow-up cleanup/retention discussion — not a blocker for merge.

kirchoni:
Final pass before tagging the team.

- CI timing: old mechanism 5m36s total, this PR ~10m10s. Fixed the checkout doing a full all-branches clone (90s → 7s) with `filter: blob:none`; pipeline now ~7m52s (-23%). Rest is real pixel-diff work the old byte-diff never did.
- Found + fixed a gh-pages leak: baseline reports (`develop` merges) had no cleanup, ~200MB/merge forever. Now prunes to just the latest.
- Considered per-theme parallel compare, skipped — ~15% more savings for a lot more moving parts. Noting as a possible follow-up, not blocking.

CI green on `ac10049`. Ready for review/merge call.

zhpenkov:
Looking great. I was just thinking...

The regressions visual tool does not actually care how the images get populated... Right ?

It just diffs whatever images there are; nothing in the reg-suit job depends on Playwright specifically.

So, why are we replacing our old image generation method with the new one?

If we use our old create-screenshots script, but feed the images to the reg-suite instead of committing them, we might simplify the approach and drop the Playwright dependency.

EDITED, PS: Potentially decreasing CI time?

kirchoni:
@zhpenkov for once, the playwright was something we were looking to do for a long time. manually navigating a firefox instance is...cumbersome.

You are right that the diffing is the main part, yet we discussed moving the images out-of-the-way and since we were refactoring the whole workflow, i suspect @Juveniel rolled with the playwright as well as a meaningful and related effort.

The CI increase time is mainly due to CPU heavy diffing (regardless of where the images came from).

.github/workflows/_create-screenshots.yml
```diff
@@ -11,24 +11,27 @@ jobs:
...
+    # Each matrix job used to run its own `playwright install --with-deps`
+    # (browser download + apt-get system deps), so every run did up to 6
+    # redundant, fully concurrent network installs - besides the waste, this
+    # made runs prone to stalling whenever the CDN/apt mirror hiccuped under
+    # that concurrency. The official Playwright image ships Chromium and all
+    # its system deps pre-baked, so no per-job installs/network calls happen
+    # at all. Keep this tag in sync with the `playwright` version in
+    # package.json - a mismatch fails fast with a clear version-mismatch
+    # error from Playwright itself.
+    container: mcr.microsoft.com/playwright:v1.62.1-noble
```
zhpenkov, line 23: here we use: `v1.62.1-noble`
but in the local docker file
it is pinned to: `playwright@1.60.0`
lets standartize this.

Also a side question:

Do we need a local docker file?
The ci uses the official Playwright container directly,
soo the Dockerfile.visual is intended for local use only.
And those file differences are intended to run on ci mainly.

zhpenkov (approved):
Tool looks great!

Headsup - there are many fix/feat ci commits that will appear on the changelog.

.github/workflows/_create-screenshots.yml
kirchoni: ye, although 1.62 gets picked up (due to `^1.60`) everytime, its good to align em.
On the docker file -ye, we need it - good luck generating images on a potential Windows machine without it.

kirchoni:
@zhpenkov nice flagging the commit messages - i've renamed em.

PR #6021, opened 2026-06-17
chore(registry): migrate to Harness artifact proxy registry

kirchoni:
## Summary

Configures the monorepo to install packages through the **Harness artifact proxy registry** instead of the public npm registry. All network traffic to `registry.npmjs.org` is blocked in this environment — only the Harness proxy is reachable.

---

### `.npmrc` (local-only, gitignored)
- **Not committed** — added to `.gitignore` to keep credentials/config out of history
- Sets Harness proxy as the default npm registry
- `@jsr` scope also routed through Harness
- `ignore-scripts=true` — lifecycle scripts blocked by default
- `min-release-age=14` — server-side age-gate enforced by Harness (14 days from first-proxy date)

### `package.json` — version pins
The Harness proxy enforces a **14-day hold** on tarballs from the date it first caches them (not from the npm publish date). Several recently-accessed packages hit this gate and were pinned to versions older than 14 days:

| Package | Pinned to | Was |
|---|---|---|
| `vitest` | `4.1.7` | `^4.1.7` |
| `sass` | `1.100.0` | `^1.100.0` |
| `esbuild` | `0.28.0` | `^0.28.0` |
| `eslint` | `10.4.1` | `^10.0.0` |
| `stylelint` | `17.12.0` | `^17.0.0` |
| `@commitlint/cli` | `20.1.0` | `^20.0.0` |
| `@commitlint/config-conventional` | `20.0.0` | `^20.0.0` |
| `@commitlint/core` | `20.1.0` | `^20.0.0` |

### `package.json` — overrides
Transitive deps also hit the age-gate, resolved via `overrides`:

| Package | Overridden to |
|---|---|
| `ast-v8-to-istanbul` | `1.0.3` |
| `@types/react` | `19.2.15` |
| `nwsapi` | `2.2.23` |
| `vite` | `8.0.14` |

### `packages/html/package.json`
- Pinned `vitest` and `@vitest/coverage-v8` to `4.1.7`

### `integrations/vite/package.json`
- Replaced `"latest"` tags (which resolved to blocked versions) with explicit pins: `sass@1.100.0`, `vite@8.0.14`

---

## ⏳ Pending — hopefully 🤞 unblocks ~2026-06-29

Two packages are present in `package.json` but **intentionally failing** — their deps were first proxied on 2026-06-15 and are under the 14-day hold:

- `sassdoc@^2.7.4` → blocked via `cdocparser` (first proxied 2026-06-15)
- `@progress/kendo-e2e@4.14.2` → blocked via `js-graph-algorithms` (first proxied 2026-06-15)

These are kept in `package.json` as **explicit markers** that `npm install` is not fully clean yet. Once the age-gate clears (~June 29), restore them by running `npm install` and verifying.

> **Impact while blocked:** `npm run docs` and `npm run test:units` will fail on fresh clones. `npm run test:create-screenshots` will also fail.

---

## ✅ Verified working

- `npm install` — 800 packages installed
- `npm run sass` — all 8 themes compile
- `npm run lint:scripts` — ESLint clean
- `npm run lint:styles` — Stylelint clean

---

## Key discovery: how the Harness age-gate works

The `min-release-age=14` setting in `.npmrc` is **ignored by npm v10+** (logged as unknown config) but is **enforced server-side by Harness**. The 14-day clock starts from when the Harness proxy **first caches** a tarball — not from the npm.com publish date. This means even packages published years ago can be blocked if they are accessed through this proxy for the first time.

> [!CAUTION]
> **This is an assumption.** We observed that packages published long before June 15 were blocked, and that blocking them resolved after pinning to slightly older versions — but we never confirmed with Harness documentation or support how the age-gate indexing actually works internally. The "14 days from first-proxy-cache-date" theory is inferred from observed behavior, not from official docs. The actual unblock date of ~June 29 is also an estimate.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ debug.log
 
 # Generated a11y test reports (not tracked; screenshots in tests/_output/ ARE tracked)
 tests/_output/*.json
+
+# Local npm registry config (not committed)
+.npmrc
--- a/integrations/vite/package.json
+++ b/integrations/vite/package.json
@@ -5,7 +5,7 @@
     "build": "vite build"
   },
   "devDependencies": {
-    "sass": "latest",
-    "vite": "latest"
+    "sass": "1.100.0",
+    "vite": "8.0.14"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,21 +19,20 @@
         "./packages/utils"
       ],
       "devDependencies": {
-        "@commitlint/cli": "^20.1.0",
-        "@commitlint/config-conventional": "^20.0.0",
-        "@commitlint/core": "^20.1.0",
+        "@commitlint/cli": "20.1.0",
+        "@commitlint/config-conventional": "20.0.0",
+        "@commitlint/core": "20.1.0",
         "@csstools/selector-specificity": "^6.0.0",
         "@eslint/js": "^10.0.1",
         "@jridgewell/source-map": "^0.3.10",
         "@mdn/browser-compat-data": "^8.0.1",
         "@nx/js": "^22.6.3",
-        "@progress/kendo-e2e": "4.14.2",
         "@stylistic/stylelint-plugin": "^5.0.1",
         "@types/node": "^25.0.3",
         "autoprefixer": "^10.4.20",
         "css-tree": "^3.2.1",
-        "esbuild": "^0.27.0",
-        "eslint": "^10.0.2",
+        "esbuild": "0.28.0",
+        "eslint": "10.4.1",
         "glob": "^13.0.0",
         "globals": "^17.3.0",
         "gulp": "^5.0.0",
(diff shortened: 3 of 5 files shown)


ivanchev (approved):

ivanchev (dismissed):
For public repos, running public runners, we'll still use the official npmjs registry, and rely on the package-lock. Make sure that you've set up the internal registry locally for each contributing user, and that the generated package-lock complies with HAR. So remove the .npmrc from the PR, and only modify the versions, so that they pass through HAR locally

kirchoni:
> For public repos, running public runners, we'll still use the official npmjs registry, and rely on the package-lock. Make sure that you've set up the internal registry locally for each contributing user, and that the generated package-lock complies with HAR. So remove the .npmrc from the PR, and only modify the versions, so that they pass through HAR locally

Thanks @ivanchev for stepping in!

If i understood correctly, the `resolved` field in the package-lock's would need to point to `registry.npmjs.org`? But then how are we supposed to maintain this lock file, when we develop against the internal registry - thus no-one can successfully re-generate with the `npmjs.org` registry?

Do you suggest we introduce a post-processing step that would have to override the `resolved` field before pushing?

This also would prevent running `npm ci` locally AFAIK

kirchoni:
Anyway, removing it for now, until further instructions

tsvetomir:
No issues with the `resolved` field, the installation will succeed even if the URL is not reachable.

kirchoni:
Thanks @tsvetomir - what about generating and maintaining the lock file in the first place then?

Juveniel (approved):

zhpenkov (approved):

tsvetomir (approved):

PR #6037, opened 2026-06-25
chore(infra): inline resolve-variables into docs script

TeyaVes:
The predocs lifecycle hook was silently skipped by Nx's run-script executor and npm's ignore-scripts config, causing stale variable values in generated documentation.


zhpenkov (approved):

zhpenkov:
just make it a chore(infra)

PR #6053, opened 2026-07-01
fix(loader): use transform-only keyframes for spinner-3 and spinner-4

zhpenkov:
## Problem

The `spinner-3` (InfiniteSpinner) and `spinner-4` keyframes animate `inset-block-start` and `inset-inline-start` — layout-triggering properties. Under heavy main-thread load (e.g. large PDF rendering in PDFViewer), layout recalculation on every animation frame causes the spinners to freeze.

## Solution

Replace layout-triggering keyframe animations with `transform: translate()` keyframes that run on the **compositor thread**, staying smooth regardless of main-thread load.

### Approach

- Two SCSS mixins (`spinner-3-keyframes`, `spinner-4-keyframes`) generate per-size keyframes using only `transform: translate()`
- Segment positions (`top`/`left`) are static — never animated
- Each segment gets its own keyframe starting at its home position, eliminating the need for `animation-delay` staggering
- Canvas rotation keyframe for spinner-4 remains unchanged (already transform-only)
- Pulsing-2 loader unchanged (already transform-only)

### Why size-specific keyframes?

`translate()` percentages are relative to the element's own size, not the parent. The travel distance depends on the container size which varies per size class (sm/md/lg), so each size needs its own set of keyframes with the correct offsets.

## What changed

| File | Change |
|------|--------|
| `packages/core/scss/components/loader/_layout.scss` | Replaced layout-triggering keyframes with transform-only ones |

## Testing

- SCSS compiles without errors
- Stylelint passes
- All three spinner variants (pulsing-2, spinner-3, spinner-4) render correctly
- Verified compiled CSS output contains proper `calc()` expressions in `transform: translate()` values

## Jira

[FE-1311](https://progresssoftware.atlassian.net/browse/FE-1311)

[FE-1311]: https://progresssoftware.atlassian.net/browse/FE-1311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kirchoni:
note: `perf` commit wont produce a patch version. changing to `fix`

kirchoni (approved):

PR #6058, opened 2026-07-02
fix(grid, table): $kendo-table-font-size and $kendo-grid-font-size overrides not working

### LINKED ISSUE #6032: [BUG] Grid font-size is not affected by $kendo-grid-font-size and $kendo-table-font-size SCSS variables
## Description

When attempting to override Kendo Grid's default font size using SCSS variables, the overrides do not take effect. The SCSS variable override method documented in the design system fails, but a CSS custom property workaround succeeds.

## Steps to Reproduce

1. Create a SCSS file with variable overrides:

```scss
@use "@progress/kendo-theme-bootstrap/scss/index.scss" as *
with (
  // GRID FONT SIZE overrides
  $kendo-grid-font-size: var(--kendo-font-size-sm),
  $kendo-table-font-size: var(--kendo-font-size-sm),
);

## Expected Behavior

- The $kendo-grid-font-size SCSS variable should reduce the grid's font size
- The $kendo-table-font-size SCSS variable should reduce the grid table's font size
- SCSS variable overrides should work as documented in the design system

### PR DESCRIPTION
**Pull request by zhpenkov**
## Description

Fixes #6032

## Root Cause

The table size system generates a `font-size` rule that **always overrides** the base `$kendo-table-font-size` value, regardless of what the user sets.

In `core/scss/components/table/_layout.scss`, the `k-when-default("md", "md")` call returns `"&,"`, producing:

```css
.k-table,          /* ← added by k-when-default (default size match) */
.k-table.k-table-md {
    font-size: var(--kendo-font-size);   /* from $kendo-table-md-font-size */
}
```

This rule appears after the base `.k-table { font-size: $kendo-table-font-size; }` in the output. With equal CSS specificity, the size-system rule always wins — silently ignoring the user's SCSS variable override.

`$kendo-grid-font-size` is also affected because the inner `.k-table` inside `.k-grid` has its own explicit `font-size`, preventing CSS inheritance from the grid.

## Fix

In all 6 theme packages (`default`, `bootstrap`, `material`, `classic`, `fluent`, `meridian`), updated the size-specific font-size variables to fall back to `$kendo-table-font-size` when it is explicitly set:

```scss
// Before
$kendo-table-md-font-size: var(--kendo-font-size) !default;

// After
$kendo-table-md-font-size: if($kendo-table-font-size != null, $kendo-table-font-size, var(--kendo-font-size)) !default;
```

This preserves the default behavior while correctly honoring user overrides via `$kendo-table-font-size` (and transitively, `$kendo-grid-font-size` now works through CSS inheritance as intended).

### DIFF
--- a/packages/bootstrap/scss/table/_variables.scss
+++ b/packages/bootstrap/scss/table/_variables.scss
@@ -49,21 +49,21 @@ $kendo-table-lg-cell-padding-y: k-spacing(2.5) !default;
 
 /// The font size of the Table for small size.
 /// @group table
-$kendo-table-sm-font-size: var(--kendo-font-size) !default;
+$kendo-table-sm-font-size: if($kendo-table-font-size != null, $kendo-table-font-size, var(--kendo-font-size)) !default;
 /// The line height of the Table for small size.
 /// @group table
 $kendo-table-sm-line-height: var(--kendo-line-height) !default;
 
 /// The font size of the Table for medium size.
 /// @group table
-$kendo-table-md-font-size: var(--kendo-font-size) !default;
+$kendo-table-md-font-size: if($kendo-table-font-size != null, $kendo-table-font-size, var(--kendo-font-size)) !default;
 /// The line height of the Table for medium size.
 /// @group table
 $kendo-table-md-line-height: var(--kendo-line-height) !default;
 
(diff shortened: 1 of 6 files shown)


zhpenkov:
duplicate of: https://github.com/telerik/kendo-themes/pull/6059

PR #6059, opened 2026-07-02
fix(grid, table): $kendo-table-font-size and $kendo-grid-font-size overrides not working

### LINKED ISSUE #6032: [BUG] Grid font-size is not affected by $kendo-grid-font-size and $kendo-table-font-size SCSS variables

zhpenkov:
Fixes #6032

## Root Cause

The table size system generates size-specific CSS rules (e.g. `.k-table, .k-table.k-table-md { font-size: ... }`) that appear **after** the base `.k-table { font-size: $kendo-table-font-size; }` rule in the compiled CSS. Because specificity is equal, the size-system rule always wins — making `$kendo-table-font-size` overrides have no effect regardless of what value the user sets.

`$kendo-grid-font-size` sets `.k-grid { font-size: ... }` correctly, but the inner `.k-table` font-size from the size system prevents CSS inheritance from reaching it.

## Fix

In each theme's `table/_variables.scss` (all 6 themes: `default`, `bootstrap`, `material`, `classic`, `fluent`, `meridian`), the size-specific font-size variables now defer to `$kendo-table-font-size` when it is explicitly set, and fall back to the CSS custom property otherwise:

```scss
// Before
$kendo-table-sm-font-size: var(--kendo-font-size) !default;
$kendo-table-md-font-size: var(--kendo-font-size) !default;
$kendo-table-lg-font-size: var(--kendo-font-size) !default;

// After
$kendo-table-sm-font-size: if($kendo-table-font-size != null, $kendo-table-font-size, var(--kendo-font-size)) !default;
$kendo-table-md-font-size: if($kendo-table-font-size != null, $kendo-table-font-size, var(--kendo-font-size)) !default;
$kendo-table-lg-font-size: if($kendo-table-font-size != null, $kendo-table-font-size, var(--kendo-font-size)) !default;
```

**Special cases:**
- The **meridian** theme uses `var(--kendo-font-size-sm)` for the `sm` variant — preserved as the fallback.
- The **fluent** theme uses CSS custom property wrapping (`var(--kendo-table-md-font-size, var(--kendo-font-size))`); the same `if()` pattern is applied while preserving this chain. `$kendo-table-font-size` and `$kendo-table-line-height` were also missing from the fluent theme entirely — they are now added with `null` defaults and forwarded to core.

## Verification

All 6 themes compile without errors.


tests/_output/meridian/table/table-list-sizes.png
kirchoni, line 1: You have `default, small, medium, large` so 4 total entries.

kirchoni (approved):
Still in draft?

PR #6060, opened 2026-07-02
fix(scheduler): fix tooltip text visibility in dark themes

### LINKED ISSUE #5472: Scheduler tooltip does not show No Events text in dark themes
Dark themes of the Scheduler does not show the Events tooltip properly:
image

https://demos.telerik.com/kendo-ui/scheduler/yearview

Same issue can be observed in the Angular demo as well:
https://www.telerik.com/kendo-angular-ui/components/scheduler

zhpenkov:
## Summary

Fixes #5472

The Scheduler tooltip "No Events" text was invisible in dark themes because `$kendo-scheduler-tooltip-bg` was set to `k-color(on-primary)`, which resolves to near-white in dark themes — matching the text color and making it invisible.

## Root Cause

`k-color(on-primary)` is intended for content placed *on top of* a primary-colored surface. In dark themes it resolves to a very light color, so using it as a background color causes text to become invisible.

## Fix

Rather than patch the value, we remove the explicit color overrides (`$kendo-scheduler-tooltip-bg`, `$kendo-scheduler-tooltip-text`) from all affected themes, letting the scheduler tooltip inherit the base `.k-tooltip` colors (`k-color(inverse)` / `k-color(on-inverse)`). These tokens are already theme-aware and produce correct contrast in both light and dark themes — no custom overrides needed.

**Files changed:**
- `packages/default/scss/scheduler/_variables.scss`
- `packages/bootstrap/scss/scheduler/_variables.scss`
- `packages/classic/scss/scheduler/_variables.scss`
- `packages/material/scss/scheduler/_variables.scss`
- `packages/meridian/scss/scheduler/_variables.scss`


zhpenkov:
The scheduler events tooltip is planned to be revamped

PR #6061, opened 2026-07-02
KB Navigation in Specs vol2

zhpenkov:
## Accessibility Specs: From JSON to Inline TypeScript

Follows up on: #5950

### Why we changed the approach

PR #5950 brought KB/ARIA specs into the HTML package via `behavior/accessibility.json` files. During review two problems were raised:

1. **Wrong home.** The JSON lives in a parallel `behavior/` subfolder, disconnected from the spec file that owns the render logic.
2. **Wrong granularity.** One JSON per folder bundles rules for all sub-components together. `BottomNav`'s JSON contained ARIA rules for `.k-bottom-nav-item` — rules that belong to `BottomNavItem`, not `BottomNav`.

Additionally, a deeper point was raised by @kirchoni: **the spec rendering IS the accessibility specification.** Validating the spec against a separate data list (ariaSpec) is circular — the spec defines what it renders, so testing it against itself adds no value.

#### `ariaSpec` removed entirely

`ariaSpec` as a static typed property on components has been removed. The accessibility story is now told in two layers:

**1. JSDoc annotations in the spec file** — structured tags that document what the component renders:

```tsx
/**
 * @aria {role="combobox"} Announces the input as a combobox widget.
 * @aria {aria-haspopup="listbox"} Indicates a listbox popup is available.
 * @aria {aria-expanded="true"|"false"} Announces popup visibility state.
 * @ux {Keyboard navigation} Arrow keys move through the suggestion list.
 */
export const Autocomplete: KendoComponent<...>

// ... component code + static assignments ...

/**
 * @keyboard {Alt + ArrowDown} Opens the suggestion list.
 * @keyboard {Escape} Closes the popup.
 *
 * @see https://www.w3.org/WAI/ARIA/apg/patterns/combobox/ WAI-ARIA Combobox Pattern
 * @see https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html WCAG 4.1.2 Name, Role, Value — combobox must have accessible name
 */

export default Autocomplete;
```

**2. WCAG compliance via axe-core** — `npm run test:a11y` validates the actual rendering against WCAG 2.x AA rules. No ARIA rule list comparison.

#### Annotation conventions

| Tag | Location | Format | Purpose |
|-----|----------|--------|---------|
| `@aria` | Component block | `@aria {attribute="value"} Description` | Documents an ARIA attribute the component renders |
| `@ux` | Component block | `@ux {Feature name} Description` | Documents a UX behavior |
| `@keyboard` | Bottom block | `@keyboard {Trigger} Description` | Documents a keyboard shortcut |
| `@see` | Bottom block | `@see https://url Display text` | WAI-ARIA spec, WCAG criterion, or reference |

All tags use the same `{label} description` format — consistent and machine-readable.

#### Sub-component ownership fixed

24 plain `.tsx` sub-component files (`WizardStep`, `TabStripItem`, `TreeviewItem`, `DrawerItem`, etc.) were promoted to `.spec.tsx` and each received their own `@aria` annotations covering only what they render. The generated docs are now **nested per sub-component**:

```
### BottomNav
#### Component Style Options  ← from Component.options
#### ARIA Attributes          ← from @aria JSDoc
#### Keyboard Navigation      ← from @keyboard JSDoc
#### UX Behavior              ← from @ux JSDoc

### BottomNavItem
#### ARIA Attributes          ← its own @aria JSDoc
```

#### ARIA annotations sourced from original JSON data

The 102 `behavior/accessibility.json` files were mined for their `ariaSpec.usage` descriptions and back-filled as `@aria` annotations across all spec files. The same information is now co-located with the rendering code that produces it.

#### Cleanup

- Deleted `behavior/accessibility.json` for 101/102 components
- Deleted `validate-a11y-json.mjs`, `schema/accessibility.schema.json`
- Removed `ajv` devDependency and `validate:a11y` npm script
- Removed `AriaRule` type and `ariaSpec?: AriaRule[]` from `KendoComponent` interface
- Removed `kbSpec`, `uxSpec`, `resources` static properties — all moved to JSDoc

## A11Y TESTS — 20× performance improvement 🚀

The test runner was completely rearchitected.

**Before:** Browser-based (Chrome via WebDriver + xvfb on CI)
- Browser startup: ~8s
- Per-template: 200ms fixed sleep + WebDriver IPC + axe-core
- Full suite: ~30–45 minutes
- CI required: virtual display (xvfb), `--affected` flag to skip unchanged components

**After:** JSDOM-based (in-process, no browser)
- No browser startup
- Per-template: ~10ms (JSDOM parse + axe-core)
- Full suite: **~5 minutes** (-83%)
- CI: no xvfb, no `--affected` needed — runs everything on every PR

Benchmark (button component, 3 templates):

| | Before | After |
|---|---|---|
| `node scripts/test-a11y-unified.mjs button` | ~15–20s | **1.9s** |

This works because `test:a11y` already disabled `color-contrast` (the one rule that requires computed CSS from a real browser). All remaining structural WCAG rules work identically in JSDOM.

CI workflow simplified: removed `xvfb` install, `xvfb-run` prefix, `--affected`, `--verbose` flags. Now runs `npm run test:a11y` directly against all components.

### What stays the same

- Doc content is identical — the generated markdown still has all ARIA, KB, UX, and resource sections
- `test:contrast` is unchanged (uses a real browser + themed CSS for color-contrast checks — separate concern)
- JSON fallback path kept in doc generator for `chart` (no spec files to migrate to)

### Trade-offs

| | JSON + ariaSpec | JSDoc annotations |
|---|---|---|
| ARIA validation | Custom JSDOM checker (often wrong on implicit roles) | axe-core WCAG only (correct, principled) |
| Co-location | `behavior/` subfolder | Same file as render function |
| Sub-component ownership | All rules merged in folder JSON | Each spec owns its own rules |
| KB / WCAG refs | Typed array | `@keyboard` / `@see` JSDoc |
| Test runtime | ~30–45 min (browser) | **~5 min (JSDOM)** |

resolves:
https://progresssoftware.atlassian.net/browse/FE-1253
https://progresssoftware.atlassian.net/browse/FE-1211
https://progresssoftware.atlassian.net/browse/FE-1213
https://progresssoftware.atlassian.net/browse/FE-1238


zhpenkov:
Opening this as an RFC

@telerik/kendo-front-end can I have your input on this topic.
Can you review both approaches (https://github.com/telerik/kendo-themes/pull/5950 and current PR), and let's decide which one will be most suitable for our case.

packages/html/src/bottom-nav/bottom-nav-item.spec.tsx
```diff
@@ -62,5 +63,27 @@ BottomNavItem.states = states;
...
+BottomNavItem.folderName = null;
+
+const ariaSpec: AriaRule[] = [
+    {
+        selector: '.k-bottom-nav-item',
+        attribute: 'role',
+        value: 'link',
+        usage: 'Associate the link purpose of the bottom navigation items.',
+        nativeElement: 'a'
+    },
```
kirchoni: This becomes obsolete now. the spec already defines `role='link'` ([ref](https://github.com/telerik/kendo-themes/blob/d7828ef7d71791464017da41f60688adf36cd9ff/packages/html/src/bottom-nav/bottom-nav-item.spec.tsx#L49)) so why would we need it here as well? i would assume thats a direct migration from the json specs, but since this type of information is already embeded in the rendering itself i believe we can skip it. I assume there are more complex selectors, but still they should be achievable by the conditional attributes inside the specs as well (e.g. selected/focused etc)

kirchoni (dismissed):

packages/html/src/bottom-nav/bottom-nav-item.spec.tsx
zhpenkov: Those objects are consumed by an a11y test
which opens a browser with our templates and tests our specs' rendering if it complies with the JSON rules, values, etc

It seems logical to have such tests, especially when migrating from one structure to another, it ensures the 1:1 parity between the before (JSON) and after (Spec attributes) structure.

It's like a validation of whether we have migrated the changes correctly in the specs.

Do you think @telerik/kendo-front-end we should drop this validation and rely only on the Spec attributes, or should we keep it?

kirchoni: Do you consider this to be a temporary artifact until we've moved everything into the rendering itself? If so, we can test, validate and remove as part of a single PR, no need to be long-lived.

kirchoni: Also consider the specs to be the "truth" that validate other stuff. if we start validating it instead, it becomes questionable if it's trustworthy in the first place!

zhpenkov: I removed the aria JSONs.

Saw an opportunity to add `@aria` annotation, for useful info that we might want to have in the docs .md format.
Last commit contains it. Might leave that for additional context for the aria attributes.

zhpenkov:
@telerik/kendo-front-end I polished the PR so it is ready for a final review.
The description is updated with the recent changes.

zhpenkov:
This PR also resolves - https://github.com/telerik/kendo-themes-private/issues/561
More info can be found in the issue.

zhpenkov:
This PR also resolves - https://github.com/telerik/kendo-themes-private/issues/547
More info can be found in the issue.

epetrow (approved):

PR #6064, opened 2026-07-08
fix(loader): use transform-only animation for spinner-3 to avoid main…

kirchoni:
## Summary

Replace `inset-block-start`/`inset-inline-start` keyframes with `transform: translate()` based animation that runs on the compositor thread. This ensures the spinner continues animating even when the main JS thread is blocked.

## Changes

- Use CSS custom properties (`--_canvas-w`, `--_canvas-h`) for container dimensions
- Animate only `transform: translate()` in keyframes (compositor-eligible)
- Add `will-change: transform` for layer promotion
- Simplify segment selectors (shared animation, per-child delay only)

## Jira

[FE-1311](https://progresssoftware.atlassian.net/browse/FE-1311)

[FE-1311]: https://progresssoftware.atlassian.net/browse/FE-1311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kirchoni:
Closing in favor of #6053 which covers both spinner-3 and spinner-4.

PR #6019, opened 2026-06-16
Enhance Scheduler

zhpenkov:
closes: https://progresssoftware.atlassian.net/browse/FE-1263
closes: https://github.com/telerik/kendo-themes-private/issues/571

## 1. Scheduler Toolbar — Responsive Flow

**What changed:** The toolbar has a new element order, uses `fillMode="flat"` on the toolbar itself, and introduces a responsive collapsing strategy where elements are progressively hidden as available width decreases.

### Rendering Before

- Navigation was a `ButtonGroup` with "Today", prev, next grouped (solid fill).
- No "New Event" button in the toolbar.
- The view switcher was always a `SegmentedControl`.

### Rendering After (Full / Level 0)

- Toolbar has `fillMode="flat"`.
- **New Event** button added (primary, icon `plus`).
- Date button (`k-nav-current`) with calendar icon, `fillMode="flat"`.
- Prev/Next are now a `ButtonGroup` with `fillMode="flat"`.
- "Today" is a standalone flat button.
- Spacer separates left items from view switcher.
- View switcher is a `SegmentedControl` with `.k-scheduler-views`.
- "Work Week" added as a view option.

### Responsive Collapsing (4 Levels)

| Level | What changes |
|-------|-------------|
| **0** | Full layout — all elements visible, SegmentedControl for views |
| **1** | "New Event" text hidden (icon-only), SegmentedControl collapses to a **MenuButton** (shows current view name), separated by `k-separator` |
| **2** | "Today" button hidden |
| **3** | Prev/Next ButtonGroup hidden; only icon button + date + spacer + MenuButton remain |

Implementation should observe the available width and progressively hide elements.

## 2. Scheduler Edit Dialogs

Three dialog types were redesigned with a new form layout using icon-based labels.

### 2a. New Event — Quick Create (Popover)

Screenshot 2026-06-30 at 18 46 37

**Component:** `Popover` (not full Dialog, no overlay)

- Title: "New event"
- Form uses `.k-scheduler-edit-form` with `orientation="horizontal"`.
- Labels are **icons** (not text): `right-double-quotes` for title field, `clock` for time.
- Title field: Textbox with placeholder "New event".
- Time field: read-only text row (`k-scheduler-edit-form-row k-color-subtle`).
- Actions: "More options" (flat, primary, icon `arrows-resizing`) + spacer + Cancel + Save (primary), rendered as `ActionButtons` with `className="k-popover-actions"`.

### 2b. New Event — More Options (Dialog)

[...]

PR #6071, opened 2026-07-13
chore(infra): recompile all themes when a core SCSS file changes

zhpenkov:
## Summary

When a file inside `packages/core/scss/` is saved, the devkit watcher previously silently ignored the event because `core` was not in `NON_THEME_PACKAGES` but also wasn't a real theme — so the watcher matched it but trying to compile `core/all.css` would fail or do nothing useful.

This fix introduces a `SOURCE_ONLY_PACKAGES` set (initially just `core`) for packages that are design-system dependencies consumed by every theme. When a file in such a package changes, **all** theme packages are scheduled for recompile instead of just the changed package.

## Changes

- `packages/html/devkit/vite.config.ts`
  - Add `SOURCE_ONLY_PACKAGES = new Set(['core'])` constant
  - Extract `getThemePackages()` helper that lists all compilable theme packages
  - Extract `scheduleRecompile()` helper with the debounce logic (DRY)
  - In the watcher `onChange` handler: fan out to all themes when the changed package is in `SOURCE_ONLY_PACKAGES`

## Related

Extracted from #6019 (Enhance Scheduler) — this is a standalone infrastructure fix.

### DIFF
--- a/packages/html/devkit/vite.config.ts
+++ b/packages/html/devkit/vite.config.ts
@@ -14,6 +14,10 @@ const repoRoot = resolve(devkitDir, '../../..');
 // Packages under packages/* that are not selectable themes.
 const NON_THEME_PACKAGES = new Set(['utils', 'html']);
 
+// Packages whose SCSS changes should trigger a recompile of ALL themes (they are
+// design-system dependencies consumed by every theme via @use / @forward).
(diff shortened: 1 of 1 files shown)


epetrow (approved):
Looks good 👍

PR #6006, opened 2026-06-09
Grid grouping

TeyaVes:
closes: https://progresssoftware.atlassian.net/browse/FE-1236

## Rendering changes:

1.  Grouping header HTML structure change (all grouping templates/tests):

      The grouping row content was migrated from a `<p class="k-reset">` with loose text to a structured `<div class="k-grouping-row-content">` with a `<span class="k-grouping-row-text">` wrapper:
    
       **Before**
    ```
       <p className="k-reset">
           <Icon icon="chevron-down" />
           ID: 1
       </p>
    ```
       
       **After**
    ```
       <div className="k-grouping-row-content">
           <Icon icon="chevron-down" />
           <span className="k-grouping-row-text">ID: 1</span>
       </div>
    ```

2. For the grid grouping inside a no-indent/column/multiple columns, a value to a newly added CSS variable needs to be set:  `--kendo-grid-grouping-level`.
 
    Set as an inline style on the grouping row's <TableTd> to drive the padding-inline-start indent calculation in SCSS:
    
    `<TableTd style="--kendo-grid-grouping-level: 2">`
    
    The value is the nesting depth (e.g. 1, 2)

PR #6079, opened 2026-07-22
chore(html): add missing grid grouping with checkboxes template

epetrow:
related to: https://github.com/telerik/kendo-themes/pull/6006


packages/html/src/grid/tests/grid-grouping-checkbox.tsx
TeyaVes, line 1: Do we need a new test file for this? Can't we add it to grid-grouping test, or will it become too large? I don't see the point of adding GridGrouping to the test as well (without the checkboxes) since we already have it in another test

epetrow: I added a new test since there are multiple grouping tests that I see - for angular, for blazor, etc.

packages/html/src/grid/templates/grid-grouping-checkbox.tsx
```diff
@@ -0,0 +1,124 @@
...
+                                <TableTd className="k-table-group-td k-group-cell" role="presentation"></TableTd>
+                                <TableTd className="k-select-cell" role="gridcell">
+                                    <Checkbox />
+                                </TableTd>
+                                <TableTd colspan={0} role="gridcell">1</TableTd>
+                                <TableTd colspan={0} role="gridcell">Chai</TableTd>
+                                <TableTd colspan={0} role="gridcell">$18.00</TableTd>
+                                <TableTd colspan={0} role="gridcell">Beverages</TableTd>
+                            </TableRow>
+                            <TableRow className="k-master-row" role="row" alt selected>
```
zhpenkov: do we need this selection in the template

zhpenkov (approved):

PR #6080, opened 2026-07-23
Update List Checkbox a11y

### LINKED ISSUE #6081: [A11y] Update List Checkbox properties
### Description

List items rendered with `showCheckbox` (used by MultiSelect, DropdownList "select all", ComboBox, etc.) render an unlabeled `\<input type="checkbox">` inside `ListItem`. Automated accessibility checkers (axe-core and similar) flag these as form controls without an associated accessible name, since there's no `\<label>` and no `aria-label`/`aria-labelledby` on the checkbox.

Additionally, the checkbox input is focusable by default, meaning `role="none"`/`aria-hidden="true"` cannot safely be applied to it without also removing it from the tab order — applying `aria-hidden="true"` to a focusable element is itself an axe violation (`aria-hidden-focus`).

### Expected behavior

The checkbox should either:

- Have a valid accessible name matching the visible item text (`aria-label`), and be removed from the tab order (`tabIndex="-1"`) since keyboard interaction/focus is owned by the `\<li role="option">` row (consistent with how `TreeviewItem` and `ColumnListItem` already handle this), or
- Be fully hidden from the accessibility tree, provided it's already non-focusable.

The fully hidden part involves rending the checkbox as a span, but for now we can proceed with addind the labels as a quick win.

### Affected component(s)

- `packages/html/src/list/list-item.spec.tsx` (`ListItem` — used by MultiSelect, DropdownList, ComboBox popups with checkboxes)

zhpenkov:
closes: https://github.com/telerik/kendo-themes/issues/6081

https://progresssoftware.atlassian.net/browse/FE-1321


packages/html/src/list/tests/list-checkboxes.tsx
```diff
@@ -0,0 +1,35 @@
...
+                </ListWithCheckboxes>
+            </section>
+
+            <section>
+                <ListWithCheckboxes>
+                    <ListItem showCheckbox>Unchecked</ListItem>
+                    <ListItem showCheckbox checked>Checked</ListItem>
+                    <ListItem showCheckbox disabled>Disabled</ListItem>
+                    <ListItem showCheckbox checked disabled>Checked + disabled</ListItem>
+                </ListWithCheckboxes>
```
epetrow: Is there an actual rendering difference between these two approaches? If not I don't see a benefit of having separate showcases of using the "API" (using the text prop/using string children)

zhpenkov: Yes, that is valid.
I'll leave just 1

epetrow (approved):

Issue #5968, opened 2026-05-26
[Dialog][Meridian] Dialog declared within Card is scoped to Card


FloppyNotFound:
### Description

When declaring a Kendo Dialog within a Kendo Card, it opens within the scope of the card and does not expand over the whole screen.

### Steps To Reproduce

see https://stackblitz.com/edit/angular-tajpkvwi?file=src%2Fapp%2Fapp.component.ts

- Declare a Dialog within a Card
- Open the Dialog

### Screenshots or video

<img width="325" height="267" alt="Image" src="https://github.com/user-attachments/assets/6cbe1be2-b0ba-424d-9a4c-3850cb6f8f00" />

<img width="327" height="274" alt="Image" src="https://github.com/user-attachments/assets/9d7ba3b6-30eb-44ba-b741-a764e5457d20" />

### Actual Behavior

The dialog is constrained to the surrounding <kendo-card>, because Meridian sets backdrop-filter: blur(...) on .k-card, and per CSS spec that establishes a new containing block for fixed-positioned descendants.

Other Kendo themes (Default, Bootstrap, Material) do not put backdrop-filter on .k-card, so this only reproduces with Meridian.

### Expected Behavior

Clicking the button opens a dialog that overlays the whole viewport (Dialog uses position: fixed).

### Browser

Chrome

### Browser version

latest

### OS type

MacOS

### OS version

26.5

### Last working version of the Kendo UI for Angular package (if regression).

other Themes

FloppyNotFound:
Temporary workaround:
```
.k-card,
.k-card-callout {
  backdrop-filter: none;
}
```

FloppyNotFound:
@epetrow 
I can confirm that the issue with the Card is fixed.

Unfortunately, other elements seem to share a similar issue - Elements with their own stacking context that are outside the card's stacking context render on top of the overlay.

Please check https://stackblitz.com/edit/angular-tajpkvwi-muk6bc8g?file=src%2Fapp%2Fapp.component.ts for a Demo.

Workaround:
```
.k-card {
  isolation: auto !important;
}
```

kirchoni:
Hey, we just released a new `dev` version of the `@progress/kendo-theme-meridian` where the issue should be fixed (should land an official version soon). Please check the updated stackblitz example:
https://stackblitz.com/edit/angular-tajpkvwi-cxxbtsu2?file=package.json

Let us know if you experience any issues.

PR #6116, opened 2026-08-10
fix(meridian): keep background customizable on .k-card while fixing dialog scoping

kirchoni:
will try to solve https://github.com/telerik/kendo-themes/issues/5968
continuation of https://github.com/telerik/kendo-themes/pull/5990

Having the background-color on the `:before` makes it unintuitive for customization, (e.g. "changing background of a card" as was the case in the [`ai-automotive-industry` PT](https://github.com/telerik/kendo-building-blocks/blob/0e027b098bde1ad27bd58c6b793d5937cf4bca2f/react/templates/src/components/automotive-industry/App.css#L205-L207))

Background is applied to the `k-card`, dropping the `isolate` rule, while the backdrop-filter remains on the `:before`.

P.S. not sure whats the issue with the images.


epetrow (approved):
Seems to work

PR #6168, opened 2026-09-04
fix(ci): reconcile visual reports with pull requests

kirchoni:
There was this strange race condition between branch push, draft PR and PR ready for review, that would prevent me from triggering the image diffs in https://github.com/telerik/kendo-themes/pull/6167 which forced me to manually trigger em. Fixing it. 

smarty pants description bellow
---

## Summary

- retry the initial commit-to-PR lookup briefly to cover the common first-push timing window
- add a lightweight PR-open/CI-completed reconciler that moves an existing commit-keyed report to `reports/pr-<number>/` and posts the visual report comment
- mark reports with their source SHA so stale, superseded, and closed-PR cleanup cannot delete a newer report
- preserve pending commit-keyed feature reports during baseline pruning and keep feature runs from rewriting the `develop` alias
- coordinate report mutation by source branch, retry non-fast-forward cleanup pushes, and remove closed PRs from the reports index

## Security and lifecycle

The reconciler uses `pull_request_target` only for metadata and trusted base-branch workflow code. It never checks out or executes pull-request code, explicitly ignores fork PRs, and grants only `contents: write` and `pull-requests: write` to the reconciliation job. Draft PRs are handled by the normal `opened` event.

The PR-open path waits briefly for an in-flight report; the `workflow_run: completed` path provides the durable follow-up when CI finishes later. Current PR state/head and the persisted source-SHA marker are checked before moving, deleting, or commenting, so canceled and superseded runs cannot overwrite newer reports.

## Validation

- `actionlint .github/workflows/associate-visual-regression.yml .github/workflows/cleanup-visual-regression.yml`
- YAML parsing for all three changed workflows
- `git diff --check`

`_visual-regression.yml` retains its two pre-existing actionlint findings: the organization-allowed `peaceiris/actions-gh-pages@v3` pin and an existing ShellCheck SC2016 notice.

### DIFF
--- a/.github/workflows/_visual-regression.yml
+++ b/.github/workflows/_visual-regression.yml
@@ -84,12 +84,24 @@ jobs:
         uses: actions/github-script@v9
         with:
           script: |
-            const { data: prs } = await github.rest.repos.listPullRequestsAssociatedWithCommit({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              commit_sha: context.sha,
-            });
-            const openPr = prs.find(pr => pr.state === 'open');
+            // Cover the common first-push race cheaply here. The PR lifecycle
+            // reconciler handles PRs opened after this bounded wait expires.
+            let openPr;
+            for (let attempt = 1; attempt <= 6; attempt++) {
+              const { data: prs } = await github.rest.repos.listPullRequestsAssociatedWithCommit({
+                owner: context.repo.owner,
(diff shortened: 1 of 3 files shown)


zhpenkov (approved):
Good

Visual regression reports get more and more stable with fixes like this one.

After some time, when we reach good stability, we can see ways to optimize/simplify it in terms of code.

PR #6169, opened 2026-09-04
feat(scheduler): dim original event and show selected state on drag hint

bmarkova-progress:
## Summary

Changes how the Scheduler event drag hint looks:
- **Before**: the moving copy faded to 50% opacity; the original event stayed unchanged.
- **After**: the moving copy (`.k-event-drag-hint`) renders with the `.k-selected` fill/border/shadow; the original event, left at its source slot, gets a new `.k-event-drag-source` class dimmed to 50% opacity.

Applies to all themes via `packages/core`.

## Changes

- `packages/core/scss/components/scheduler/_theme.scss` — `.k-event-drag-hint` now shares the same fill/box-shadow rule as `.k-selected`.
- `packages/core/scss/components/scheduler/_layout.scss` — removed opacity from `.k-event-drag-hint`; added new `.k-event-drag-source { opacity: .5; }`.
- `packages/html/src/scheduler/tests/scheduler-events.tsx` — added a `k-event-drag-source` visual test case alongside the existing `k-event-drag-hint` one.

## Follow-up needed (outside this repo)

The Scheduler component packages (kendo-react/-angular/-jquery/-vue) apply these classes during drag interactions and will need to be updated to add `k-event-drag-source` to the original event (in addition to the existing `k-event-drag-hint` on the moving clone).

## Review

@zhpenkov — please review the SCSS implementation (class naming, structure, whether merging `.k-event-drag-hint` into the `.k-selected` fill rule is the right approach) and flag anything that would complicate wiring this up on the component side.

---
🤖 Generated with [Copilot](https://github.com/copilot)

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/packages/core/scss/components/scheduler/_layout.scss
+++ b/packages/core/scss/components/scheduler/_layout.scss
@@ -595,8 +595,8 @@
 
 
     // Draging hint
+    // The moving copy of the event, rendered with the `.k-selected` fill (see _theme.scss).
     .k-event-drag-hint {
-        opacity: .5;
         z-index: k-z-index("base", 3);
 
         .k-event-top-actions,
@@ -608,6 +608,11 @@
         .k-event-time {
             display: block;
         }
+    }
+
+    // The original event, left in place at its source slot while it is being dragged.
+    .k-event-drag-source {
+        opacity: .5;
     }
 
 
--- a/packages/core/scss/components/scheduler/_layout.scss
+++ b/packages/core/scss/components/scheduler/_layout.scss
@@ -595,8 +595,8 @@
 
 
     // Draging hint
-    // The moving copy of the event, rendered with the `.k-selected` fill (see _theme.scss).
     .k-event-drag-hint {
+        opacity: .5;
         z-index: k-z-index("base", 3);
 
         .k-event-top-actions,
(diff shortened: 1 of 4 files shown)


kirchoni:
@zhpenkov when you get to this, please be careful with introducing new class names - the `-source` one.

bmarkova-progress:
little interactive prototype: https://bmarkova-progress.github.io/scheduler-drag-hint-prototype/

Try dragging the event — the original slot dims (`.k-event-drag-source`), and the moving copy shows the new selected-state look (`.k-event-drag-hint`).

zhpenkov:
Great, looks better this way.

Currently, the "source" event has a `k-event` class, and the "dragged" event receives `k-event-drag-hint`.

I was just wondering... If we want to make the "source" event dimmed, shouldn't we just switch the logic without any style updates?

After the switch, the "source" event will have a `k-event-drag-hint` class, and the "dragged" event will receive only a `k-event` class.

bmarkova-progress:
Yes, this way we don't need to add a new class; the change will be in the logic: during a drag interaction, apply `k-event-drag-hint` to the original/source event element, and apply `k-selected` (instead of `k-event-drag-hint`) to the moving/dragged copy.

GitHub copilot did: Pushed a revert of the CSS changes from this PR — `.k-event-drag-hint` and `.k-selected` are back to their original styling, no new class needed on the theme side. The component-side change (swapping which class gets applied to which element during drag) is tracked as follow-up in the consuming Scheduler packages.

zhpenkov:
closing PR as no changes are needed from the theme

PR #6184, opened 2026-09-14
feat(segmented-control): add roundness options

kirchoni:
By design we're given 9 roundess options, however, i'm sticking to the common 5 ones (none, small, medium, large, full) for consistency with the rest of the components.

---

## Summary

- add the established `none`, `small`, `medium`, `large`, and `full` roundness options to SegmentedControl
- propagate explicit container roundness to the thumb and segment buttons for consistent nested radii
- configure the default `full` roundness across Default, Bootstrap, Classic, Material, Fluent, and Meridian
- add a dedicated visual roundness matrix and regenerate customization documentation

## Related work

- [FE-1351](https://progresssoftware.atlassian.net/browse/FE-1351)
- [UX-646](https://progresssoftware.atlassian.net/browse/UX-646)
- [Finalized Figma specification](https://www.figma.com/design/wGlDL8f8LP74oyRSsY9EIw/SegmentedControl?node-id=107-1913&p=f&t=VsudMaadRbSDZt4w-0)

[FE-1351]: https://progresssoftware.atlassian.net/browse/FE-1351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[UX-646]: https://progresssoftware.atlassian.net/browse/UX-646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/scss/components/segmented-control/_layout.scss
```diff
@@ -21,6 +22,21 @@
...
         box-sizing: border-box;
 
+        // Roundness
+        @each $roundness in $kendo-segmented-control-roundness {
+            #{k-when-default($kendo-segmented-control-default-roundness, $roundness)}
+            &.k-rounded-#{$roundness} {
+                border-radius: k-border-radius($roundness);
+            }
+
+            &.k-rounded-#{$roundness} {
```
zhpenkov: For every `n` roundness, we set

```
                .k-segmented-control-thumb,
                .k-segmented-control-button {
                    border-radius: inherit;
                }
```

Are those style repetitions redundant? They can be outside the loop.

kirchoni: Good catch. I moved the invariant thumb/button `border-radius: inherit` rule outside the roundness loop, so it is emitted once while the loop only generates the per-roundness container radius.

zhpenkov (approved):

zhpenkov:
Great, just consolidate all commits into one feat for the changelog.

TeyaVes (approved):

kirchoni:
Well, they were, in fact, not concentric...rebased and corrected the calculations. Please revisit after the snapshots update! @TeyaVes @zhpenkov

packages/core/scss/components/segmented-control/_layout.scss
```diff
@@ -21,6 +23,28 @@
...
+
+                @if $roundness != "full" {
+                    @include concentric-radius(
+                        k-border-radius($roundness),
+                        $kendo-segmented-control-padding-y,
+                        ".k-segmented-control-thumb, .k-segmented-control-button"
+                    );
+                } @else {
+                    border-radius: k-border-radius($roundness);
+                }
```
zhpenkov: The concentric border radius needs 2 value parameters: radius and padding

Here we cover the `radius={n}`/`size={undefined}` case excellently.
But the `radius={n}`/`size={n}` case will not produce concentric border radii, as the second padding parameter is hardcoded: `$kendo-segmented-control-padding-y`.

This needs to be put inside the size loop, and the padding parameter needs to be passed.

zhpenkov (changes requested):
We need to make a `roundness/size` matrix test.

zhpenkov:
We need a size/rounded test (like the switch) to catch changes in the above-mentioned behaviour:

Screenshot 2026-09-18 at 9 57 09

packages/core/scss/components/segmented-control/_layout.scss
zhpenkov: Actually:

Concentricity is measured against the container padding, which is the
actual offset between the outer edge and the thumb/button, not the
button's own (internal) padding around its text/icon.

zhpenkov (approved):

PR #6185, opened 2026-09-14
chore: make Kendo SVG Icons v5 the default

kirchoni:
Markup classNames were using the `name` attribute from the svg's path, not the `props.name` prop, this is why we had `-outline` classNames in the html.

---

## Summary

- update `@progress/kendo-svg-icons` from v4.11 to v5.4 across all theme packages and the HTML package
- migrate remaining v4-only HTML fixture names to names available in v5
- replace the Rating component's temporary hardcoded star SVGs with the v5 `star` variants
- retain runtime compatibility responsibility in the product components; themes continue to provide version-agnostic icon styling


TeyaVes (changes requested):
We need to make sure that the default (outline) icons use the stroke-width variable as they currently do not. As a result:
1. their stroke width is 1, not 1.5, which makes them really light;
2. their stroke-width cannot be modified;

TeyaVes (approved):
great, we just need to bump the icons to the stable 5.5 version and we're good to g

zhpenkov (approved):

zhpenkov:
Merging with `kendo-svg-icons` version - 5.5.0-dev.0
We will proceed to publish a dev version of `kendo-themes`.

Before publishing a stable version of `kendo-themes`, we need to bump `kendo-svg-icons` to an official version.

PR #6188, opened 2026-09-15
fix(toolbar): apply default border radius to overflow buttons

ivailo-al:
https://github.com/telerik/kendo-themes/issues/6196

## Summary
The Toolbar's overflow anchor button (`.k-toolbar-overflow-button`) and the buttons inside the overflow popup (`.k-overflow-button`) hardcoded `border-radius: 0`, causing them to look flat/square instead of following the theme's default Button/List roundness.

## Changes
- Removed the hardcoded `border-radius: 0` on `.k-toolbar-overflow-button` in `packages/core/scss/components/toolbar/_layout.scss` so it now inherits the Button component's default roundness.
- Added a new variable `$kendo-toolbar-overflow-button-border-radius` (null in core) and used it for `.k-overflow-button` (the popup list items) instead of a hardcoded `0`.
- Set the new variable to `k-border-radius(sm)` in the **Meridian** theme only, matching the List item's border radius (`$kendo-list-item-border-radius`). Other themes keep their previous (0/omitted) look since the variable stays `null` for them.
- Updated the HTML toolbar templates/tests to drop the now-unneeded `rounded="none"` prop on the overflow toggle button, since the SCSS no longer forces a square corner.

## Testing
- `npm run sass` — compiles cleanly across all themes.
- `npm run lint:styles` — no violations.
- Verified compiled CSS: Meridian's `.k-overflow-button` now has `border-radius: var(--kendo-border-radius-sm)`; other themes (`default`, `bootstrap`, `material`, `fluent`, `classic`) have no `border-radius` declaration (unchanged visual result).

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>


zhpenkov:
Great fix. Just a few things to clean up.

Lets remove the newly added variable for the overflow button border-radius, and leave the border radius to be handled by the default border radius of the button.

So the fix would be

Remove the radius="none" from the TSX

and remove the 

border-radius: 0;

of the overflow button

zhpenkov (approved):

PR #6195, opened 2026-09-17
FloatingToolbar as a standalone component

epetrow:
relates to: https://progresssoftware.atlassian.net/browse/DT-2092


packages/html/src/floating-toolbar/floating-toolbar.spec.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+export type KendoFloatingToolbarProps = KendoFloatingToolbarOptions & KendoPopupProps & KendoToolbarProps & {
+    /** Renders a permanently-visible drag handle that bypasses the toolbar's overflow/scroll strategy. */
+    draggable?: boolean;
+    /** Accessible name for the drag handle button. */
+    dragHandleAriaLabel?: string;
+};
+
+/**
+ * @aria {role="button"} The drag handle is a non-semantic element exposed as a button; it is a Tab stop independent of the Toolbar's own roving-tabindex group.
```
zhpenkov, line 33: Should the drag handle be a button
Lets research wether this is the most suitable role

packages/html/src/pdf-viewer/tests/pdf-viewer-search-panel.tsx
```diff
@@ -29,13 +29,9 @@ export default () =>(
...
             <PDFViewerWithAnnotations />
-            <FloatingToolbar className="annotations-search-panel">
-                <Button className="k-search-dialog-draghandle" fillMode="flat" icon="handle-drag-dots"/>
-                <ToolbarSeparator />
+            <FloatingToolbar className="annotations-search-panel" draggable>
                 <Textbox fillMode="flat" placeholder="Search" suffix={<Button className="k-match-case-button" fillMode="flat" icon="convert-lowercase" />}  />
-                <div className="k-search-matches">
-                   <span>0</span> of <span>0</span>
-                </div>
+                <span><span>0</span> of <span>0</span></span>
```
zhpenkov, line 34: lets revert that

packages/html/src/floating-toolbar/floating-toolbar.spec.tsx
epetrow: It is not actually a button, just has a role button which here is correct and is actually the modern recommended pattern.

- The legacy drag-and-drop ARIA properties (aria-grabbed/aria-dropeffect) were deprecated in ARIA 1.1 specifically because they were unreliable with screen readers.
- The WAI-ARIA Authoring Practices' replacement pattern is exactly what's documented here: expose the handle as a toggle button (`role="button" + aria-pressed`) that engages a keyboard-drivable "grabbed"/"Move" mode, then use arrow keys to reposition and Enter/Escape to commit/cancel. That matches this spec's `@aria {aria-pressed=...}` + `M`/arrow-key/`Enter/Escape` keyboard contract precisely.

zhpenkov:
Looking Great, just two small things!

This is transparent:
Screenshot 2026-09-23 at 16 43 54

And also check meridian translucency

zhpenkov (approved):
Great!

TeyaVes (approved):
We just need to check the Draggable + Scrollable toolbar with translucency in the merdian theme

